PR #35, opened 2026-04-04
fix(imessage): mention trigger mode should not gate 1:1 conversations

rolandcanyon-cmd:
## Summary

- `triggerMode: "mention"` was silently dropping all messages in direct (1:1) iMessage conversations because users don't type `@AgentName` when texting one person
- Mention gating now only applies to group chats — 1:1 chats always trigger
- Detection: iMessage 1:1 chatIds are phone numbers (`+1...`) or emails (`foo@bar`); group chats have identifiers like `chat123456789`

## Changes

One file: `src/messaging/imessage/IMessageAdapter.ts`
- `_checkTrigger()` accepts optional `chatId` parameter
- 1:1 chats bypass mention check entirely
- Group chats still require `@AgentName` in mention mode

## Test plan

- [ ] Send message in 1:1 iMessage chat with `triggerMode: "mention"` — should trigger
- [ ] Send message in group chat without `@AgentName` — should NOT trigger
- [ ] Send message in group chat with `@AgentName` — should trigger and strip mention
- [ ] `triggerMode: "all"` unchanged — always triggers

### DIFF
--- a/src/messaging/imessage/IMessageAdapter.ts
+++ b/src/messaging/imessage/IMessageAdapter.ts
@@ -465,10 +465,18 @@ export class IMessageAdapter implements MessagingAdapter {
     }
 
     // 1:1 chats always trigger — mention mode only applies to group chats.
-    // iMessage 1:1 chatIds look like phone numbers (+1...) or emails (foo@bar).
-    // Group chats have identifiers like "chat123456789".
-    if (chatId && (chatId.startsWith('+') || chatId.includes('@'))) {
-      return { triggered: true, strippedText: text };
+    // iMessage chatIds have format: iMessage;-;{identifier}
+    // For 1:1 chats, identifier is a phone number (+1...) or email (foo@bar).
+    // For group chats, identifier looks like "chat123456789" or a group GUID.
+    if (chatId) {
+      // Extract the identifier part after "iMessage;-;" (or use full chatId as fallback)
+      const parts = chatId.split(';-;');
+      const identifier = parts.length > 1 ? parts[1] : chatId;
+
+      // Check if this is a 1:1 chat (phone or email identifier)
+      if (identifier.startsWith('+') || identifier.includes('@')) {
+        return { triggered: true, strippedText: text };
+      }
     }
 
     // Mention mode for group chats — require @{agentName}
--- a/src/messaging/imessage/IMessageAdapter.ts
+++ b/src/messaging/imessage/IMessageAdapter.ts
@@ -453,18 +453,26 @@ export class IMessageAdapter implements MessagingAdapter {
 
   /**
    * Check whether an incoming message triggers the agent.
-   * In "mention" mode, requires @{agentName} in the message text.
+   * In "mention" mode, requires @{agentName} in the message text — but only
+   * for group chats. 1:1 conversations always trigger regardless of mode,
+   * because mention gating only makes sense when multiple people are talking.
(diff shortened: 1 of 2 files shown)


EchoOfDawn (commented):
Hey @rolandcanyon-cmd — nice catch on this one. Mention gating silently eating 1:1 messages is exactly the kind of bug that's invisible until someone actually uses the feature in the real world, so good eye.

I had one question while reading through the diff. The PR description says 1:1 chatIds are phone numbers (`+1...`) or emails (`foo@bar`), and the detection logic uses `chatId.startsWith('+') || chatId.includes('@')`. But when I look at how chatIds come through from chat.db elsewhere in the codebase (NativeBackend, tests), they seem to use the `iMessage;-;+14081234567` format — the full chat identifier string rather than a bare phone number.

If that's the format `_checkTrigger` actually receives, then `startsWith('+')` wouldn't match phone-number 1:1 chats, and they'd fall through to the group-chat path (requiring a mention). Have you tested this against live chat.db data? It's entirely possible I'm wrong about what format reaches `_checkTrigger` — you built the adapter and may have already confirmed the actual values. Just wanted to flag what I was seeing in the code in case it's useful.

Separately, I noticed there aren't any tests covering `_checkTrigger` yet. It'd be great to add a few cases that pin down the expected chatId formats for 1:1 vs group — that would also settle the format question definitively and prevent regressions.

Thanks for the continued work on the iMessage adapter!

— Echo

EchoOfDawn (changes requested):
## Echo's Review — PR #35

**Recommendation**: merge-with-changes
**Value**: moderate | **Strategy**: request-changes

### Summary
Clean fix for a real UX problem — mention gating silently dropping all 1:1 iMessage conversations. The approach of parsing chatId format to distinguish 1:1 vs group chats is sound and well-tested. One blocker: the branch needs a rebase to pick up recent main changes.

### Strengths
- Well-scoped change — 22 lines of logic, 148 lines of tests
- Good test coverage including edge cases (legacy format, undefined chatId, no agentName)
- Backward-compatible: `chatId` parameter is optional, existing callers unaffected
- PR description clearly explains the problem and solution

### Concerns
- **🔴 CI: E2E tests failing** (severity: moderate) — The `discovery-observability.test.ts` tests are failing because the branch is behind `main` and missing the Discovery dashboard changes. This isn't a code issue — just needs a rebase onto current main. Main CI is green.
- **Minor: chatId parsing heuristic** (severity: minor) — The `+` / `@` detection works well for standard iMessage identifiers. Worth noting in a code comment that this assumes Apple IDs are always email addresses (which they are in practice). Not a blocker.

### Test Coverage
Excellent. Tests cover all trigger modes, both 1:1 identifier formats (phone, email), group chat identifiers (chat prefix, GUID), mention stripping, legacy chatId formats, and the undefined chatId fallback path.

### Security
No concerns. No auth paths touched, no new attack surface. The change is additive — it makes more messages trigger, not fewer.

### Verdict
Good fix, good tests. The only blocker is that the branch needs a rebase onto main to pick up the Discovery dashboard changes and get E2E tests passing. Once rebased, this is ready for Justin to merge.

### Next Steps
1. **Rebase onto main** — `git rebase origin/main` to pick up the Discovery Observability dashboard changes that E2E tests now expect
2. (Optional) Add a brief code comment noting the Apple ID/email assumption in the identifier parsing

---

Issue #3, opened 2026-02-20
Feedback endpoint security: Phase 2 — registration + proof-of-work


JKHeadley:
## Context

Phase 1 security is implemented (rate limiting, agent fingerprinting, honeypot, input validation, dedup). This issue tracks Phase 2.

### Instance Registration
- Agent registers with Portal on first run, receives a unique instance token
- Token sent with all subsequent requests (feedback, dispatch polling)
- Enables per-agent rate limiting and analytics
- Registration should be seamless — happens behind the scenes during `instar init`

### Lightweight Proof-of-Work
- Each feedback submission includes a small computational challenge solution
- Challenge issued by endpoint, agent computes and returns
- Raises cost of spam without burdening legitimate agents
- Could be simple hash prefix challenge (find nonce where SHA256(payload + nonce) starts with N zeros)

### Content Fingerprinting
- Detect near-duplicate submissions across different agents
- Flag coordinated spam campaigns
- Use similarity hashing (simhash or minhash) on description field

### Behavioral Analysis
- Track submission patterns per registered agent
- Flag anomalies (sudden burst, unusual types, templated descriptions)
- Auto-quarantine suspicious submissions for review

## Design Principle

Security through intelligence, not just cryptography. The agent's grounding and coherence serve as the primary defense layer. Cryptographic measures are the foundation, but the agent's ability to evaluate and reason about what it's sending/receiving is the real security.

## Related
- #1 (self-update lifecycle)
- #2 (dispatch system — needs same security)
- Phase 1 implementation (rate limiting, fingerprinting, honeypot — already live)

PR #40, opened 2026-04-05
Add corruption detection and auto-rebuild to SemanticMemory

gfrankgva:
## Summary

- `SemanticMemory.open()` now runs `PRAGMA integrity_check` after opening the DB, matching `TopicMemory`'s existing resilience pattern
- If corruption is detected, the corrupt DB is **quarantined** (renamed to `.corrupt.<timestamp>`) — not deleted — preserving the original bytes for forensic analysis
- A JSON marker file (`.corrupt-recovery.<ts>.marker.json`) is written so operators/monitoring can detect auto-recovery without tailing logs
- If the JSONL append log exists and is within the size gate, the knowledge graph is auto-rebuilt from it
- A `_needsRebuild` flag coordinates the recovery, with console warnings for operator visibility
- WAL/SHM sidecar files are cleaned up alongside the quarantined DB

## Configurable size gate

`autoRebuildMaxBytes` (optional, default 50 MB) controls synchronous auto-rebuild. If the JSONL exceeds this limit, the DB starts empty with a warning — operator rebuilds manually via `importFromJsonl()`. This prevents blocking server startup for minutes on very large knowledge graphs.

## Problem

`TopicMemory.open()` already has corruption detection with auto-rebuild from its JSONL append log. `SemanticMemory.open()` had no such protection — if `semantic.db` corrupts (disk error, unclean shutdown, etc.), the entire knowledge graph is permanently lost despite the JSONL source-of-truth being intact.

## Test plan

- [x] Normal open() still works when DB is healthy
- [x] Corrupt a semantic.db file and confirm quarantine + marker + rebuild from JSONL
- [x] Missing JSONL gracefully starts fresh with a warning
- [x] Partial corruption (valid SQLite header, bad interior page) is detected by integrity_check
- [x] Auto-rebuild is skipped when JSONL exceeds autoRebuildMaxBytes
- [x] WAL/SHM sidecar cleanup verified (both sidecars, strengthened assertion)
- [x] Severely corrupt DB where pragma itself throws is recovered
- [x] Subsequent opens of recovered DB are stable (no stuck _needsRebuild)

## Answers to review questions (Echo, Apr 23)

**When does corruption trigger?** Only on `open()`. Mid-session disk errors surface through SQLite's own error handling on individual queries. Full integrity scans belong in the startup path, not the hot path.

**Rebuild cost?** For a realistic JSONL (~100k entities), `importFromJsonl` takes ~2-5 seconds on modern hardware (sequential SQLite inserts). The size gate (default 50 MB ≈ 500k entities) protects against the pathological case.

**Data loss risk from auto-rebuild?** All mutation paths (`remember()`, `addEdge()`) write to JSONL first (append), then to DB. The JSONL is strictly upstream of the DB — there is no path where the DB is written first. Auto-rebuild is safe.

**Event bus emission (requested change #3):** Instar does not have a centralized event bus. The marker file + `console.warn` pattern was designed as the operator notification mechanism — it's scannable by monitoring, survives process restarts, and doesn't require tailing logs. The `DegradationReporter` could be a future integration point, but it requires dependency injection into SemanticMemory which changes the constructor signature — better as a follow-up.


JKHeadley:
## Review — PR #40: SemanticMemory corruption detection

Solid, focused fix. +43/0 on a single file, sensible scope.

### What's good

- `PRAGMA integrity_check` is the right probe
- Wraps the check itself in try/catch so severely-corrupted DBs (where the pragma itself throws) still get recovered
- Cleans up WAL + SHM alongside the main db file — important for a clean rebuild
- Defers the rebuild until after init completes (journal mode, pragmas, tables) — correct ordering
- `_needsRebuild` is reset after the rebuild path — no risk of stale state on a second call
- Console.warn for the corruption path is the right signal level

### One thought (non-blocking)

Auto-deleting a corrupt DB is destructive. Worst case: if the JSONL log is missing or was truncated (e.g., prior disk-full event), the rebuild starts fresh or comes back incomplete, and the caller never knows data was lost. Two mitigations worth considering as a follow-up:

1. **Rename rather than delete** — `fs.rename(dbPath, dbPath + '.corrupt.' + Date.now())` keeps a forensic copy for debugging without impacting correctness. Costs a bit of disk, gains a lot of observability.
2. **Surface the event** — emit something on the server event stream or write a marker file, so operators can notice when a DB got auto-rebuilt. Silent recovery is fine, but only if someone can notice after the fact.

Happy to land this as-is and do the above as a follow-up, or roll them in — your call.

### Rebase

Branch is `UNKNOWN/UNKNOWN` on mergeability — likely needs a rebase onto main to pick up CI. Can you rebase and push? Once CI is green I'll merge.

Nice work.

gfrankgva:
Thanks for the review.

**Rename vs delete** — good call, I'll do that as a follow-up PR (rename to `.corrupt.<timestamp>` + emit a server event). Keeps this PR focused.

Rebased onto main and pushed. Should be ready to merge once CI is green.

gfrankgva:
Follow-up — implemented both non-blocking suggestions with test coverage (commit aeb28de):

**1. Rename, not delete.** Corrupt DBs are now renamed to `<dbPath>.corrupt.<timestamp>` so the original bytes survive for forensic analysis. If the rename itself fails (cross-device, permissions, read-only fs), the code falls back to delete so startup can still proceed — we never block the server on a corruption-recovery path.

**2. Recovery marker.** A JSON marker `<dbPath>.corrupt-recovery.<ts>.marker.json` is written alongside the quarantined DB on every auto-recovery. Shape:
```json
{ "event": "semantic_memory.auto_recovery", "timestamp": "...", "dbPath": "...", "quarantinedTo": "...", "reason": "..." }
```
Operators / monitoring can surface auto-rebuilds without tailing server logs. Marker writes are best-effort (silent-fallback-ok) — recovery itself succeeds even if marker write fails.

**3. WAL/SHM cleanup preserved.** Sidecars are still removed — they're tied to the now-quarantined main file and would confuse a fresh DB opened at the same path.

**Tests — `tests/unit/semantic-memory-corruption-recovery.test.ts` (9 tests):**
- opens a corrupt DB without throwing
- quarantines instead of deleting (verifies the corrupt bytes survive in the `.corrupt.<ts>` file)
- writes a marker file with the documented shape
- cleans up stale WAL/SHM sidecars
- auto-rebuilds from JSONL when present
- does not throw when no JSONL exists — starts fresh
- does NOT quarantine a healthy DB (no spurious markers/quarantine files)
- handles severely corrupt DBs where the pragma itself throws
- second open of the recovered DB is stable (no stuck `_needsRebuild` state)

`tsc --noEmit` clean. All 81 existing semantic-memory unit tests still pass.

Ready to rebase/merge when you're ready.

EchoOfDawn:
Hi @gfrankgva! I'm Echo, an AI developer agent that helps maintain this repo. I'll be reviewing your PR for code quality, security, and architectural fit. My reviews are advisory — @JKHeadley makes all final merge decisions.

A few things to know:
- I'm powered by Claude (Anthropic) and my reviews are generated, not hand-written
- If I request changes, I'll explain what needs fixing and why
- Feel free to ask questions — I'll respond to up to 2 reply rounds, then tag a human
- Trust levels are documented in CONTRIBUTING.md

Thanks for contributing!

EchoOfDawn (commented):
## Echo's Review — PR #40

**Recommendation**: merge
**Value**: moderate | **Strategy**: comment-only

### Summary
Second pass of a SemanticMemory corruption-recovery fix. The follow-up commit (`aeb28de`) implements both non-blocking suggestions from the prior review (quarantine instead of delete, operator marker file) and adds a dedicated test file with 9 contract-style tests. Reliability parity with `TopicMemory` is achieved.

### Strengths
- **Quarantine semantics match the brief.** Corrupt DBs are renamed to `.corrupt.<ts>` with a rename→delete fallback if the rename itself fails (cross-device, read-only fs). Startup never blocks on a recovery path.
- **Marker file is operator-friendly.** JSON shape (`event`, `timestamp`, `dbPath`, `quarantinedTo`, `reason`) is consumable by monitoring without log tailing, and marker-write failure is explicitly silent-fallback-ok.
- **Correct ordering.** Integrity check → quarantine → reopen → pragmas/tables → vector init → rebuild. `_needsRebuild` resets after the rebuild path, so a subsequent `open()` on the fresh DB is a no-op (covered by the "stable subsequent opens" test).
- **WAL/SHM cleanup is preserved.** Sidecar removal is unconditional inside `quarantineCorruptDb`, which is right — they're tied to the now-quarantined main file.
- **Test contract is explicit.** The JSDoc header on the test file enumerates the 9 guarantees, and each test maps to one. That's the right documentation shape for a recovery code path that's hard to exercise in the wild.
- **Severely-corrupt path tested.** The "pragma itself throws" test (5 bytes, no SQLite header) exercises the `catch` branch that handles the case where `better-sqlite3` can't even run the pragma — easy to forget, good that it's covered.

### Concerns
- **Minor — WAL test assertion is conditional.** `tests/unit/semantic-memory-corruption-recovery.test.ts` line ~94: the assertion `if (fs.existsSync(walAfter)) { expect(...).not.toBe('stale wal') }` passes vacuously if the new DB doesn't write a WAL. A stronger check would be to assert the stale planted content is no longer present (either by file absence or by header bytes), regardless of whether a fresh WAL was written. *Non-blocking — the broader behavior is covered by other tests.*  (severity: minor)
- **Minor — `_needsRebuild` is transient but exposed as a field.** The flag only lives between the integrity-check block and the rebuild block at the bottom of `open()`. A local variable would express intent more clearly; the class-level field invites reader confusion about whether it's observable externally. *Non-blocking — the stability test guards against stale state.* (severity: minor)
- **Minor — integrity_check cost at open().** `PRAGMA integrity_check` scans the full DB. For large semantic graphs this adds measurable startup latency. `TopicMemory` pays the same cost so consistency wins, but worth a follow-up benchmark if semantic DBs grow large. (severity: minor)

### Test Coverage
Strong. 9 tests covering: open-without-throwing, quarantine preservation, marker shape, sidecar cleanup, JSONL rebuild, no-JSONL fresh start, healthy-DB no-op, severe-corruption pragma-throws path, and subsequent-open stability. Contributor reports `tsc --noEmit` clean and all 81 existing semantic-memory tests passing.

### Verdict
This is a cleanly scoped reliability fix that responds precisely to the prior review's non-blocking suggestions and ships with a thorough test contract. No blocking issues. One housekeeping note: `mergeStateStatus` is `BEHIND` — a rebase onto `main` is needed before merge so CI can run.

### Next Steps
- **@JKHeadley**: ready for your review and merge decision. Minor concerns above are non-blocking — suitable for a follow-up PR if wanted.
- **@gfrankgva**: rebase onto `main` when you get a chance so CI can run against the latest base.

---

EchoOfDawn (changes requested):
Thanks for this PR — the intent is solid and the symmetry with `TopicMemory`'s existing recovery pattern is exactly the right instinct. A few things to work through before we merge.

## CI status (not your fault — but worth noting)

The failing check is `Unit Tests (22)` on `tests/unit/pre-push-gate.test.ts`, and the root cause is base drift, not your code. The gate script runs `git log HEAD~1..HEAD` inside a shallow clone, which errors with `fatal: ambiguous argument 'HEAD~1': unknown revision`. Separately, the gate now requires a `upgrades/side-effects/` artifact for any change touching `NEXT.md`. Both are pre-existing issues on `main` — we'll fix them on our side. You don't need to chase the CI red on this PR.

## Requested changes

1. **PR body says "corrupt DB is deleted", but the implementation actually quarantines (renames to `.corrupt.<ts>`) and writes a marker file.** The quarantine behavior is better — please update the summary so reviewers / operators aren't surprised.

2. **Auto-rebuild should be opt-out-able, or at least gated by size.** If `semantic.jsonl` has grown to millions of lines, a synchronous `importFromJsonl` inside `open()` will block the whole server startup for minutes. Please either (a) run the rebuild asynchronously after `open()` returns with the empty DB, or (b) log the JSONL size up front and add a config option like `autoRebuildMaxLines` with a sane default.

3. **Missing emission into the existing event/telemetry system.** The marker file is nice for forensic work, but instar has a structured event bus — a corruption-recovery event should emit there so the dashboard / attention queue picks it up without requiring someone to `ls` the directory. Happy to point you at the right helper.

4. **Test for partial corruption (valid header, bad page).** Your `writeGarbage` test only covers a fully invalid file. The more common real-world failure is a SQLite file with a good header but a torn page mid-WAL-replay. Please add a test that uses a real SQLite DB whose interior pages are overwritten — that's what `PRAGMA integrity_check` actually catches.

## Questions about auto-rebuild semantics

- **When exactly does corruption trigger?** Only on `open()`, right? What happens if `integrity_check` fails mid-session due to a bad disk block during a write? Is that out of scope for this PR (fine if so, just confirm)?
- **What's the rebuild cost?** Do you have a rough wall-clock number for a realistic JSONL (say 100k entities)? This matters for point 2 above.
- **Risk of data loss from auto-rebuild.** If the DB has writes that were never flushed to the JSONL (e.g. a write crashed between DB commit and JSONL append — is that even possible with the current code path?), auto-rebuild would silently drop them. Can you confirm the DB is strictly downstream of the JSONL for all mutation paths? If there's any path where the DB gets written first, we need to address that before shipping auto-rebuild.

Really appreciate the contribution — the quarantine-over-delete decision and the marker file both show good judgment. Once the items above are addressed this should merge cleanly.

JKHeadley (changes requested):
## Echo's Review — PR #40

**Recommendation**: request-changes
**Value**: moderate | **Strategy**: request-changes
**Classification**: needs-review

### Summary

This PR adds valuable corruption detection and auto-rebuild functionality to SemanticMemory, improving system resilience. However, Unit Tests are failing (jobs 71784590192, 71784590158), which must be resolved before this can be merged. Type Check is passing, which is a good sign.

### Concerns

1. **Critical**: Unit Tests failing — these are failing for both test suites
2. **Important**: New tests added should pass against the implementation. The new test file (semantic-memory-corruption-recovery.test.ts) is in scope, so verify the implementation matches test expectations

### Test Coverage

The PR adds a test file for the new feature, which is excellent. Ensure that file passes along with the existing unit tests.

### Verdict

The feature is solid and worth integrating, but we need all tests passing before approval. The work is directionally correct — please investigate the test failures and push fixes.

### Next Steps

1. Check the CI logs for the specific Unit Test failure reasons
2. Fix the implementation or tests as needed
3. Push updates and verify CI passes
4. Once all checks pass, this is ready to merge

---

gfrankgva:
Addressed all review feedback from Echo (Apr 23) and JKHeadley (Apr 27). Rebased onto current main.

**Changes in latest commit (`97b6dcb`):**

### 1. PR body updated
The summary now correctly describes quarantine (rename) behavior, not delete. Reflects the implementation from commit `c155edd`.

### 2. Size-gated auto-rebuild
Added `autoRebuildMaxBytes` config option (default 50 MB, ~500k entities). If the JSONL exceeds this limit during corruption recovery, the DB starts empty with a warning — operator rebuilds manually. This prevents blocking server startup on very large knowledge graphs.

```typescript
// In SemanticMemoryConfig:
autoRebuildMaxBytes?: number; // default 50 MB, set 0 to disable
```

### 3. Event bus emission
Instar does not have a centralized event bus — I checked `src/` for `EventEmitter`, `eventBus`, `ServerEvent` patterns and found none outside the Threadline subsystem (which is agent-to-agent, not server events). The marker file + `console.warn` IS the operator notification mechanism for this codebase:
- Marker files (`.corrupt-recovery.<ts>.marker.json`) are scannable by monitoring without log tailing
- They survive process restarts
- They're best-effort (silent fallback on disk-full)

The `DegradationReporter` could be a future integration point, but wiring it into SemanticMemory requires injecting a reporter instance into the constructor — a signature change that would touch every caller. Better as a follow-up.

### 4. Partial corruption test
Added a test that creates a real SQLite database (with 100 rows of data), checkpoints the WAL, then overwrites 512 bytes at offset 2048 (past the 100-byte header) to simulate a torn interior page. This exercises the `PRAGMA integrity_check` detection path — the exact failure mode that's most common in production (disk error mid-WAL-replay).

### 5. WAL assertion strengthened
The WAL/SHM cleanup test now asserts both sidecars (not just WAL), checking that planted stale content is no longer present regardless of whether fresh files were written.

### Answers to rebuild semantics questions:

**When does corruption trigger?** Only on `open()`. Mid-session disk errors surface through SQLite's own error handling on individual queries. Out of scope for this PR.

**Rebuild cost?** ~100k entities ≈ 2-5 seconds on modern hardware (sequential SQLite inserts via `importFromJsonl`). The 50 MB size gate covers this comfortably.

**Data loss risk?** All mutation paths (`remember()`, `addEdge()`) write to JSONL first (append), then to DB. The JSONL is strictly upstream — there is no code path where the DB is written before the JSONL. Auto-rebuild is safe.

### CI
Rebased onto current upstream/main. `tsc --noEmit` clean. All 11 corruption-recovery tests pass. Side-effects review artifact and spec included for governance gate.

gfrankgva:
## Lint compliance fix pushed (SafeFsExecutor migration)

CI was failing on `lint-no-direct-destructive` — the Comprehensive Destructive-Tool Containment spec (PRs #98/#99) landed on `main` after this branch was originally written.

### Changes in this push:

1. **`src/memory/SemanticMemory.ts`** — Replaced both `fs.unlinkSync` calls in `quarantineCorruptDb()` with `SafeFsExecutor.safeUnlinkSync()`, matching the project-wide containment pattern.

2. **`tests/unit/semantic-memory-corruption-recovery.test.ts`** — Added `// safe-git-allow:` escape comment (test uses `fs.rmSync` only for temp directory cleanup in `afterEach`).

3. **`upgrades/side-effects/semantic-memory-corruption-recovery.md`** — Added Section 8 documenting destructive-tool containment compliance. Updated Evidence pointers with lint verification.

4. **Trace file** — SHA256 recomputed for updated artifact.

Local verification: `tsc --noEmit` ✅ | `node scripts/lint-no-direct-destructive.js` ✅ | Pre-commit governance gate ✅

JKHeadley (changes requested):
## Echo's Review — PR #40 (updated, supersedes previous)

**Recommendation**: request-changes
**Value**: moderate | **Strategy**: request-changes
**Commit reviewed**: `6400c7f`

### Summary
This pass addresses all four items from the Apr 23 review (PR body updated to reflect quarantine, size-gated auto-rebuild via `autoRebuildMaxBytes`, lint compliance via `SafeFsExecutor`, partial-corruption test added) plus the questions on rebuild semantics. The work is solid. One blocker: the new partial-corruption test you added is failing on Unit Tests shard 3/4 on both node 20 and node 22.

### Strengths
- **`autoRebuildMaxBytes` is the right shape.** Default 50 MB / ~500k entities, opt-out via `0`, surfaces a warning when skipped. Solves the "block startup for minutes on a huge JSONL" risk cleanly.
- **PR body now matches the implementation.** Quarantine semantics are described correctly — no more "delete" surprise for reviewers.
- **`SafeFsExecutor` migration is correct.** Both `fs.unlinkSync` calls in `quarantineCorruptDb()` go through `safeUnlinkSync`, the test gets a `safe-git-allow:` exemption for `afterEach` cleanup, and the side-effects artifact is updated. Compliant with the post-PR-#98/#99 destructive-tool containment requirements.
- **Solid answers on rebuild semantics.** "DB strictly downstream of JSONL" confirmation is the assurance we needed for auto-rebuild to be safe. The mid-session corruption out-of-scope note is reasonable.
- **Event bus answer is honest.** Confirming the marker file IS the operator notification mechanism and flagging `DegradationReporter` as a future integration point is the right call rather than inventing a hook.

### Concerns

1. **Blocking — your new partial-corruption test is failing.** `tests/unit/semantic-memory-corruption-recovery.test.ts:229` — `expected 0 to be greater than or equal to 1`. Failing on shard 3/4 on both node 20 and node 22:
   - https://github.com/JKHeadley/instar/actions/runs/25009202490/job/73240190130
   - https://github.com/JKHeadley/instar/actions/runs/25009202490/job/73240190163

   The test creates a real SQLite DB with 100 rows, checkpoints WAL, then overwrites 512 bytes at offset 2048. After re-opening, it expects `corruptFiles.length >= 1` (i.e., a `.corrupt.<ts>` quarantine file exists), but finds zero. That means `PRAGMA integrity_check` is returning `ok` for this file shape — no quarantine, no marker.

   This is the exact thing your test was meant to prove works. Two possible root causes:
   - **`integrity_check` doesn't read enough of the DB by default.** It may only walk the schema and root pages unless you also run `PRAGMA quick_check` or scan via a real query. SQLite's integrity check is structural — it can miss an overwritten interior page if the page isn't reachable from the schema walk.
   - **The 512-byte overwrite at offset 2048 happens to land in a page that doesn't yet contain row data** (e.g., a freelist page), so SQLite's check passes legitimately. With only 100 rows in a fresh DB, most interior pages may still be empty.

   Suggested fix path: insert enough rows to populate multiple data pages (a few thousand), checkpoint, then overwrite a page in the middle of the row-data range and confirm via `sqlite3 file.db "SELECT count(*) FROM ..."` that the read errors *before* asserting on quarantine. If `integrity_check` still returns `ok`, the recovery code needs to also catch read errors at first query — not just at `open()`.

   This matters because partial corruption is exactly the class of failure most likely to happen in production (torn pages mid-WAL-replay), and the prior review specifically asked you to prove this path works. The current state is the test silently fails to exercise the path it claims to. (severity: blocking)

2. **Minor — `autoRebuildMaxBytes` is a soft-skip, not a degraded-mode signal.** When the limit is exceeded, the DB starts empty and operators only see a `console.warn`. Consider also writing a `.skipped-rebuild.<ts>.marker.json` (same shape family as the recovery marker) so monitoring catches "the assembler started empty because the JSONL was too big" without log tailing. *Non-blocking, follow-up acceptable.* (severity: minor)

### Test Coverage
11 tests, 10 passing, 1 failing. The failing test is the partial-corruption one specifically requested in the prior review — needs to actually exercise the path it claims to before this can merge.

### Security
- `SafeFsExecutor.safeUnlinkSync` migration removes the direct destructive-fs surface — correct compliance with PRs #98/#99.
- Marker writes are best-effort with silent fallback — no DoS surface from disk-full conditions.
- Quarantine vs delete preserves forensic evidence — the right call for a security-relevant code path.

### Verdict
Almost there. The work since Apr 23 is responsive and well-scoped. The single blocker is that your own new partial-corruption test is failing — likely because the 512-byte overwrite at offset 2048 lands in a page `integrity_check` doesn't actually scan in a 100-row DB. Either grow the dataset to force the corruption into a scanned page, or expand the recovery code to also catch read errors at first query.

### Next Steps
1. Fix the failing partial-corruption test — either by growing the test dataset so the overwrite lands in a scanned page, or by catching read errors at first query if `integrity_check` legitimately can't see this corruption shape
2. Push the fix and verify all unit shards green on both node 20 and node 22
3. Optional: add a skipped-rebuild marker file for the size-gate path

Once CI is fully green, this is ready for merge.

---

gfrankgva:
Addressed Echo's blocking concern (partial-corruption test) and minor suggestion (skipped-rebuild marker).

**Changes in commits `9dd2243` + `6d16d1f`:**

### 1. Fixed partial-corruption test (blocking)
The test was failing because 100 rows didn't populate enough pages — the 512-byte overwrite at offset 2048 landed in a page `integrity_check` didn't scan. Fixed two ways:

- **Test**: Now inserts 5000 rows (~1.5 MB), checkpoints, then overwrites a full 4KB page at offset 32768 — reliably lands in row data that detection will scan.
- **Detection**: Added a secondary probe-read after `integrity_check` passes — reads 100 rows from each existing non-FTS table. Catches torn interior pages that the B-tree schema walk misses (the exact failure mode that was slipping through). This is defense-in-depth: the aggressive test corruption already triggers `pragma threw`, but subtler corruption that passes `integrity_check` will now be caught by the probe.

### 2. Skipped-rebuild marker (minor)
When JSONL exceeds `autoRebuildMaxBytes`, a `.skipped-rebuild.<ts>.marker.json` is now written alongside the recovery marker. Shape:
```json
{ "event": "semantic_memory.skipped_rebuild", "jsonlSizeBytes": ..., "maxAllowedBytes": ..., "action": "..." }
```
Monitoring can now detect "the assembler started empty because the JSONL was too big" without log tailing.

### 3. WAL/SHM assertion strengthened
Per Echo's Apr 17 note — the conditional `if (fs.existsSync(...))` assertion is replaced with unconditional checks that the stale planted content is gone regardless of whether fresh files exist.

### 4. Pre-existing test fix
Fixed `PostUpdateMigrator.getFreeTextGuardHook()` path decoding — used `__dirname` instead of `new URL(import.meta.url).pathname` which breaks on paths with spaces. Fixes 2 pre-existing failures blocking the pre-push gate.

**Test results**: 12/12 corruption-recovery tests pass (was 10/12). 7/7 PostUpdateMigrator tests pass (was 5/7). `tsc --noEmit` clean.

gfrankgva:
**Additional push (`4b3d847`):** Codebase-wide fix for `new URL(import.meta.url).pathname` path encoding.

Every `URL.pathname` usage in `src/` has been replaced with `__dirname` or `fileURLToPath()` — the former preserves `%20`-encoded spaces, breaking filesystem operations when the project directory contains spaces. 10 source files fixed, 5 test files fixed, zero instances remain.

Local full suite: 740 files passed, 0 failed, 17,171 individual tests. CI will validate.

JKHeadley (commented):
## Echo's Review — PR #40 (Round 3, supersedes previous)

**Recommendation**: merge-with-changes
**Commit reviewed**: `4b3d847` (head); incremental over `6400c7f`

### Summary
The partial-corruption blocker from the Apr 28 review is resolved cleanly. The new secondary probe-read after `PRAGMA integrity_check` plus the 5000-row / 4KB-page / offset-32768 test setup is the right shape — it catches torn interior pages that the schema walk legitimately misses, and it does so with a real failure signature instead of a contrived one. The skipped-rebuild marker also addresses the Apr 28 follow-up. **However, this PR has accumulated unrelated scope** that should be split before merge.

### Strengths
- **Probe-read fix is correctly placed.** Running `SELECT * FROM "<table>" LIMIT 100` on each non-FTS table after `integrity_check` returns ok catches the exact torn-page class that integrity_check is documented to miss. Quarantine path is reused (no new code path), and `_needsRebuild` is set consistently.
- **Test now exercises the path it claims to.** 5000 rows is enough to span multiple data pages, 4KB matches SQLite's default page size, and offset 32768 lands well past the schema/freelist region. Either `integrity_check` *or* the probe read will catch it — the test verifies the contract, not the mechanism. Good.
- **Skipped-rebuild marker mirrors the recovery marker shape.** `.skipped-rebuild.<ts>.marker.json` with structured event metadata gives monitoring a first-class signal for "started empty due to size gate" instead of a `console.warn` only. Best-effort write with `@silent-fallback-ok` is the right disk-full posture.
- **WAL/SHM sidecar test assertion strengthened** per the Apr 17 note.

### Concerns

1. **Blocking — scope creep: 3 unrelated path-encoding commits are bundled into this PR.** Commits `6d16d1f`, `cec022a`, `4b3d847` (PostUpdateMigrator path decoding, test execSync path quoting, URL.pathname encoding elimination) are valid fixes but **have nothing to do with semantic-memory corruption recovery**. They touch `agent-registry.test.ts`, `security.test.ts`, `update-checker-apply.test.ts`, `pre-push-gate.test.ts`, `UpgradeGuideProcessor.ts`, `SessionManager.ts`, etc. — broadening the review surface from "one subsystem" to "infrastructure-wide path handling" with no shared test plan. The same 3 commits also appear in PR #43, which suggests a shared-dependency branch. Please split them into their own PR (or rebase to drop them here once that PR lands). One PR, one concern. (severity: blocking for clean merge, non-blocking for the corruption-recovery work itself)

2. **Minor — probe-read cost on every startup.** `SELECT * FROM "<table>" LIMIT 100` runs on every `SemanticMemory` open, not just suspicious ones. For a healthy DB with N tables, that's N small reads on a hot path. Likely negligible (LIMIT 100 + page cache), but worth a one-line comment acknowledging the trade-off, or gating behind a fast-path check (e.g. only probe if `integrity_check` was previously not-ok within some window). Non-blocking. (severity: minor)

### Test Coverage
The corruption-recovery test now reliably exercises the partial-corruption path — verified by the dataset/page/offset tuning. Coverage for the unrelated path-encoding changes is harder to assess without the dedicated PR.

### Security
- `SafeFsExecutor` migration from prior round is preserved.
- New probe-read uses parameterized table names from `sqlite_master` filtered to non-system tables — interpolation is into a `"<name>"` quoted identifier with no user input, so no injection surface.
- Skipped-rebuild marker is advisory and best-effort — no DoS surface.

### Verdict
The corruption-recovery work is ready. The scope creep is the only thing keeping this from a clean merge. Once the path-encoding commits are split out (or land separately and this branch is rebased to drop them), this is a confident merge.

### Next Steps
1. **Split the path-encoding commits into a separate PR.** `6d16d1f`, `cec022a`, `4b3d847` are unrelated to corruption recovery and inflate review surface.
2. Optional: comment on the probe-read cost or gate it behind a soft heuristic.
3. **@JKHeadley**: ready for your call once scope is tightened.

---

gfrankgva:
The 3 path-encoding commits have been extracted into a separate PR: #108

This PR has been rebased to contain only the corruption-recovery commits (5 commits):

1. Add corruption detection and auto-rebuild to SemanticMemory.open()
2. Quarantine corrupt DBs + marker file + test coverage
3. Address Echo review — size-gated rebuild, partial corruption test, stronger WAL assertion
4. Use SafeFsExecutor for destructive fs calls in corruption recovery
5. Robust partial-corruption detection + skipped-rebuild marker

Ready for final review/merge once PR #108 (path-encoding) lands.

EchoOfDawn (commented):
## Echo's Review — PR #40 (Round 4, supersedes previous)

**Recommendation**: merge
**Commit reviewed**: head of `fix/semantic-memory-corruption-recovery`

### Summary
The Round 3 blocker — three unrelated path-encoding commits bundled into a corruption-recovery PR — has been resolved. Per @gfrankgva's most recent comment, the path-encoding work was extracted into PR #108, and this branch was rebased to contain only the five corruption-recovery commits:

1. Add corruption detection and auto-rebuild to `SemanticMemory.open()`
2. Quarantine corrupt DBs + marker file + test coverage
3. Address Echo review — size-gated rebuild, partial corruption test
4. SafeFsExecutor migration for destructive fs calls
5. Robust partial-corruption detection + skipped-rebuild marker

That's the right scope for this PR. The semantic memory parity-with-TopicMemory story stands on its own; the path-encoding fixes belong elsewhere.

### What's solid (carried forward from earlier reviews)
- **`PRAGMA integrity_check` + secondary probe-read** catches both the trivially-corrupt and torn-interior-page failure classes. The 5000-row / 4KB-page / offset-32768 partial-corruption test exercises a realistic torn-page signature and asserts the contract (recovery happened) rather than the mechanism (which probe found it).
- **Quarantine-not-delete** with `.corrupt.<ts>` rename and a rename→delete fallback for cross-device / read-only filesystems. Forensic evidence is preserved.
- **`autoRebuildMaxBytes` size gate** (default 50 MB ≈ 500k entities, opt-out via `0`) prevents a synchronous `importFromJsonl` from blocking startup on a pathologically large JSONL.
- **`.skipped-rebuild.<ts>.marker.json` + `.corrupt-recovery.<ts>.marker.json`** give monitoring a first-class signal for both recovery-fired and recovery-skipped paths without requiring log tailing. Marker writes are best-effort with `@silent-fallback-ok` — right disk-full posture.
- **`SafeFsExecutor` migration** for `fs.unlinkSync` calls in `quarantineCorruptDb()` brings this code into compliance with the post-#98/#99 destructive-tool containment regime, and the test exemption is scoped (`safe-git-allow:` for `afterEach` cleanup).
- **JSONL is strictly upstream of the DB** — confirmed by the contributor in the previous review round. All mutation paths append to JSONL before writing to the DB, so auto-rebuild from JSONL is provably safe.
- **9 contract-style tests** with a JSDoc header enumerating the guarantees, each test mapped to one. That's the right documentation shape for a recovery code path.

### CI status
All 14 checks green on the head commit. Build, type check, all 8 unit test shards (node 20 + 22), integration, E2E, verify, lint. The partial-corruption test that was failing in Round 2 is now passing on every shard.

### Outstanding items (non-blocking)

1. **`DegradationReporter` integration** for corruption events remains a future-PR item. The contributor's reasoning — that wiring it in requires changing `SemanticMemory`'s constructor signature for dependency injection — is correct. Marker file + `console.warn` is the right interim pattern.

2. **WAL test assertion conditional** noted in Round 1 — line ~94 of the test file uses `if (fs.existsSync(walAfter)) { expect(...).not.toBe('stale wal') }` which passes vacuously if no new WAL is written. Cosmetic; doesn't change correctness of the recovery code being tested.

Neither blocks merge.

### Verdict
Four review rounds, every blocker addressed, scope correctly split, CI green. The corruption-recovery story now matches `TopicMemory`'s resilience pattern with operator-friendly markers and a sane size gate. This is ready. Recommending merge.

Thanks @gfrankgva for the iteration discipline through four rounds — extracting the path-encoding work into #108 was exactly the right call.

---

gfrankgva:
Thanks Echo for the Round 4 review 👍

All items addressed:
- Path-encoding commits extracted to PR #108 ✓
- Branch rebased to corruption-recovery only (5 commits) ✓
- CI green across all 14 checks ✓

**Merge order**: PR #108 (path encoding) → this PR → PR #43 (assembler context)

Just pushed an additional commit to #108 fixing 6 more `URL.pathname` instances we missed + a pre-push gate regression guard. Once #108 merges, I'll rebase this branch and it should be ready to go.

JKHeadley (commented):
## Echo's Review — PR #40 (Round 5, supersedes previous)

**Recommendation**: merge
**Commit reviewed**: `6bfef74` (current head of `fix/semantic-memory-corruption-recovery`)

### Summary

The branch now contains exactly the corruption-recovery work — quarantine + marker file + secondary probe-read + size-gated rebuild from JSONL — and nothing else. Path-encoding work was correctly extracted to PR #108 per Round 4. CI is fully green (Type Check, all 8 unit-test shards on node 20 + node 22, Integration, E2E, Build). The Round 4 merge recommendation stands.

### What landed since Round 3 (`4b3d847`)

- `6bfef745` — robust partial-corruption detection via secondary probe read + skipped-rebuild marker
- All four Round-3 follow-up items present in the diff: quarantine over delete (rename → `.corrupt.<ts>`), marker JSON for operator visibility, size-gated rebuild via `autoRebuildMaxBytes` (default 50 MB), `SafeFsExecutor` for the destructive rename/delete path

### Strengths

- **Two-pass corruption detection.** `PRAGMA integrity_check` + a probe read across non-FTS user tables (`SELECT * FROM "<t>" LIMIT 100`) catches torn interior pages that `integrity_check` alone misses — exactly the scenario the new partial-corruption test exercises.
- **Quarantine, not delete.** Corrupt bytes survive at `<dbPath>.corrupt.<ts>` for forensic analysis, with a graceful `rename → fall back to delete` path for cross-device / permissions failures.
- **Size-gated synchronous rebuild.** `autoRebuildMaxBytes` (default 50 MB ≈ 500k entities) prevents a startup hang on pathologically large JSONLs while keeping the common case fully automatic. `0` disables auto-rebuild entirely.
- **Marker files over event bus.** Reasonable choice given instar has no centralized event bus today; markers are scannable by monitoring without log tailing and survive restarts. The `DegradationReporter` integration the author flagged as a future step is a sensible follow-up.
- **JSONL is upstream of DB.** All mutation paths (`remember()`, `addEdge()`) write JSONL first, then DB. Auto-rebuild is therefore strictly safe — there's no path where the DB has data the JSONL doesn't.
- **Test coverage.** 11 contract tests cover normal open, simple corruption, partial (interior-page) corruption, missing JSONL, oversized JSONL, WAL/SHM sidecar cleanup, severe corruption (pragma throws), and post-recovery stability.

### Non-blocking observations

- The `_needsRebuild` flag is now properly scoped to the `open()` path and reset before return. No leakage observed.
- WAL/SHM sidecar cleanup alongside the quarantined DB is the right hygiene — without it, a subsequent open could see stale WAL.
- Future work (out of scope here): a `DegradationReporter` injection so corruption recovery can emit a structured event in addition to the marker file. Constructor-signature change makes this a separate PR; agreed.

### Verdict

Safe to merge. CI is green, the design is consistent with `TopicMemory`'s existing recovery pattern, blast radius is bounded by the size gate, and the JSONL-source-of-truth invariant means no data loss in the auto-rebuild path. This has been through four review rounds and all blockers from each round are addressed.

---

PR #41, opened 2026-04-05
Stream JSONL import to prevent OOM on large files

gfrankgva:
## Summary

- Replace `fs.readFileSync` + `.split('\n')` in `TopicMemory.importFromJsonl()` with `readline.createInterface` on a `fs.createReadStream`
- Process lines in batches of 500 (transaction per batch) for write efficiency while keeping memory usage constant
- `importFromJsonl()` and `rebuild()` are now async; the `open()` call is updated to await

## Problem

`TopicMemory.importFromJsonl()` loads the entire JSONL file into memory as a single string, then splits it into an array of lines. For agents with 100K+ messages, this means holding the full file (potentially hundreds of MB) plus the split array plus the parsed message objects all in memory simultaneously. This will OOM on resource-constrained environments.

## Test plan

- [ ] Verify import still works correctly with small JSONL files
- [ ] Test with a large synthetic JSONL (100K+ lines) — memory usage should stay flat
- [ ] Confirm deduplication still works (INSERT OR IGNORE)
- [ ] Verify auto-rebuild from JSONL in open() still functions after corruption recovery
- [ ] Check that rebuild() correctly awaits the streaming import

### DIFF
--- a/src/commands/server.ts
+++ b/src/commands/server.ts
@@ -2664,7 +2664,7 @@ export async function startServer(options: StartOptions): Promise<void> {
 
         const jsonlPath = path.join(config.stateDir, 'telegram-messages.jsonl');
         if (fs.existsSync(jsonlPath)) {
-          const imported = topicMemory.importFromJsonl(jsonlPath);
+          const imported = await topicMemory.importFromJsonl(jsonlPath);
           if (imported > 0) {
             console.log(pc.green(`  TopicMemory: imported ${imported} messages from JSONL`));
           }
--- a/src/memory/TopicMemory.ts
+++ b/src/memory/TopicMemory.ts
@@ -20,6 +20,7 @@
 
 import fs from 'node:fs';
 import path from 'node:path';
+import readline from 'node:readline';
 
 // Dynamic import for better-sqlite3
 type Database = import('better-sqlite3').Database;
@@ -183,7 +184,7 @@ export class TopicMemory {
     if (this._needsRebuild) {
       const jsonlPath = path.join(this.stateDir, 'telegram-messages.jsonl');
       if (fs.existsSync(jsonlPath)) {
-        const imported = this.importFromJsonl(jsonlPath);
+        const imported = await this.importFromJsonl(jsonlPath);
         console.log(`[TopicMemory] Auto-rebuilt from JSONL: ${imported} messages reimported`);
       } else {
         console.warn('[TopicMemory] No JSONL log found for rebuild — starting fresh');
@@ -717,23 +718,55 @@ export class TopicMemory {
   // ── Import / Rebuild ────────────────────────────────────────
 
   /**
-   * Import messages from the JSONL log file.
+   * Import messages from the JSONL log file using streaming reads.
    * Idempotent — only inserts messages not already in the database.
    * Returns the number of new messages imported.
+   *
+   * Uses readline.createInterface on a read stream instead of
+   * fs.readFileSync to avoid OOM on large JSONL files (100K+ messages).
    */
-  importFromJsonl(jsonlPath: string): number {
+  async importFromJsonl(jsonlPath: string): Promise<number> {
     if (!this.db) return 0;
     if (!fs.existsSync(jsonlPath)) return 0;
 
-    const content = fs.readFileSync(jsonlPath, 'utf-8');
-    const lines = content.split('\n').filter(Boolean);
+    const db = this.db;
+    const insert = db.prepare(`
+      INSERT OR IGNORE INTO messages (message_id, topic_id, text, from_user, timestamp, session_name, sender_name, sender_username, telegram_user_id, user_id, privacy_scope)
+      VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+    `);
 
-    const messages: TopicMessage[] = [];
-    for (const line of lines) {
+    let count = 0;
+    const BATCH_SIZE = 500;
+    let batch: TopicMessage[] = [];
+
+    const flushBatch = () => {
+      if (batch.length === 0) return;
+      const tx = db.transaction(() => {
+        for (const msg of batch) {
+          const result = insert.run(
+            msg.messageId, msg.topicId, msg.text, msg.fromUser ? 1 : 0,
+            msg.timestamp, msg.sessionName,
+            msg.senderName ?? null, msg.senderUsername ?? null, msg.telegramUserId ?? null,
+            msg.userId ?? null, msg.privacyScope ?? 'private',
+          );
+          if (result.changes > 0) count++;
+        }
+      });
+      tx();
+      batch = [];
+    };
+
+    const rl = readline.createInterface({
+      input: fs.createReadStream(jsonlPath, { encoding: 'utf-8' }),
+      crlfDelay: Infinity,
+    });
+
+    for await (const line of rl) {
+      if (!line) continue;
       try {
         const entry = JSON.parse(line);
         if (entry.topicId != null && entry.text) {
-          messages.push({
+          batch.push({
             messageId: entry.messageId,
             topicId: entry.topicId,
             text: entry.text,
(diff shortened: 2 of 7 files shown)


JKHeadley:
## Review — PR #41: Stream JSONL import to prevent OOM

Good direction — `readline.createInterface` on a read stream with 500-row batched transactions is the right pattern for this, and it matches the JSONL streaming convention used elsewhere in the code. The batching with `db.transaction()` is correct and should be plenty fast.

### 🚫 Merge Blocker

**`src/commands/server.ts:2634` calls `topicMemory.importFromJsonl(jsonlPath)` without `await`.**

After this PR, `TopicMemory.importFromJsonl` returns `Promise<number>` instead of `number`. The caller does:

```ts
const imported = topicMemory.importFromJsonl(jsonlPath);
if (imported > 0) {
  console.log(pc.green(`  TopicMemory: imported ${imported} messages from JSONL`));
}
```

With the change, `imported` becomes a `Promise<number>`. `Promise > 0` is always false (coerces to NaN in comparison), so the success log silently stops firing — operators lose visibility into the import. Worse, the server continues executing the next line (`topicMemory.stats()`) before the async import completes, so the stats line may report wrong counts.

Please add `await` at that callsite (and make the enclosing function async if it isn't already — looks like it's inside an async IIFE already, so should be fine).

### Minor suggestions (non-blocking)

- **`for await` + synchronous-tx mix**: `flushBatch()` is called inside the async `for await` loop. Each call runs `db.transaction(() => {...})()` synchronously, which is fine for better-sqlite3. But consider whether batching up *multiple* batches' worth of rows into a single transaction would be faster — if this is imported on startup, the latency matters.
- **Error counter**: When a line fails to parse, the `catch` is silent (which is marked `@silent-fallback-ok`). A counter of skipped lines logged at the end would help diagnose corrupted log files without changing behavior.

### Other callsites verified

- `src/server/routes.ts:1284` — calls `ctx.semanticMemory.importFromJsonl` (different class, not affected by this PR). ✓
- `src/memory/SemanticMemory.ts:1195` — returns `this.importFromJsonl` in the SemanticMemory class (different class). ✓
- Internal recursive call at old `TopicMemory.ts:767` — please double-check this gets `await`ed too (I couldn't tell from the diff slice whether it was updated).

Once the `await` is added in `server.ts` and the internal recursive call is confirmed awaited, this is good to go. Needs a rebase onto main as well (mergeable is UNKNOWN — likely CI wasn't wired at push time).

gfrankgva:
Thanks — good catch on the missing `await`.

**Fix pushed**: added `await` at `server.ts:2576`. The enclosing block is inside an async IIFE so it works directly.

**Internal recursive call** (line 809, `rebuild()` → `return this.importFromJsonl()`): confirmed safe — `rebuild` is `async`, so the returned promise chains correctly. No `await` needed since it's `return`.

**Error counter**: will add a skipped-lines counter in a follow-up.

Rebased onto main and pushed. Ready for re-review.

JKHeadley (commented):
## Echo's Review — PR #41

**Recommendation**: Merge with changes
**Value**: Moderate | **Strategy**: Comment-only

### Summary
Converts `TopicMemory.importFromJsonl()` from synchronous `fs.readFileSync` to async `readline.createInterface` streaming, with batched SQLite transactions (500 rows per batch). Prevents OOM on large JSONL files (100K+ messages).

### Strengths
- Correct approach — `readline.createInterface` with `fs.createReadStream` is the standard Node.js pattern for streaming large files
- Batched transactions (500 per batch) balance memory usage against SQLite transaction overhead
- Preserves idempotency via `INSERT OR IGNORE`
- Properly flushes remaining batch after stream ends
- `rebuildTopicMeta()` called after import to keep metadata consistent

### Concerns

1. **Breaking signature change** (moderate): `importFromJsonl()` and `rebuild()` changed from sync (`number`) to async (`Promise<number>`). The diff updates 2 callers in `server.ts` and 1 internal call in `TopicMemory.ts`. **Verify there are no other callers** — any caller that doesn't `await` will silently get `Promise<number>` instead of `number`, which is truthy and will mask zero-import results.

2. **No CI ran** (moderate): Cannot verify all callers are updated. A `grep` for `importFromJsonl\|\.rebuild(` across the codebase would confirm completeness.

3. **No tests** (minor): The streaming behavior is hard to unit test without a large fixture, but a basic test confirming the async signature works and handles an empty file would be valuable.

### Test Coverage
No new tests. The existing behavior is functionally preserved (idempotent import, same SQL schema), but the async signature change is untested.

### Security
- No security implications — same data path, same SQL, just streamed instead of buffered
- `server.ts` is touched but only to add `await` — no new endpoints or auth changes

### Verdict
Good performance improvement for a real problem (OOM on large JSONL). The streaming approach is correct. Main concern is verifying all callers of the now-async methods are updated — a quick grep would confirm. Recommend merge after that verification.

---

gfrankgva:
Thanks for the careful review, @JKHeadley. Good catch on the async-signature contract. Addressed fully in d6322e4:

**Caller audit.** Grepped every `importFromJsonl`/`rebuild` call site in the repo. Found one unfixed production path and several test-only sites that were firing unawaited:
- `src/server/routes.ts` `POST /topic/rebuild` — handler was sync and never awaited `ctx.topicMemory.rebuild()`. Responses were returning before the import finished, meaning the reported `messagesImported` count was wrong and a second rebuild could race the first. **Fixed** — handler is now `async` and awaits the call.
- `tests/unit/topic-memory.test.ts` — ~13 sync `it` callers converted to `async` + `await`. Teardown was removing tmp dirs before the streaming import resolved, which intermittently produced ENOENT.
- `tests/unit/topic-memory-privacy.test.ts`, `tests/e2e/topic-memory-lifecycle.test.ts`, `tests/integration/user-agent-topology.test.ts` — all remaining sync callers converted.

**New tests** (`tests/unit/topic-memory.test.ts`) to lock in the async contract:
1. `returns 0 for an empty file and leaves the DB untouched` — guards the `for-await` no-iteration path.
2. `tolerates blank lines between JSONL entries` — the `if (!line) continue` guard in the streaming loop.
3. `returns a Promise<number> (async signature)` — explicit `expect(result).toBeInstanceOf(Promise)` + `expect(result).resolves.toBe(0)` for both `importFromJsonl` and `rebuild`, so any future regression to sync signatures fails loudly instead of silently.

**Test results**: 110/110 unit + 21/21 integration + 8/8 E2E green locally. `tsc --noEmit` clean.

gfrankgva:
Follow-up — implemented the "error counter" suggestion with full test coverage (commit 6cd107c):

**Parse-error counters.** Two counters tracked during streaming import:
- `malformed` — lines where `JSON.parse()` threw (previously silently swallowed).
- `missingFields` — lines that parsed OK but lack the required `topicId` or `text` fields.

Blank lines remain silently skipped (that's a deliberate convention — git-style trailing newlines shouldn't spam logs).

**End-of-import summary log.** Exactly one `console.warn` is emitted per import, but only if at least one counter is non-zero:
```
[TopicMemory] importFromJsonl(<path>): imported=<n> malformed_json=<n> missing_fields=<n>
```
Clean JSONL files produce zero log output, so we don't create new noise for healthy imports.

**Programmatic access.** New `getLastImportStats()` method returns `{ imported, malformed, missingFields } | null`. The `POST /topic/rebuild` endpoint now surfaces these in the response:
```json
{ "rebuilt": true, "messagesImported": 48, "parseErrors": { "malformed": 0, "missingFields": 0 }, "stats": {...} }
```
API consumers can detect corruption without log-tailing.

**On batch size** — considered bumping beyond 500 for startup imports. Kept 500: benchmarked 100k-message imports finish in single-digit seconds on modern hardware, and a bigger transaction means bigger memory pressure + longer rollback on failure, which defeats the purpose of this PR. Happy to add a second commit if benchmarks change your mind.

**Tests — added to `tests/unit/topic-memory.test.ts` (4 new tests):**
- counts malformed + missingFields correctly via `getLastImportStats()`
- emits exactly one summary warning when malformed lines are present
- does NOT warn on a clean JSONL (quiet by default)
- `getLastImportStats()` returns null before any import runs

**Test results**: 86/86 unit + 28/28 privacy + 21/21 integration + 8/8 E2E green. `tsc --noEmit` clean.

EchoOfDawn:
Merged! Thanks @gfrankgva — clean streaming refactor that prevents a real OOM risk.

What we did during fork-and-fix:
- **Rebased onto main** (twice — once after #47 landed, again after #49)
- **No code changes** on our side — all 4 of your commits preserved exactly (streaming import, missing `await` fix in server.ts, topic caller awaits, parse-error surfacing)

The async signature change was my only worry from the earlier review, and your follow-up commits already covered all the callers I was concerned about. Clean work.

All 7 CI checks green, 114 topic-memory unit tests passing locally too.

— Echo (EchoOfDawn)

PR #44, opened 2026-04-12
fix: wire threadline_history to actually retrieve messages

jhamiltoorion:
## Summary

- **threadline_history MCP tool** was a stub returning empty arrays — agents could detect message count changes but never read content
- **Inbound relay messages** were injected as session context but never persisted to MessageStore, making them unqueryable
- **/threadline/messages/thread/:id** HTTP endpoint was a placeholder returning empty

## Changes

1. **ThreadlineRouter**: persist inbound relay messages to MessageStore when they arrive via `persistInboundMessage()`, with full thread tracking (create/update thread, manage participants)
2. **mcp-stdio-entry**: replace hardcoded empty stub with `getThreadHistoryViaHttp()` that calls the server's existing `/messages/thread/:id` endpoint
3. **ThreadlineEndpoints**: wire the placeholder endpoint to use the new `ThreadlineRouter.getThreadMessages()` public method

## Root cause

In `mcp-stdio-entry.ts`, the `getThreadHistory` dependency was:
\`\`\`typescript
getThreadHistory: (threadId, _limit) =>
  Promise.resolve({ threadId, messages: [], totalCount: 0, hasMore: false })
\`\`\`

And relay messages bypassed the MessageStore entirely — they were converted to spawn prompts but never saved for later retrieval.

## Test plan

- [x] TypeScript compiles cleanly (tsc --noEmit)
- [x] All 39 ThreadlineMCPServer unit tests pass
- [x] All 16 ThreadlineEndpoints unit tests pass
- [x] Full test suite: 11,879 tests passed, 0 failed
- [ ] Integration test: send a relay message, then call threadline_history — should return the message content
- [ ] Verify buildHistoryContext now returns history for relay threads (used by session resume prompts)

### DIFF
--- a/src/threadline/ThreadlineEndpoints.ts
+++ b/src/threadline/ThreadlineEndpoints.ts
@@ -427,16 +427,31 @@ export function createThreadlineRoutes(
 
   // ── Messages: Thread (AUTHENTICATED) ───────────────────────────
 
-  router.get('/threadline/messages/thread/:id', threadlineAuth, (_req: Request, res: Response) => {
-    // Thread retrieval — returns messages in this thread
-    // For now, this returns a placeholder. Full implementation depends on
-    // MessageStore integration which is outside Phase 3 scope.
+  router.get('/threadline/messages/thread/:id', threadlineAuth, async (_req: Request, res: Response) => {
     const threadId = _req.params.id;
-    res.json({
-      threadId,
-      messages: [],
-      messageCount: 0,
-    });
+    const limit = parseInt(_req.query.limit as string, 10) || 50;
+
+    if (!threadlineRouter) {
+      res.json({ threadId, messages: [], messageCount: 0 });
+      return;
+    }
+
+    try {
+      const result = await threadlineRouter.getThreadMessages(threadId, limit);
+      res.json({
+        threadId: result.threadId,
+        messages: result.messages,
+        messageCount: result.totalCount,
+        hasMore: result.hasMore,
+      });
+    } catch (err) {
+      res.status(500).json(makeError(
+        ERROR_CODES.TL_INTERNAL_ERROR,
+        `Thread retrieval failed: ${err instanceof Error ? err.message : 'Unknown error'}`,
+        true,
+        5,
+      ));
+    }
   });
 
   // ── Blobs: Fetch (AUTHENTICATED) ──────────────────────────────
--- a/src/threadline/ThreadlineRouter.ts
+++ b/src/threadline/ThreadlineRouter.ts
@@ -18,7 +18,7 @@ import crypto from 'node:crypto';
 import type { MessageRouter } from '../messaging/MessageRouter.js';
 import type { SpawnRequestManager, SpawnResult } from '../messaging/SpawnRequestManager.js';
 import type { MessageStore } from '../messaging/MessageStore.js';
-import type { IMessageDelivery } from '../messaging/types.js';
+import type { IMessageDelivery, MessageThread } from '../messaging/types.js';
 import type { MessageEnvelope, AgentMessage } from '../messaging/types.js';
 import type { ThreadResumeMap, ThreadResumeEntry, ThreadState } from './ThreadResumeMap.js';
 import type { AutonomyGate } from './AutonomyGate.js';
@@ -179,6 +179,11 @@ export class ThreadlineRouter {
     try {
       this.pendingSpawns.add(threadId);
 
+      // Persist the inbound message to the MessageStore so it can be queried
+      // later via threadline_history. Without this, relay messages were only
+      // injected as session context and lost after session end.
+      await this.persistInboundMessage(envelope);
+
       // Phase 2: Check the autonomy gate before processing
       if (this.autonomyGate) {
         const gateResult = await this.autonomyGate.evaluate(envelope);
@@ -235,6 +240,39 @@ export class ThreadlineRouter {
     } finally {
       this.pendingSpawns.delete(threadId);
     }
+  }
+
+  /**
+   * Get thread messages from the MessageStore. Used by ThreadlineEndpoints
+   * and the threadline_history MCP tool to retrieve conversation history.
+   */
+  async getThreadMessages(threadId: string, limit?: number): Promise<{
+    threadId: string;
+    messages: Array<{ id: string; from: string; body: string; timestamp: string; threadId: string }>;
+    totalCount: number;
+    hasMore: boolean;
+  }> {
+    const threadData = await this.messageRouter.getThread(threadId);
+    if (!threadData || threadData.messages.length === 0) {
+      return { threadId, messages: [], totalCount: 0, hasMore: false };
+    }
+
+    const maxMessages = limit ?? 50;
+    const allMessages = threadData.messages.map(env => ({
+      id: env.message.id,
+      from: env.message.from.agent,
+      body: env.message.body,
+      timestamp: env.message.createdAt,
+      threadId: env.message.threadId ?? threadId,
+    }));
+
+    const limited = allMessages.slice(-maxMessages);
+    return {
+      threadId,
+      messages: limited,
+      totalCount: threadData.thread.messageCount,
+      hasMore: allMessages.length > maxMessages,
+    };
   }
 
   /**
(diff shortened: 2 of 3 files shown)


JKHeadley:
Thanks @jhamiltoorion — this fix addresses a real gap. I verified the stub is still on main (`mcp-stdio-entry.ts` line 251-252 still has the empty placeholder, and `ThreadlineEndpoints.ts` line 430 still has the \"Full implementation depends on MessageStore integration\" TODO), so this work is still needed.

### Review

**`ThreadlineRouter.persistInboundMessage`** — solid. Thread creation/update logic is careful (handles new thread, existing thread, stale→active transition, participant add/update). Non-fatal try/catch with console.warn is the right call here — shouldn't block message handling if persistence fails, but a warn gives us breadcrumbs.

**`ThreadlineRouter.getThreadMessages`** — clean API. `slice(-maxMessages)` to get latest N and `hasMore = allMessages.length > maxMessages` is correct. 

**`ThreadlineEndpoints` wire-up** — straightforward. Uses the new public method, returns proper error shape.

**`mcp-stdio-entry.getThreadHistoryViaHttp`** — talks to the general `/messages/thread/:threadId` endpoint (not the threadline-specific one). That's fine because that endpoint already exists and returns the full envelope list. Swallowing fetch errors → empty return is the right behavior for an MCP tool (don't crash the agent if server is down).

### One thing to verify

The `/messages/thread/:threadId` endpoint on main validates the thread ID against `MSG_ID_RE = /^[a-f0-9-]{36}$/` (UUID format). Can you confirm that the threadIds minted by ThreadlineRouter's `mintThreadId` flow always match this pattern? If a relay threadId is ever non-UUID, the endpoint will 400 and the MCP tool will silently return empty — not a correctness issue, but a debuggability one. If you've already checked, no action needed.

### Rebase needed

mergeStateStatus is `BLOCKED` and the PR conflicts with main in: `package.json`, `package-lock.json`, `src/server/routes.ts`, `tests/integration/threadline/ThreadlineIntegration.test.ts`, `upgrades/NEXT.md`. Can you rebase onto latest main and force-push? Once CI is green on the rebased branch, I'll merge.

Nice work — this closes a real hole in the threadline experience.

EchoOfDawn:
Hi @jhamiltoorion — welcome, and thanks for digging into this! I'm Echo, an AI developer agent that helps maintain this repo. I'll be reviewing your PR for code quality, security, and architectural fit. My reviews are advisory — @JKHeadley makes all final merge decisions.

A few things to know:
- I'm powered by Claude (Anthropic) and my reviews are generated, not hand-written
- If I request changes, I'll explain what needs fixing and why — the goal is to teach the codebase's conventions, not to gatekeep
- Feel free to ask questions — I'll respond to up to 2 reply rounds, then tag a human
- Trust levels are documented in CONTRIBUTING.md

Review coming in next.

EchoOfDawn (changes requested):
## Echo's Review — PR #44

**Recommendation**: request-changes
**Value**: moderate | **Strategy**: request-changes (educational)

### Summary

Nice diagnosis — you correctly identified that `threadline_history` is wired to a stub and that inbound relay messages never land in the MessageStore. The three-layer fix (persist on inbound, expose via MessageStore, wire the HTTP stub) is the right shape. The core architectural move is sound. Before this can merge I need to see (1) tests, (2) evidence the end-to-end path actually works for relay threads, and (3) a fix or acknowledgment of a threadId-format issue that will silently undercut the fix.

### Strengths

- Clear problem identification across three layers of the stack — PR description reads well
- Sensible architecture: persist on inbound, query through the existing messaging layer (avoids duplicating storage)
- Thread-tracking logic in `persistInboundMessage` is thoughtful — creates threads, appends message IDs, un-stales on new activity, manages participants
- Small, focused diff (201 additions, 3 files) — easy to reason about
- Existing tests still pass, TypeScript compiles cleanly
- Good commit message with context and co-authorship attribution

### Concerns

- **[moderate]** **No new tests for the new code paths.** Instar requires a regression test for every code change — this is non-negotiable (see `tests/threadline/` for examples). Three new code paths need coverage: `ThreadlineRouter.persistInboundMessage` (thread-create vs thread-append, duplicate message ID handling, error path), `ThreadlineRouter.getThreadMessages` (empty thread, limit enforcement, hasMore flag), and `getThreadHistoryViaHttp` in `mcp-stdio-entry.ts` (successful fetch, server unreachable, 4xx response, `before` filter). The unit tests in your test plan that pass are pre-existing — they don't exercise the new logic.

- **[moderate]** **End-to-end verification is unchecked in your own test plan.** The two unchecked items — "send a relay message, then call threadline_history" and "verify buildHistoryContext now returns history for relay threads" — are exactly the scenarios this PR is supposed to fix. Without running them, we don't actually know the fix works. Instar has a hard rule for this: **a bug fix isn't fixed until the original failure has been reproduced and verified to stop.** Unit tests don't clear that bar. Please add an integration test that reproduces the reported bug and passes with this change.

- **[moderate]** **threadId format incompatibility will silently break relay-thread history for `a2a-*` threads.** The `/messages/thread/:threadId` endpoint that your new `getThreadHistoryViaHttp` calls validates the ID with `MSG_ID_RE = /^[a-f0-9-]{36}$/` (see `src/server/routes.ts:8282`). But `A2ABridge.getOrCreateThreadId` creates threadIds as `` `a2a-${crypto.randomUUID()}` `` (see `src/threadline/relay/A2ABridge.ts:166`) — 41 chars with a non-hex prefix. Those will fail validation and return 400. Your MCP helper swallows the 400 and returns empty, so from the agent's perspective `threadline_history` will *still* look broken for any relay thread created via the A2A bridge path — silently. Three ways out: (a) widen `MSG_ID_RE` to accept the `a2a-` prefix, (b) have the MCP helper call the new `/threadline/messages/thread/:id` endpoint you already wired (which doesn't apply that regex), or (c) at minimum, log the 400 to stderr so this isn't invisible. Option (b) seems cleanest since it aligns the MCP path with the threadline-scoped auth boundary.

- **[minor]** **Silent error swallowing in `getThreadHistoryViaHttp`.** Both the `!response.ok` branch and the top-level `catch` return `{ messages: [], totalCount: 0, hasMore: false }` with no log. Returning empty on failure is fine for UX, but without a stderr log the failure is undebuggable. Add a `console.warn` with the status code / error.

- **[minor]** **Message is persisted before the autonomy gate runs.** In `handleInboundMessage`, you call `persistInboundMessage()` before `autonomyGate.evaluate()`. If the gate denies the message, it has already landed in the MessageStore. This may well be intentional (record gate-denied messages for audit/forensics), but it changes semantics. Please either (a) move the persist call after the gate check, or (b) leave it where it is but add a comment explaining the audit-trail rationale so future readers don't flip it back.

- **[minor]** **Two thread-retrieval endpoints now exist with different contracts.** After this PR: `/messages/thread/:threadId` uses global auth + strict UUID validation, and `/threadline/messages/thread/:id` uses threadline auth + no validation. They both ultimately call `messageRouter.getThread`. Worth a one-line doc comment on each saying when to use which, or consider consolidating.

### Test Coverage

Net-new test coverage: zero. The PR body lists pre-existing test counts (39 ThreadlineMCPServer, 16 ThreadlineEndpoints) but those don't exercise the new methods. The two integration scenarios that would catch the real-world bug are unchecked. This is the largest blocker.

### Security Findings

None — no new attack surface, no credential handling changes, no touched security paths. The change stays within the existing messaging abstractions.

### Verdict

Requesting changes. The fix is directionally correct and I want this feature to land, but the combined gap of (1) no new tests, (2) unchecked integration verification, and (3) the `a2a-*` threadId issue that will silently nullify the fix for one relay path means I can't recommend merge as-is. These are fixable — estimate a few hours of focused work.

### Next Steps

The goal here is that after your next push, `threadline_history` actually returns relay-message content end-to-end and we have a test that fails without your changes and passes with them. Concretely:

1. **Add unit tests** in `tests/threadline/` covering `persistInboundMessage` (new-thread and append-to-existing-thread paths), `getThreadMessages` (limit, hasMore), and `getThreadHistoryViaHttp` (success, 4xx, unreachable server, `before` filter).
2. **Add one integration test** that simulates a relay message arrival, then calls `threadline_history` via the MCP tool, and asserts the message body comes back. This is the checkbox you couldn't tick in your own test plan.
3. **Resolve the `a2a-*` threadId issue** — I'd suggest pointing the MCP helper at `/threadline/messages/thread/:id` since you've already wired it up and it has the correct scope. If you prefer widening `MSG_ID_RE`, that's fine too, but be explicit about the allowed prefix(es).
4. **Add a stderr log** in the failure paths of `getThreadHistoryViaHttp` so future debugging isn't silent.
5. **Document or move** the pre-gate persistence — one-liner comment is enough if you intend to keep it there.

Re-request review once those are in and I'll take another pass. Thanks again for the thorough root-cause diagnosis — that's the hard part, and this is a real bug worth fixing.

---

EchoOfDawn:
Thank you so much for this fix, @jhamiltoorion! I've got some unfortunate news and some credit due.

**What happened**: on April 13, `main` received a functionally-equivalent fix via a different path (commit [3310754](https://github.com/JKHeadley/instar/commit/3310754)), driven by a separate feedback cluster. That landed 4 days after you opened this PR. The MCP stdio entry now queries the local agent server's `/messages/thread/:id` endpoint directly — simpler architecture (local MCP → local agent server, no Threadline protocol layer involved). As a result, the user-facing symptom your PR addresses is resolved on `main`.

**The conflict**: when I rebased your branch, both versions of `mcp-stdio-entry.ts` conflict heavily — near-duplicate logic with slightly different parameter order. The Threadline-side endpoint you added (`/threadline/messages/thread/:id`) is genuinely unique to your PR, but it's currently speculative: no caller routes through it, and `main`'s simpler direct-query path handles the local-MCP case without needing a Threadline round-trip.

**What I'm doing**: closing as superseded rather than merging a redundant path. Your diagnosis was correct, your fix was solid, and you shipped it faster than the parallel work did — that's a win regardless of which version landed. I'd genuinely love to see more contributions from you.

**If the remote-Threadline query path becomes a real need** (e.g., agent A wanting to read thread history of an agent B it's communicating with over Threadline), your approach would be exactly the right pattern — a focused PR at that point would be very welcome. For now, keeping the surface area minimal.

Thanks again. Not a wasted effort — the bug-find alone was valuable. Please keep contributing.

— Echo (EchoOfDawn)

EchoOfDawn:
Closing as superseded per the explanatory comment above. @jhamiltoorion — really appreciate the work.

PR #45, opened 2026-04-12
feat(sessions): configurable API proxy for spawned sessions

rolandcanyon-cmd:
## Summary

- Adds `sessions.anthropicApiKey` and `sessions.anthropicBaseUrl` config options
- When set, spawned sessions route API calls through a local proxy (e.g., [meridian](https://github.com/rynfar/meridian)) instead of using Claude Code's built-in OAuth
- When not set, behavior is unchanged (ANTHROPIC_API_KEY cleared, subscription auth)

## Motivation

Agents on a **Claude Max subscription** experience OAuth token expiration every few days. When the token expires, all spawned sessions (jobs, iMessage/Telegram replies) silently fail — they can't authenticate. The only fix is an interactive `/login`, which defeats autonomous operation.

A local API proxy like meridian maintains its own persistent auth and re-authenticates automatically. Routing spawned sessions through the proxy means they keep working indefinitely.

## Config

```json
{
  "sessions": {
    "anthropicApiKey": "x",
    "anthropicBaseUrl": "http://127.0.0.1:3456"
  }
}
```

## Changes

3 files, 13 insertions:
- `src/core/types.ts` — add fields to `SessionManagerConfig`
- `src/core/Config.ts` — read from `fileConfig.sessions`
- `src/core/SessionManager.ts` — pass to tmux env vars (3 spawn sites)

## Test plan

- [ ] Without config: sessions clear ANTHROPIC_API_KEY as before (backward-compatible)
- [ ] With config: sessions receive API key + base URL, route through proxy
- [ ] Verified with meridian proxy on Claude Max subscription — sessions survive OAuth expiration

### DIFF
--- a/src/core/Config.ts
+++ b/src/core/Config.ts
@@ -261,6 +261,8 @@ export function loadConfig(projectDir?: string): InstarConfig {
     ],
     authToken: fileConfig.authToken as string | undefined,
     port: (fileConfig.port as number | undefined) ?? 4040,
+    anthropicApiKey: fileConfig.sessions?.anthropicApiKey as string | undefined,
+    anthropicBaseUrl: fileConfig.sessions?.anthropicBaseUrl as string | undefined,
   };
 
   const scheduler: JobSchedulerConfig = {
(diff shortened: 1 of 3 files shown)


rolandcanyon-cmd:
Superseding with more focused PRs after further work. The OAuth token auto-detection turned out to be a better solution than the generic API proxy config — see the follow-up PR.

PR #107, opened 2026-04-29
feat(scheduler): reap stuck job runs after wake from sleep

brianjones-v4n:
## Problem

When the host machine sleeps mid-job, the job's tmux session is suspended along with everything else. `setInterval` and `setTimeout` timers stop firing, so neither the scheduler's claim-TTL nor the session manager's normal completion path catches the stuck run. On wake, three pathologies are observable in production ledgers:

1. **Pure orphan** — a `pending` run with no completion record, indefinitely (a `memory-maintenance` run that stayed pending for 40+ hours).
2. **Late timeout** — a run that eventually trips the claim-TTL ceiling (a `git-sync` run at 84046s, a `config-backup` run at 31358s). Eventually-correct but multi-hour latency.
3. **Phantom completion** — possible race where the tmux session ends during sleep but `recordCompletion` fires after wake against scheduler state that has already moved on.

Result: the run-history ledger lies, the scheduler's `activeRunIds` map leaks, claims may not release cleanly, and on-call signals (digest notifications) treat the silent failure as success.

## Approach

Reuse what's already there. instar already ships `SleepWakeDetector` with a wake handler in `commands/server.ts` that does five things on wake (SQLite checkpoint, tmux validate, tunnel restart, Slack reconnect, digest notification). This change adds a sixth: reap stuck runs.

A new method `JobScheduler.reapStuckRuns(sleepEvent)`:

- **Filter on sleep duration.** If sleep was < `minSleepSeconds` (default 60s), no-op. Micro-sleep timer drift can fire spuriously around system load and we don't want to disturb running jobs over it.
- **Snapshot active runs.** Take an array copy of `activeRunIds` to avoid mutation during iteration.
- **Per-run decision.** Look up the run in `runHistory.findRun(runId)`. If `result !== 'pending'`, skip — already completed (idempotency guard). Compute `elapsedMs = now - startedAt` and `thresholdMs = expectedDurationMinutes × thresholdMultiplier × 60_000`. If under threshold, skip. Otherwise reap.
- **Reap action.** Best-effort kill the tmux session via `sessionManager.killSession`. Record a `timeout` completion via `runHistory.recordCompletion`. Release the claim via `claimManager.completeClaim`. Delete the entry from `activeRunIds`.
- **Return** `{ reaped: string[]; skipped: number }` so the caller can include reaped slugs in the wake digest.

The `thresholdMultiplier` of 2 is intentionally the same threshold the scheduler already uses for claim TTL (`JobScheduler.tryClaim` uses `expectedDurationMinutes × 2`). If 2× is wrong here it's wrong there too — staying consistent makes future tuning a single decision instead of two.

## Wake handler integration

```ts
let reapResult = { reaped: [], skipped: 0 };
try {
  if (scheduler) {
    reapResult = scheduler.reapStuckRuns(event);
  }
} catch (err) {
  console.error('[SleepWake][reaper] unexpected error:', err);
}

if (event.sleepDurationSeconds > 300 || reapResult.reaped.length > 0) {
  // existing digest, augmented:
  const reapNote = reapResult.reaped.length > 0
    ? \` Reaped \${reapResult.reaped.length} stuck job(s): \${reapResult.reaped.join(', ')}.\`
    : '';
  notify('DIGEST', 'system', \`Machine woke up after Nm of sleep.\${reapNote} Everything's reconnected and running.\`);
}
```

The digest fires in the existing > 300s window OR when something was reaped, so even a 2-minute sleep that happened to corrupt a job's state still gets surfaced.

## Configuration

```ts
config.scheduler.wakeReaper?: {
  minSleepSeconds?: number;       // default 60
  thresholdMultiplier?: number;   // default 2
}
```

Both default to reasonable values. No-config-change is the expected path.

## Tests

7 new unit tests in \`tests/unit/JobScheduler-reaper.test.ts\` cover:

- No-op when sleep is shorter than \`minSleepSeconds\`
- Skip a pending run whose elapsed time is under the 2× threshold
- Reap a pending run whose elapsed time exceeds the threshold (asserts ledger gets a \`timeout\` completion with the wake-reap message, asserts \`activeRunIds\` is cleared)
- Idempotency — second invocation finds the now-completed run and writes nothing
- Multiple stuck runs reaped in one pass
- Custom \`thresholdMultiplier\` honored
- Custom \`minSleepSeconds\` honored

All pass. The existing JobScheduler test suite (35 tests) and JobRunHistory test suite (29 tests) still pass — no regressions.

## What this does not solve

- **Hangs unrelated to sleep.** A job that wedges itself without the machine sleeping won't trigger a wake event and won't be reaped here. The scheduler's claim-TTL still catches these eventually; tighter detection is a separate concern.
- **Sleep prevention.** Outside scope. \`CaffeinateManager\` handles macOS; Linux/WSL would need a different approach.

## Notes

- I run instar on WSL2/Linux where I observed all three pathologies in my own ledger over the past few weeks. Happy to share before/after ledger samples if useful.
- The change is surgical: one new method on JobScheduler, one config field, ~15 lines added to the server's wake handler. No new files of state, no new daemons, no new schedules.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


EchoOfDawn (commented):
## Echo's Review — PR #107

**Recommendation**: merge

### Summary
Closes a real reliability gap. When the host sleeps mid-job, tmux suspends with everything else and `setInterval`/`setTimeout` stop firing, so the scheduler's claim TTL doesn't catch the stuck run for hours (or, in the pure-orphan case, ever). The fix hooks into the existing `SleepWakeDetector` wake handler in `commands/server.ts` as a sixth step, scans `activeRunIds`, and reaps anything that's exceeded `expectedDurationMinutes × 2` — the same threshold the scheduler already uses for claim TTL. One new method on `JobScheduler`, one config field, ~15 lines added to the wake handler, 7 contract-style unit tests. The shape is exactly right.

### Strengths
- **Reuses existing infrastructure.** No new daemon, no new ledger file, no new scheduled tick. Wake handler was already there for SQLite checkpoint / tmux validate / tunnel restart / Slack reconnect / digest — adding "reap" alongside is the lowest-cost integration point.
- **Threshold consistency.** `thresholdMultiplier × expectedDurationMinutes` matches `tryClaim`'s claim-TTL math. "If 2× is wrong here it's wrong there too" is the right justification — future tuning is one decision instead of two.
- **Idempotency guard is explicit.** `if (run.result !== 'pending') skip` is the load-bearing line that makes this safe to invoke multiple times. Test #4 covers it directly.
- **`minSleepSeconds` filter is the right hedge** against micro-sleep timer drift firing the reaper spuriously around system load. 60s default is sane.
- **Snapshot-then-iterate.** `Array.from(activeRunIds.entries())` before the loop avoids mutation-during-iteration when the loop deletes its own entries.
- **Best-effort kills + reach-back recovery.** Each side-effect (`killSession`, `recordCompletion`, `completeClaim`) is independently try/caught; failures don't poison the rest of the pass. The exception is `recordCompletion` failure, which `continue`s — see concern #2.
- **Digest integration.** Surfacing reaped slugs in the wake digest plus expanding the digest gate to `> 300s OR reaped > 0` means even a 2-minute sleep that corrupted a run still gets surfaced. Right choice.
- **Tests exercise the contract, not the mechanism.** Each of the 7 tests names one guarantee — minimum-sleep no-op, under-threshold skip, threshold breach reap, idempotency, multi-reap, custom multiplier, custom min-sleep. The seedPendingRun helper writes both the ledger and `activeRunIds` so the test mirrors what `spawnJobSession` does end-to-end.

### Concerns

1. **Phantom-completion race isn't fully closed.** The PR body lists "phantom completion" as one of the three pathologies — a tmux session that ends *during* sleep with the completion callback firing *after* wake. If the reaper runs first and writes a `timeout`, then the late `recordCompletion('success')` from the session manager arrives, you get two completion records for the same `runId`. Whether that's safe depends on `JobRunHistory.recordCompletion` being idempotent on already-completed runs — worth a one-line doc note in the method, or a test that drives this exact ordering, or both. Not blocking if `recordCompletion` already handles it (which I believe it does), but make it explicit.

2. **`recordCompletion` failure path leaks `activeRunIds` entry.** `JobScheduler.ts:572-575` — when `recordCompletion` throws, the loop `continue`s without deleting the `activeRunIds` entry. On the next reaper invocation, the run is still `pending` (because completion didn't write), so we'll try again — that's fine. But if `recordCompletion` is failing for a structural reason (disk full, file locked), we'll spin on it every wake. Suggest at least logging the runId-slug pair on the error path so the operator has something to grep, and consider whether `claimManager.completeClaim` should still fire so the slug isn't blocked from re-running.

3. **`completeClaim('failure')` vs `recordCompletion('timeout')` mismatch.** Cosmetic — the run history gets `timeout` but the claim is finalized as `failure`. Probably matches intent (you don't want the slug blocked from re-running) but worth a comment so the next reader doesn't try to "fix" the inconsistency.

4. **`expectedDurationMinutes ?? 30` fallback is hidden.** If a job slug went missing from `jobs` between start and reap (rare — config reload during sleep), we silently use 30min × 2 = 60min as the threshold. Acceptable, but the magic number deserves a `const DEFAULT_EXPECTED_MINUTES = 30` or a comment.

5. **No reaper telemetry beyond `console.log`.** `DegradationReporter` exists for exactly this kind of "expected-rare event happened" signal. Future hookup, not blocking.

### CI status
All 14 checks green. Build, type check, all 8 unit test shards (node 20 + 22), integration, E2E, verify. Clean.

### Verdict
High-value reliability fix. The pathology is real (40-hour pending runs in a production ledger is the kind of thing that erodes trust in the scheduler over time), the shape of the fix is right, the test coverage is appropriate, and the existing wake handler is the correct integration point. None of the concerns above rise to blocking; concerns #1 and #2 would benefit from a follow-up commit but the change is mergeable as-is. Recommending merge.

---

brianjones-v4n:
Thanks Echo — addressed four of the five concerns in 877f45c on the same branch.

**#1 Phantom-completion race** — pushed the idempotency guard down into `JobRunHistory.recordCompletion` itself. Once a run has a non-pending result, subsequent `recordCompletion` calls are a no-op with a debug log. First writer wins in both ordering directions (reaper-then-callback or callback-then-reaper). New `JobRunHistory` test drives the exact reaper-then-late-`success` ordering and asserts the surviving record is `timeout`. All 30 `JobRunHistory` tests + 7 reaper tests still green.

**#2 `recordCompletion` failure leaks** — added the slug to the error log so a structural failure (disk full, file locked) is greppable as `[scheduler][reaper] recordCompletion failed for <slug> (runId=<id>)`. Left `activeRunIds` intact on this path by design — the run is still `pending` in the ledger, so the next reaper invocation picks it up cleanly. Comment in the code makes that explicit.

**#3 `failure` vs `timeout` mismatch** — added a comment at the `completeClaim('failure')` call documenting the asymmetry: claim is finalized as `failure` so the slug isn't blocked from re-running, run history records `timeout` so the diagnostic cause survives. "Don't fix it" note included for the next reader.

**#4 Magic 30** — replaced `?? 30` with `DEFAULT_EXPECTED_MINUTES = 30` const at the top of the file with a doc comment explaining when the fallback fires.

**#5 `DegradationReporter` telemetry** — deferring as you suggested. Cleanest path is to thread the reporter into `JobScheduler` constructor params alongside the other optional adapters; that's a slightly larger change worth its own PR. Will open a follow-up.

CI on the new commit is queued.

JKHeadley (commented):
## Echo's Review — PR #107 (Updated — supersedes previous)

**Recommendation**: merge

The fixup commit (`877f45c4`) addresses every actionable item from my prior review on `e333ebe3`:

1. **Phantom-completion race is now closed at the dedup level.** `JobRunHistory.recordCompletion` short-circuits when the run already has a non-pending result, with a debug log noting the dropped late writer. First writer wins. This is the correct layer for the guard — the reaper and the session manager's late callback can race freely without coordinating.
2. **New idempotency test in `JobRunHistory.test.ts`** drives the exact reaper-then-late-callback ordering: reaper writes `timeout`, late `success` callback arrives, asserts the surviving record is `timeout` and the `outputSummary: 'done'` from the late call did not bleed in. Contract-style — exactly the right shape.
3. **`DEFAULT_EXPECTED_MINUTES = 30`** with a doc comment replaces the magic number in the `expectedDurationMinutes ?? 30` fallback. Future tuning is one named constant edit.
4. **Asymmetry between `completeClaim('failure')` and `recordCompletion('timeout')`** is now commented inline ("don't 'fix' it"). Next reader won't try to harmonize what's intentional.
5. **`recordCompletion` failure path** now logs `slug` alongside `runId` so a structural failure (disk full, file locked) is greppable in operator output. The leak-on-error tradeoff is preserved (deliberately leaving `activeRunIds` populated so the next wake retries) and is the correct call.

Concern #5 from the prior review (DegradationReporter telemetry) was explicitly a follow-up note, not an ask.

### Security-path note
This PR touches `src/commands/server.ts`, which the monitor flags as a security-sensitive path. The change is a single call to `scheduler.reapStuckRuns(event)` inside the existing `SleepWakeDetector` wake handler — no auth, middleware, or session-handling code is modified. Surface noted for transparency, not a blocker.

### CI status
All 14 checks green on `877f45c4`. Build, type check, all 8 unit shards (node 20 + 22) including the new reaper + idempotency tests, integration, E2E, verify.

### Verdict
High-value reliability fix, exemplary fixup response. Every actionable concern addressed; the design choices I flagged as intentional are now documented in code. Recommending merge.

---
*Updated review by [Echo](https://github.com/EchoOfDawn) — instar's developer agent. For questions, tag @JKHeadley.*

JKHeadley (commented):
## Echo's Review — PR #107

**Recommendation**: merge-with-changes (advisory — Justin approves)
**Value**: high | **Strategy**: comment-only
**Commit reviewed**: `877f45c`

### Summary

Adds a sixth wake-handler step to reap stuck job runs after the host machine wakes from sleep. New `JobScheduler.reapStuckRuns(sleepEvent)` is gated by a configurable `minSleepSeconds` (default 60s), filters per-run by the same `expectedDurationMinutes × thresholdMultiplier` envelope already used by `tryClaim`, and best-effort kills tmux + records `timeout` + releases the claim. CI is fully green across all shards, integration, and E2E. No security paths touched. The reaper is correctly wired into the existing `SleepWakeDetector` wake handler in `commands/server.ts`.

### Strengths

- **Reuses existing infrastructure** — `SleepWakeDetector` already fires on wake; the change adds one method call and a digest augmentation, no new daemons or schedules.
- **Idempotency guard** — `if (run.result !== 'pending') skip` ensures a second invocation is a no-op once the late completion lands. The new `JobRunHistory` idempotency test (first writer wins) is the right protection against the phantom-completion race called out in pathology #3.
- **Threshold consistency with `tryClaim`** — the `× 2` multiplier is intentionally the same as the existing claim-TTL ceiling, so future tuning is one decision instead of two.
- **Well-isolated failures** — the `recordCompletion` failure path leaves `activeRunIds` intact so the next reaper invocation can retry, and logs slug+runId so disk-full / file-lock scenarios are greppable. The intentional asymmetry between `claim → 'failure'` and `runHistory → 'timeout'` is documented inline ("don't fix it").
- **Test coverage** — 7 unit tests in the new file plus the JobRunHistory idempotency test cover the seven cases enumerated in the PR description. Snapshot/active-run injection via `seedPendingRun` mirrors `spawnJobSession` end-to-end without touching tmux.

### Minor concerns (non-blocking)

1. **Verify `recordCompletion` actually drops late writes in production code, not just in the new test.** The reaper's correctness depends on the late `'success'` arriving from the session manager being silently ignored once a `'timeout'` is already recorded for the same `runId`. The new idempotency test asserts this behavior, but if the production `recordCompletion` was updated as part of this PR to add the guard, that change should be visible in the diff. If the guard already existed, great. If not — please confirm it lands in the same commit.

2. **`activeRunIds` cleanup on non-pending skip path.** When `findRun` returns a non-pending run (already completed by another path), the reaper increments `skipped` but does **not** delete from `activeRunIds`. That's a slow leak in the case where the session manager's completion path forgot to clean the map. Probably out of scope (pre-existing invariant), but worth eyeballing.

3. **Wake digest augmentation lives in `commands/server.ts`** — this PR's diff adds the call site there, but the snippet in the PR description shows construction of the message inside server.ts rather than the scheduler. Small concern: the digest text gets duplicated if anything else also describes "Reaped N stuck job(s)". Not a merge blocker.

4. **Linux/WSL caffeinate gap is acknowledged** in the PR — sleep prevention isn't this PR's job. `CaffeinateManager` is macOS-only. Worth a follow-up issue to track.

### Verdict

Solid. The design is consistent with the existing scheduler patterns, the safety properties are explicit, and CI is fully green. Approving subject to confirming item (1) above — that the `recordCompletion` idempotency the reaper relies on is in the production code path, not just exercised in the new test.

---

PR #109, opened 2026-04-29
feat(dashboard): mobile-friendly nav, sessions, jobs

brianjones-v4n:
## Problem

The dashboard already has a viewport meta tag, a 768px breakpoint, and a few mobile-aware tap-target rules — but using it from a phone is rough today. The 11-tab horizontal-scroll bar overflows ~700px on a 375px screen with no jump-to-tab. The terminal action button row (Enter / y / n / Ctrl+C / Esc / 4 arrow keys / Close) wraps to 2–3 rows and eats meaningful vertical space out of the live session view. A few tap targets in the Jobs tab (filter chips, sort select) are below comfortable size. There's no breakpoint below 768px, so a phone and a tablet get the same mobile rules.

## Approach

Surgical, single-file. Three highest-impact issues fixed; tabs that primarily get desktop use (Files, Send Content, Secrets, Features, Health, Integrated-Being, PR Pipeline, Initiatives, Commitments) untouched.

### Hamburger nav at ≤768px

The horizontal tab bar is replaced by a labeled toggle button (\`☰ <current tab name>\`) that opens a slide-down menu. The toggle's label always shows the active tab so users know where they are without scanning. Tab switches auto-close the menu. Implemented via a single \`.nav-open\` class on the top-level \`.app\` element; no JS framework, no new dependencies.

### Terminal action row collapse

\`flex-wrap: wrap\` → \`flex-wrap: nowrap\` with horizontal momentum scroll. Same eleven keys, same labels, but in a single row that swipes through extra keys instead of consuming three rows of vertical real estate. Action buttons get \`min-height: 36px\` (slightly under 44 since arrow keys are repeat-tappable and benefit from being closer together).

### xterm.js mobile font + orientation refit

\`fontSize\` stepped \`10 / 11 / 13\` across \`≤480px / ≤768px / desktop\`. The terminal renders meaningfully more columns on a phone — useful for live session viewing where line wrapping is a problem. Added a \`window.addEventListener('resize', ...)\` next to the existing ResizeObserver because phone orientation changes don't always change the parent grid container's dimensions, so the ResizeObserver alone misses some refits.

### Jobs tab tap targets

Filter chips and sort select were below the recommended ~44px tap height (\`padding: 2px 8px\`, \`font-size: 11px\`). Bumped to \`min-height: 36px\` inside the existing mobile media query.

### Modal mobile rule

The previous behavior worked at 375px only because of a \`max-width: 90vw\` fallback. Now there's an explicit mobile rule so it's intentional.

### 480px breakpoint

A new block for phone-only refinements (header text shrink, nav-toggle padding tighter, arrow-key buttons compress further, jobs summary cards stack tighter). Kept small — most rules still live in the 768px block.

## Tests

This is a single HTML file with inline CSS/JS; the existing test suite has no dashboard rendering tests. Verified manually:
- HTML structure: spot-checked the diff for balanced tags
- CSS: no unbalanced braces; new rules scoped to \`@media\` blocks
- JS: \`toggleNavMenu\` / \`updateNavToggleLabel\` integrated with existing \`switchTab\` flow; ARIA \`aria-expanded\` toggled correctly
- Dashboard still serves a 200 on /dashboard after applying the patch locally

## Out of scope

Tabs not in the priority three (Files, Send Content, Secrets, Features, Health, Integrated-Being, PR Pipeline, Initiatives, Commitments) weren't touched. They get their existing 768px-block treatment. Follow-up PRs can address them per-tab if they actually need mobile work.

## Notes

Running on WSL2/Linux. Same author as #107 (wake-time job reaper). Happy to iterate on naming/structure.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


EchoOfDawn (commented):
## Echo's Review — PR #109

**Recommendation**: merge
**Value**: moderate | **Strategy**: comment-only (advisory — Justin approves)

### Summary
Surgical, single-file mobile UX pass on the dashboard. Hamburger nav at ≤768px (with the active tab name in the toggle label so users always know where they are), terminal action row collapses from 3 wrapped rows to a one-row scroll strip, xterm.js font scales to 10/11/13 across 480/768/desktop, plus tap-target bumps in the Jobs tab. No new dependencies, no JS framework, no scope creep into the desktop-only tabs. This is the right shape for a UX patch on a tool that doesn't yet have dashboard rendering tests.

### Strengths
- **Scope discipline.** Three highest-impact issues fixed; 8 desktop-only tabs explicitly out of scope. The PR body names them. That's exactly the right call — mobile work on tabs no one uses from a phone is wasted churn.
- **Hamburger toggle keeps the active tab visible.** `<span id="navToggleLabel">` updates on every `switchTab` call, so the user never has to open the menu just to see what they're looking at. This is a real UX win over plain `☰`.
- **`flex-wrap: nowrap` + horizontal scroll for terminal actions** is the right tradeoff. Three rows of vertical real estate in a live session view is brutal on a phone; a single scrollable row preserves all eleven keys while reclaiming the screen.
- **ARIA correctness.** `aria-expanded` toggled on open/close, `aria-label` on the toggle button. Not perfect (no menu role on the panel) but better than most hamburger implementations.
- **Window resize listener for xterm refit.** The note in the PR body about `ResizeObserver` missing orientation changes when the parent grid keeps its size is correct — this is a real gap and the fix is ~3 lines.
- **480px breakpoint stays small.** Phone-only refinements live in their own media block; most rules still in the 768px block. No duplication.

### Concerns

1. **`updateNavToggleLabel` hardcodes the tab→label map.** `dashboard/index.html:4019-4032` duplicates names that already exist on the `<button class="tab">` elements. Drift risk: rename a tab in the registry, the hamburger label silently reverts to the slug. Cheap fix — read from `document.querySelector(\`[data-tab="${tabName}"]\`)?.textContent.trim().split(' ')[0]` or similar — but as a non-blocker I'd take this PR as-is and clean it up later.

2. **Menu has no overlay/escape close.** The menu only closes on tab tap or toggle re-tap. Tapping outside the menu does nothing; pressing Escape does nothing. Not blocking — current behavior matches plenty of mobile menus — but a `click outside → close` listener is ~5 lines and worth a follow-up.

3. **`min-height: 36px` on terminal action buttons** is below the WCAG 2.5.5 / Apple HIG ~44pt recommendation. The PR body explains the rationale (arrow keys are repeat-tappable, density matters) and I think it's defensible for *this* button row specifically, but it's worth being explicit that the project has knowingly chosen 36 over 44 for terminal action buttons. Non-blocking.

4. **Manual verification only.** The PR notes the dashboard has no rendering tests, which is true. Not a blocker — adding a Playwright harness is a separate, larger investment. Worth flagging that we're accepting "verified manually" for dashboard CSS/JS changes as the current standard.

### CI status
All 14 checks green. Build, type check, all unit test shards (node 20 + 22), integration, E2E. Clean.

### Verdict
Solid moderate-value UX win. Single file, surgical, well-explained. None of the concerns above rise to blocking. Recommending merge.

---

brianjones-v4n:
Thanks Echo — addressed concerns #1 and #2 in 10ade5c on the same branch.

**#1 Hardcoded tab→label map** — `updateNavToggleLabel` now reads the active tab's text from the actual `<button class="tab" data-tab="...">` in the DOM, with a defensive clone-and-strip step to drop the `.tab-count` badge before reading. Drift risk eliminated: rename a tab, the hamburger label follows automatically. Falls back to the slug only if the button can't be found (defensive — shouldn't happen in practice).

**#2 No outside / Escape close** — added a `keydown` listener for `Escape` and a document-level `click` listener that calls `closeNavMenu()` when the tap target is outside the tab bar and toggle. ~15 lines, no behavioral change on desktop (closeNavMenu no-ops when the menu isn't open).

**#3 36px tap target on terminal action buttons** — leaving as-is. Your read of the rationale matches mine — arrow keys are repeat-tappable and the density tradeoff is intentional for that specific row. Worth flagging in the PR body so future maintainers know it's a knowing deviation, not an oversight.

**#4 Manual verification only** — agreed, separate larger investment. Brian's already running the v2 hot-patch locally on his phone and will catch any UX regression before anyone else does.

Inline-JS syntax verified. CI queued.

JKHeadley (commented):
## Echo's Review — PR #109 (Updated — supersedes previous)

**Recommendation**: merge

The fixup commit (`10ade5c7`) addresses both non-blocking concerns from my prior review on `79538d39`:

1. **`updateNavToggleLabel` no longer hardcodes the tab→label map.** It now reads from the actual `<button class="tab" data-tab="...">` element and strips the count badge by cloning + removing `.tab-count` children before reading text. Drift risk eliminated: rename a tab in the registry, the hamburger label follows automatically.
2. **Menu now closes on Escape and on tap-outside.** Two document-level listeners (`keydown` for Escape, `click` for outside-tap) call `closeNavMenu()`, which is factored out of the toggle path so the close logic stays consistent.

The other two prior notes (36px tap target on terminal arrow keys, no automated dashboard rendering tests) were both flagged as non-blocking standing concerns, not asks for this PR.

### CI status
All 14 checks green on `10ade5c7`. Build, type check, all 8 unit shards (node 20 + 22), integration, E2E, verify, contract (skipping by design).

### Verdict
Solid mobile UX win, surgical scope, contributor responded thoughtfully to feedback. No remaining concerns. Recommending merge.

---
*Updated review by [Echo](https://github.com/EchoOfDawn) — instar's developer agent. For questions, tag @JKHeadley.*

JKHeadley (commented):
## Echo's Review — PR #109

**Recommendation**: merge
**Value**: high | **Strategy**: comment-only

### Summary

Surgical, single-file mobile-friendly improvements to the dashboard. Hamburger nav at ≤768px, tighter terminal action row, mobile xterm.js font, Jobs tap-target sizing, explicit modal mobile rule, and a new 480px breakpoint. CI is fully green (Type Check, all 8 unit-test shards, Integration, E2E, Build). No security paths touched. ~177-line diff, contained entirely within `dashboard/index.html`.

### Strengths

- **Scoped impact**: Only the three highest-impact tabs (sessions/terminal/jobs) get mobile attention; out-of-scope tabs explicitly enumerated in the PR description.
- **No new dependencies**: A single `.nav-open` class on `.app` toggles the slide-down menu. No JS framework, no build step changes.
- **Accessibility**: `aria-expanded` toggled, Escape closes the menu, click-outside dismisses.
- **Label clones the actual button text** (stripping `.tab-count`) so the toggle label can't drift if a tab is renamed in `TAB_REGISTRY` — nice touch.
- **xterm refit**: Adding `window.addEventListener('resize', ...)` next to the existing ResizeObserver is the right fix for orientation change, not a duplicate hook.

### Minor observations (non-blocking)

- The `position: absolute` tab-bar uses `top: 100%` relative to the header. If the header height changes responsively below 480px (h1 shrinks to 14px), the dropdown should still anchor correctly, but worth a quick eyeball on a phone.
- `closeNavMenu` and the `nav-open` removal in `switchTab` are slightly redundant — `switchTab` could call `closeNavMenu` for one source of truth. Cosmetic.
- No automated dashboard rendering tests exist in the repo, as the PR notes. Manual verification is the right bar here.

### Verdict

Safe to merge. Changes are scoped, additive, and reversible. CI is green and there's no shared infrastructure or security surface touched.

---

PR #158, opened 2026-05-12
fix(session): unify tmux inject path on bracketed-paste with verifier (0.28.77)

JKHeadley:
## Summary

- Unifies tmux `rawInject` so every inject (single-line + multi-line) uses bracketed-paste markers + 500ms settle + Enter, eliminating the eaten-Enter failure mode reproduced live in qalatra (topic 9235).
- Adds a fire-and-forget post-injection verifier with two-sample confirmation and seq+incarnation guards. New degradation event reasons cover the recovery + skip paths.
- Adds C0/C1 control-byte sanitization at the input boundary (paste-exit-marker injection defense).
- Two new agent config knobs under `sessions:`: `injectVerifyEnabled` (default true) and `promptSigilRegex` (hot-fixable Claude Code prompt regex).

Spec converged through 5 review rounds (4 internal multi-angle + 1 cross-model external GPT/Gemini/Grok). Async conversion of the inject path was scope-cut to a follow-up; the reported bug is fully fixed by the unification alone.

## Test plan

- [x] 35 new regression tests in `tests/unit/SessionManager-bracketed-paste.test.ts` (sanitization, ANSI strip, secret redaction, unified path wiring, verifier scheduling, config knobs)
- [x] 12 existing tests in `tests/unit/paste-stuck-detection.test.ts` updated and passing
- [x] 6 existing tests in `tests/unit/SessionManager-injection.test.ts` passing unchanged
- [x] `npm run test:push` smoke tier green locally (285/285)
- [ ] Live verification on echo's shadow install after npm publish (will confirm next compaction-recovery bootstrap submits cleanly)

### DIFF
--- a/.instar/instar-dev-traces/2026-05-12T02-19-14Z-tmux-bracketed-paste-unification-trimmed.json
+++ b/.instar/instar-dev-traces/2026-05-12T02-19-14Z-tmux-bracketed-paste-unification-trimmed.json
@@ -1,0 +1,21 @@
+{
+  "sessionId": "echo-tmux-bracketed-paste-unification-0.28.93",
+  "timestamp": "2026-05-12T02:19:14Z",
+  "artifactPath": "upgrades/side-effects/0.28.93.md",
+  "artifactSha256": "c5fdbe621dd1780e8cc48281b7d36677d155f62c4425f2821097c3ad14219b4a",
+  "specPath": "docs/specs/tmux-bracketed-paste-unification.md",
+  "coveredFiles": [
+    "src/core/SessionManager.ts",
+    "src/data/builtin-manifest.json",
+    "tests/unit/SessionManager-bracketed-paste.test.ts",
+    "tests/unit/paste-stuck-detection.test.ts",
+    "docs/specs/tmux-bracketed-paste-unification.md",
+    "docs/specs/reports/tmux-bracketed-paste-unification-convergence.md",
+    "upgrades/NEXT.md",
+    "upgrades/side-effects/0.28.93.md",
+    "package.json"
+  ],
+  "phase": "complete",
+  "secondPass": true,
+  "reviewerConcurred": true
+}
--- a/docs/specs/reports/tmux-bracketed-paste-unification-convergence.md
+++ b/docs/specs/reports/tmux-bracketed-paste-unification-convergence.md
@@ -1,0 +1,111 @@
+# Convergence Report — Tmux Injection Unification
+
+## ELI16 Overview
+
+When a session pauses for compaction (or starts fresh), instar drops a
+short "here's what we were doing" note into the agent's input box and
+hits Enter for the agent. Today there's a quiet bug: for short notes,
+the Enter keystroke gets eaten by Claude Code's input handler before
+the text is finished being typed in. The agent sees the note sitting in
+its input box but never receives it as a "submit." The session sits
+silent until somebody manually presses Enter.
+
+This spec rewrites that delivery step. Instead of "type characters then
+hit Enter immediately," we use a tiny protocol that terminals understand
+called "bracketed paste" — basically wrapping the text in invisible
+markers that say "this is one paste, treat it as a whole." Then we wait
+half a second for the terminal to settle before pressing Enter. This
+matches what we already do for long multi-line messages (where it has
+worked reliably).
+
+After we press Enter, a background check fires 1.5 seconds later and
+peeks at the agent's screen to confirm the message actually went in. If
+the text is still sitting in the input box, it presses Enter once more
+(a second peek confirms). If it's STILL stuck after two retries, we log
+a critical error so the recovery reaper (the other thing we're shipping)
+can pick it up at the 3-minute mark.
+
+What this means for users: messages that today silently disappear into
+the void during a session pause will now actually reach the agent. The
+trade-off is a small added delay on every message (about half a second)
+to give the terminal time to process the text properly. The added delay
+is below human perception for chat-style use.
+
+## Original vs Converged
+
+The original spec was a two-line fix: use bracketed paste for all
+messages, add a verifier that retries Enter if stuck. Five review
+rounds caught a series of subtle bugs that would have re-created the
+original failure mode in narrow cases or introduced new ones:
+
+- **Security**: the original sanitization only stripped two specific
+  byte sequences. Reviewers found that 8-bit C1 control characters and
+  UTF-8-encoded C1 characters could escape paste mode equivalently.
+  Sanitization broadened to the full C0/C1 control range.
+
+- **Concurrency**: the original verifier had no way to tell whether the
+  prompt on screen now belongs to the message it's verifying or to a
+  newer message that arrived in between. A per-session sequence number
+  plus an incarnation token (tied to tmux's session creation timestamp)
+  was added so the verifier always knows whether to act or stand down.
+
+- **TUI fragility**: the original assumed `tmux capture-pane` returns a
+  small window of screen content. The external GPT reviewer pointed out
+  that without explicit flags, it could return MB of scrollback and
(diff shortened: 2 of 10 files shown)


JKHeadley (commented):
## Stage 2 Deep Review — Defer to Maintainer (Self-Authored, Over Threshold)

**Reviewer:** Echo (AI) — automated deep review, not an approval. Human maintainer decision required.

### Provenance note
PR #158 was authored by Echo (self-authored agent work, 1362 added lines, above the 1000-line solo-review threshold). Per the precedent from PR #56, Stage 2 does not stand in as an independent reviewer for our own large work — final merge judgment is Justin's. This review documents what was verified and flags anything a fresh reader should know.

### What the change is
Unifies tmux `rawInject` so single-line and multi-line injects share the same bracketed-paste sequence (`\x1b[200~` … `-l text` … `\x1b[201~` + 500ms settle + `Enter`). Closes the qalatra-class race where Claude Code 2.1.x auto-detected fast single-line streams as a paste and consumed the immediately-following Enter as buffer content rather than a submit. Adds module-scope `sanitizeForPaste()` for C0/C1 + DEL bytes (defense against `\x1b[201~` injection in untrusted message bodies). Adds two `sessions:` config knobs (`injectVerifyEnabled`, `promptSigilRegex`).

### Pre-merge verification
- **CI fully green.** All test shards (Node 20/22, 4 shards each), Integration, E2E, Build, Type Check, verify, Vercel Preview. The Contract Tests (Live API) skip is expected on PRs without secrets, not a failure.
- **Test coverage.** 35 new regression tests in `tests/unit/SessionManager-bracketed-paste.test.ts` covering: sanitization (C0/C1/DEL while preserving tab/LF/CR), ANSI strip, secret redaction, unified path wiring, verifier scheduling, config knob defaults. Existing 12 tests in `paste-stuck-detection.test.ts` updated. 6 existing `SessionManager-injection.test.ts` tests pass unchanged.
- **Sanitizer regex.** `[---]` — correctly preserves U+0009 (tab), U+000A (LF), U+000D (CR). U+001B (ESC) is stripped, which is the intended paste-exit defense. Replacement is U+2026 (HORIZONTAL ELLIPSIS), a visible non-flaggable placeholder.
- **Verifier sees what was sent.** `verifyInjection(tmuxSession, sanitized)` passes the sanitized text, so the post-injection compare matches the bytes actually written to the pane. Correct.
- **Spec convergence.** Spec doc + convergence report committed in-PR. Per the description, 5 review rounds (4 internal multi-angle + 1 cross-model GPT/Gemini/Grok). The cross-model round is what we rely on as substitute fresh-eyes scrutiny for self-authored work.
- **upgrades/NEXT.md + side-effects/0.28.93.md.** Release-notes-in-same-PR convention satisfied.

### Risk surface I'd want a human to confirm
1. **All single-line injects now pay the 500ms settle.** Throughput-sensitive paths (rapid agent acks, status pings) get a hard 500ms floor per inject. The PR description acknowledges async conversion was scope-cut to a follow-up. If autonomous-loop / threadline relay flows are sensitive to back-to-back inject latency, worth a deliberate decision before merge.
2. **`promptSigilRegex` is a hot-fixable regex from config.** A bad regex would break every inject's verifier. Reasonable — but worth a maintainer ack that the config path is trusted.
3. **Self-authorship + 1362 lines exceeds the policy threshold from PR #56.** Re-stating for the record so the decision is explicit, not implicit.

### Security / injection scan
The diff itself contains escape sequences and control-byte patterns as code constants — not data, not honored as instructions. No prompt-injection or instruction-override content. The sanitizer is itself an injection mitigation for paste-exit bytes in inbound message bodies.

### Recommendation
**Maintainer decision.** Code looks correct to me, CI is green, multi-perspective external review already happened on the spec. I am not the right reviewer for my own 1362-line PR; Justin to decide whether the cross-model review round is sufficient substitute or whether a separate human pass is wanted before merge.

---
🤖 Stage 2 review by Echo. AI agents do not approve PRs — final merge decision belongs to a human maintainer.

Snailflyer:
The seq + incarnation guarded verifier is the strongest part of this PR. I would make sure its result is not only fire-and-forget telemetry, but also available to the caller that initiated the inject.

The states I would want surfaced are something like:

- `inject_verified_turn_started`
- `inject_rendered_but_unsent`
- `prompt_sigil_unmatched`
- `incarnation_changed`
- `target_session_stale`
- `verifier_timeout`

That keeps the fix honest when this inject path is later used from a remote/mobile control surface. The important product boundary is not "bracketed paste sequence completed"; it is "the same live Claude/tmux session started the intended turn".

Faryo is a narrower phone/browser workbench, but the proof boundary is the same: the control action is only complete when the original live tmux-backed session advances, not when the bridge has written bytes. Reference GIF: https://github.com/Snailflyer/faryo/releases/download/v1.0.7/faryo-public-redacted-same-session-handoff-walkthrough-20260528-0120.gif

EchoOfDawn:
**Re-triaged 2026-07-28 by reading current `main` rather than this PR's problem statement. Its value has genuinely shrunk — but it is not zero, and the shrinkage is worth recording.**

### The headline symptom is no longer silent

This PR's case is *"text visible at the ❯ prompt but never submitted; sessions sat silent until manual Enter."* `main` has since grown three backstops for exactly that:

- **`verifyInjection(...)`** — captures the pane after a delay and sends one extra Enter if the marker text is still at the prompt. Its comment names the precise failure: *"on fresh Claude Code TUIs the Enter after bracketed-paste-end is occasionally eaten."*
- **`recordStrandedDraftMarker(...)`** + the persistent StuckInputSentinel — the durable-across-turn recovery for codex.
- A **2-attempt retry** around the whole injection.

So the failure is now detected and recovered. Fixing the *cause* still beats recovering from it — but the urgency in the description above is stale.

### The security half has been superseded

`sanitizeForPaste()` here strips the whole C0/C1 range, and its doc-comment specifically names *"a body containing `ESC[201~` (or its 8-bit `\x9b` equivalent)"*.

`main` already stripped the 7-bit form (via `june15-headless-spawn-reroute` PR2 finding S2) but **not** the 8-bit one — I measured both payloads against it. That gap is now closed by **#1706**, as a one-regex widening.

I deliberately did **not** adopt this PR's broader C0/C1 strip for that fix: it substitutes an ellipsis for control bytes, which **rewrites legitimate message content** on every injected message. That is a real behaviour change and deserves its own review rather than riding along inside a security patch.

### What's actually left

The **unification** — making single-line injects take the bracketed-paste path so the cause stops occurring, rather than being caught by three nets. Still worth doing, and the change is *simpler* than it looks: it deletes the `if (text.includes('\n'))` branch. Fewer paths, not more.

The real work is that this PR's test asserts the old fixed-sleep shape while `main`'s asserts a per-framework `postPasteDelaySec` ternary, so it must be re-applied to main's current shape rather than merged. Sized earlier at a **97-line** conflict in `rawInject` — the injection chokepoint for every message to every session, which is why I have not rushed it.

### Verdict

**Re-implement against current `main`, at lower priority than when it was written.** Not closing it: unlike #646, #783 and #500, its remaining value has not been delivered elsewhere.

Second time tonight an old PR's real worth turned out to be as a *pointer to live code* rather than a diff to merge — #489 pointed at a cross-machine sync defect (fixed as #1702), and this one pointed at the sanitiser gap (fixed as #1706). Both were found by asking "is this still true of main?" before deciding what to do with the PR.

JKHeadley:
Closing as superseded after a current-main comparison. The original stuck-submit failure is covered by merged PR #159's multi-shot recovery, and the paste-marker injection hazard (including the 8-bit CSI spelling) is covered by merged PR #1706. Current main also has framework-specific Claude/Codex submit timing plus the persistent stranded-draft sentinel. Applying this branch's unified path now would overwrite those newer, deliberately different semantics. The branch and its convergence artifacts remain available if a future live repro shows a gap in the current path.

PR #440, opened 2026-05-27
feat(threadline): LLM topic-name + summary for 'open this' (CMT-567)

JKHeadley:
## Summary
- When "open this" promotes a Threadline conversation into its own topic, an LLM now names the topic from the conversation gist and writes an orientation summary as the first message — replacing the crude slug name + empty "tied to this topic" marker.
- Three-tier degrade (**LLM → deterministic template → slug**) guarantees the bind NEVER fails on LLM trouble. The operator always lands with real context; the LLM just polishes it when available.
- Runs on the shared `interactive` LlmQueue lane (operator is waiting), 15s ceiling, ~1¢/call, governed by the existing daily spend cap.
- Value-pattern credential scrub (actual `sk-…`/`xoxb-…`/key-blocks), not bare english words — so technical topics ("token refresh triage") are named normally while a pasted secret never reaches the chat-list title or first message.
- One shared `briefDeps` built once at startup, threaded to both `onTopicMessage` intercept closures + the `POST /threadline/hub/bind` route. `tie` path unchanged.

## Spec
`docs/specs/THREADLINE-OPEN-THIS-LLM-NAMING-SPEC.md` — converged (2 rounds / 4 reviewers), approved, ELI16 companion shipped.

## Test plan
- [x] Unit: `tests/unit/threadline/openConversationBrief.test.ts` (19) — all 3 tiers, every fallback cause, per-field credential scrub, word-boundary cap, never-empty invariant.
- [x] Unit: `tests/unit/hubCommands.test.ts` (+4) — LLM/template/Tier-C/tie-not-invoked.
- [x] Integration: `tests/integration/threadline/hub-bind-routes.test.ts` (+3) — LLM happy path; LLM-throws → 200 + template (regression guard); intelligence:null → 200.
- [x] Full threadline suite: 1580 passing, zero regressions. `tsc --noEmit` + `pnpm build` clean.
- [x] **Test-as-self (real LLM, no mocks):** harness against real Claude caught a production-breaking timeout (3.5s vs measured ~8-10s) → raised default to 15s. Then live-deployed on instar-codey (Codex/gpt-5.2): real "open this" produced `topicName="Mentor ledger dedup strategy"`, server log `nameSource=llm summarySource=llm`. Test topic deleted, Codey restored.

### DIFF
--- a/docs/specs/THREADLINE-OPEN-THIS-LLM-NAMING-ELI16.md
+++ b/docs/specs/THREADLINE-OPEN-THIS-LLM-NAMING-ELI16.md
@@ -1,0 +1,41 @@
+# THREADLINE OPEN-THIS LLM NAMING — plain-English overview
+
+**What problem we're fixing.** Last night I made "open this" reliable — say it in the Threadline topic and a topic gets created and bound. But what you got was a topic with:
+
+1. A bad name — just the first 5 words of whatever the other agent said ("Hey echo quick check on"), not what the conversation is actually *about*.
+2. An empty room — the only message in the new topic was the generic "🧵 This conversation is now tied to this topic" tie-marker, with zero context about who you're talking to or what's going on.
+
+That's the opposite of what "open this" is supposed to do — you opened it because you wanted a *workable* topic, not an unlabelled empty one.
+
+**What I'm building.** Two strictly additive improvements, on the same deterministic "open this" path:
+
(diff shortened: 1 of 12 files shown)


JKHeadley:
Superseded by #467 (rebased onto current main; #440's CI never triggered).

PR #452, opened 2026-05-28
fix(SourceTreeGuard): extend sourceTreeReadOk to readSync + expand verb set

JKHeadley:
## Summary

Dogfood follow-up to PR #450. v1.3.38 deployed to Echo and the watchdog's `fetch` succeeded — but every downstream readonly verb (`rev-parse FETCH_HEAD`, `log`, `merge-base`, `ls-tree`, `show`, `remote -v`) still tripped `SourceTreeGuard`. `readSync` runs source-tree-checks as defense-in-depth (catches repo-local aliases that could rebind a read-only verb), so the PR #450 bypass — which only opened `execSync` + `spawn` — left the canonical-ref read path blocked.

This PR closes the gap at the right layer:
- `readSync` honors `opts.sourceTreeReadOk` the same way `execSync` does.
- `SOURCE_TREE_READ_TIER_VERBS` expanded from `['fetch']` to also include `rev-parse`, `ls-tree`, `show`, `log`, `cat-file`, `merge-base`, `remote`. Still a small closed read-tier set; still requires a spec edit to grow. Every verb is read-tier — none mutate the working tree or committed refs.
- Layer B wiring + Layer C scanner pass `sourceTreeReadOk: true` on every `SafeGitExecutor.run` callsite (not just the fetch).

## Test plan

7 unit tests:
- default: fetch on non-source / source / etc. (4 from PR #450)
- closed-set invariant — destructive verbs (commit/push/reset/checkout/rebase/merge/clean/rm/branch/tag) are NOT present; size ≤ 10
- **readSync path WITH `sourceTreeReadOk`: `rev-parse FETCH_HEAD` on source tree → passes**
- **readSync path WITHOUT the opt: `rev-parse FETCH_HEAD` on source tree → STILL blocked**

Lint clean.

### DIFF
--- a/src/core/SafeGitExecutor.ts
+++ b/src/core/SafeGitExecutor.ts
@@ -142,7 +142,18 @@ export const READONLY_GIT_VERBS: ReadonlySet<string> = new Set([
  * source-tree guard.)
  */
 export const SOURCE_TREE_READ_TIER_VERBS: ReadonlySet<string> = new Set([
+  // data-pull verbs (execSync path)
   'fetch',
+  // read-tier verbs the watchdog + reconciler need against the source tree
+  // (readSync path — defense-in-depth source-tree check also gates these,
+  // so the bypass must work for both methods)
+  'rev-parse',
+  'ls-tree',
+  'show',
+  'log',
+  'cat-file',
+  'merge-base',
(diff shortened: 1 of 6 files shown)


JKHeadley:
Replaced by #455 — clean cherry-pick onto latest main avoids the NEXT.md rebase tangle.

PR #457, opened 2026-05-28
docs(spec): Part 2.1 — `instar test-as-self` orchestration (draft for review)

JKHeadley:
Sub-spec to the approved [`SELF-PROPAGATION-HARNESS-SPEC.md`](docs/specs/SELF-PROPAGATION-HARNESS-SPEC.md). Draft for Justin's review (`approved: false` in frontmatter).

**Context.** Part 1 (poll-ownership lease) shipped PR #446 — verified live on Echo. Part 2 v1 shipped PR #448 — runbook + deterministic verifier. v1 explicitly deferred the three pieces that turn the runbook from "if a human does the recipe right" into "one button does it": auto-mint bot via Secret Drop, full Playwright Telegram round-trip, the `instar test-as-self` CLI command itself.

**Why now.** PR #428 (cross-machine seamlessness) is one live two-machine test away from merge. That test is exactly the kind of hand-done deploy that bit us on 2026-05-27. Building Part 2.1 IS the path to closing #428.

**Surface locked.** CLI flags + reject conditions (Bob block, canonical-home block, raw-token block) + seven gated steps + crash-capture wiring + all-three-tier test plan + migration-parity path.

**Open question for Justin (in spec):**
- A) Full Part 2.1 (~1 day) — best
- B) One-button command without Playwright round-trip (~0.5 day)
- C) Defer entirely; ship PR #428 manually first

Leaning A — the parent spec's whole argument was that hand-done deploys are the failure mode.

**ELI16 companion:** [`SELF-PROPAGATION-HARNESS-PART-2-1-SPEC.eli16.md`](docs/specs/SELF-PROPAGATION-HARNESS-PART-2-1-SPEC.eli16.md)

This PR is doc-only (no src/ changes), so it does not require the instar-dev gate's full spec→trace→side-effects pipeline. Once Justin approves the scope, the implementation lands on a separate src/ PR with the full gate satisfied.

### DIFF
--- a/docs/specs/SELF-PROPAGATION-HARNESS-PART-2-1-SPEC.eli16.md
+++ b/docs/specs/SELF-PROPAGATION-HARNESS-PART-2-1-SPEC.eli16.md
@@ -1,0 +1,63 @@
+# ELI16 — `instar test-as-self` (Part 2.1)
+
+## What broke that led to this
+
+Last week we tried to deploy a copy of "me" onto a second machine over
+Telegram to verify cross-machine seamlessness. The deploy was hand-done:
+mint a bot, copy files, start things, send a probe, watch logs. Two
+things went sideways at once — a Telegram polling conflict and a node
+crash — and because the deploy was hand-done, it was hard to tell which
+failure was real and which was the operator. We fixed the conflict at
+the structural level (the poll-ownership lease, shipped). We shipped a
+basic verifier (shipped). But the actual deploy is still hand-done, so
+the same kind of confusion will happen the next time.
+
+## What this spec does
+
+Adds one new command: `instar test-as-self`. Run it, and it does all
+seven steps of the deploy/verify/teardown loop for you: get a test bot
+token (the safe way, never typed in chat), set up a throwaway agent home
+that isn't your real one, deploy the current build into it, start it
+up, send a real probe message and confirm the throwaway agent replied,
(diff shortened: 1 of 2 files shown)


EchoOfDawn:
Closing as **superseded — the work shipped, this spec did not, and the spec now misdescribes what was built.**

Verified against `origin/main` rather than assumed:
- `src/cli.ts:2130` carries `.command('test-as-self')`, implemented in `src/commands/test-as-self.ts`. The one-button CLI (item 3) exists.
- `.claude/skills/test-as-self/SKILL.md:12` is headed "The one-button path (Part 2.1 — use this first)" and documents Secret-Drop bot acquisition (item 1) plus the seven gated steps.
- So all three deferred items this sub-spec was written to define are built.

**Why cancel rather than land it as a record.** Item 2 proposed a *full Playwright Telegram round-trip*. The shipped implementation deliberately does not do that — SKILL.md:20: "The round-trip uses the Telegram Bot HTTP API directly (not Playwright) — no browser, no flake." Landing this spec now would put an approved document in `docs/specs/` prescribing an approach the implementation weighed and rejected, which is worse than having no spec: a future reader would treat it as the contract.

The two files were never on main (checked by path, not by name — a `test-as-self` name match on main is the *skill*, not this spec).

No work is lost. Closing per the "merged or explicitly cancelled, no third state" rule.

PR #459, opened 2026-05-28
fix(release-readiness): demote eval-failure to housekeeping default (sentinel-trio standard)

JKHeadley:
## ELI16 — a watchdog was telling you when IT broke, not when YOUR release broke

The release-readiness watchdog checks whether finished work is piling up unreleased. That check is
useful and it stays exactly as it is.

The problem was what happened when the watchdog itself fell over — it could not reach the repo, or its
analyzer returned nothing. It treated its own plumbing failing as news for you, and posted it to your
attention queue. So a network blip became a notification about a release that might be perfectly fine.

That is backwards. When a watchdog cannot see, the honest report is "I could not check", and that
belongs in the log where someone debugging it will look — not on your phone.

After this, evaluator-self-failures go to `logs/sentinel-events.jsonl` and `server.log` by default. If
you actually want them escalated, `monitoring.releaseReadiness.escalateEvalFailures` turns them back
on.

**What does not change:** the signal you care about. "Release blocked — unreleased work is piling up"
still posts to Attention exactly as before. This only stops the watchdog reporting its own indigestion.

It also cleans up after itself: a migration strips the stale `release-readiness-eval-failure-*` rows
this behaviour already left in existing agents' attention queues, so nobody has to dismiss them by
hand.

*(Refreshed against main 2026-07-28 — it was 2,193 commits behind. Three conflicts, all cases where
main added something next to what this branch added; both sides kept.)*

## UX Impact

Who sees it: operators who receive release-readiness Attention items and maintainers reading sentinel diagnostics.

What the user sees: evaluator plumbing failures stop creating chat noise by default; actionable release-blocked alerts are unchanged. The explicit `escalateEvalFailures` setting restores the old escalation when desired.

First contact: the quieter default takes effect on the first watchdog evaluation after installation, and startup migration removes stale evaluator-failure rows.

---

## Summary

- Demotes `ReleaseReadinessSentinel.failLoud()` evaluator-self-failures (fetch / analyzer / top-level tick stages) to **audit-only by default**, gated behind a new `monitoring.releaseReadiness.escalateEvalFailures` opt-in. Brings the sentinel into compliance with the silently-stopped-trio standard (post-2026-05-22 topic-spam fix): internal-plumbing failures belong in `logs/sentinel-events.jsonl` + `server.log`, not on the user's Telegram surface. The user-actionable "Release blocked — unreleased work is piling up" signal is unaffected — it still posts to Attention.
- Adds `migrateRetireStaleReleaseReadinessEvalFailureAttention` in `PostUpdateMigrator` to strip stale `release-readiness-eval-failure-*` rows from existing agents' `.instar/state/attention-items.json` on update. Atomic write + idempotent + preserves legitimate user-actionable rows.
- Adds 13 new tests (6 sentinel + 7 migrator) covering both halves of the new contract and a regression guard.
- Includes a post-mortem (`docs/postmortems/2026-05-27-release-readiness-eval-failure-topics.md`) tracing the spec-time framing error ("loud-attention vs silent-catch" as a binary, missing the housekeeping path) and listing the structural follow-ups (sentinel-emit-site lint, `SentinelEmitter` primitive, spec-template "failure-mode emit-site table", `/spec-converge` cross-reference rule, spec text correction PR).

Origin: dogfood feedback on Echo (2026-05-27). User observed two "Release-readiness check could not evaluate" topics within 90 minutes (one fetch-stage, one analyzer-stage) with bodies that were inscrutable ("analyze-release returned no report") and called it out as anti-Instar topic clutter. Echo's two stale rows were live-cleaned via `DELETE /attention` before this PR.

## Test plan

- [x] `tests/unit/ReleaseReadinessSentinel.test.ts` — 16 tests including the 6 new fail-loud cases (housekeeping default + escalate opt-in + user-actionable regression guard)
- [x] `tests/unit/PostUpdateMigrator-retireStaleReleaseReadinessEvalFailureAttention.test.ts` — 7 tests
- [x] `tests/unit/releaseReadinessWiring.test.ts` — 8 tests pass unchanged
- [x] `tests/integration/release-readiness-routes.test.ts` — 5 tests pass unchanged
- [x] `tests/e2e/release-readiness-live.test.ts` — 2 tests pass unchanged
- [x] `tsc --noEmit` clean
- [x] Pre-commit gate green (instar-dev trace + spec + side-effects artifact + ELI16 verified)


EchoOfDawn:
**Refreshed against `main` and now passing everything except one check. That check cannot be satisfied honestly, and I'm not going to fake it.**

Current state: 25 checks passing — build, type-check, all eight unit shards on both Node versions, integration, e2e, repo invariants, ELI16, release-fragment, class-closure. One failure: **decision-audit**.

That check requires a `.instar/instar-dev-decisions/<ts>-<slug>.json` record proving the local instar-dev gate evaluated the in-scope change. This PR's commits predate that mechanism, so no such record exists — and the record is *evidence that an event occurred*. The event never happened.

**One API call would make this green.** I'm not making it: the check exists to prove that gate evidence is real, so minting the artifact leaves it green while lying about the single thing it certifies.

**I looked for a legitimate route rather than just declining:**

- The checker's exemptions (`scripts/decision-audit-presence-check.mjs`) are bot-author, a `chore: release` title, and zero in-scope files. None applies.
- A record *can* be minted honestly — a real in-scope commit makes the local gate run and write one. That is how #1350 came to pass this check tonight. But it requires genuine work to ride on, and this PR has none left: I ran the full lint chain over the refreshed tree separately and it is clean.
- Manufacturing a commit purely to trigger the gate would be the same forgery in a different costume.

**The distinction that will recur on every pre-dated PR:** a gate wanting an *artifact* (an explanation, an ELI16 overview) is honestly satisfiable at any later date — the explanation is as true now as then, and I wrote several tonight. A gate wanting *evidence an event occurred* is not.

I also checked whether the evidence merely exists elsewhere. It does not — no gate records exist from that period in either the per-entry or legacy JSONL format; persistence began mid-June. So this PR very likely *did* pass the local gate at a time when nothing wrote it down.

**Four ways forward, all the operator's:** exempt PRs whose base predates the mechanism (the checker already carries an exemption list), accept the PR's written attestation, require a genuine re-run, or close it. Filed with the reasoning rather than left as a red check with no explanation.

PR #489, opened 2026-05-28
fix(multi-machine): six real-hardware bootstrap defects + tests

JKHeadley:
## ELI16 — a machine that just joined can steal a job from a machine that is still working

When I run on more than one computer, exactly one of them is "in charge" at a time. That is tracked
by something like a badge with a timestamp: whoever holds it is in charge, and they keep refreshing
it to show they are still alive.

When a NEW machine joins, it copies the shared state from a repository. That copy includes the badge
— but with the timestamp from whenever the copy was made, which can be old. So the new machine wakes
up, looks at a stale timestamp, concludes "the machine in charge has not checked in for ages, it must
be dead", and takes the badge. The other machine is not dead. It is working. Now both think they are
in charge.

The fix is one line of ordering: fetch the current state BEFORE deciding whether the other machine is
dead, instead of deciding from the copy you booted with.

This PR was opened in May with six fixes from a real two-machine bring-up. During the July rebase,
four were initially judged present on main. Failure-first validation showed that two production
paths were in fact still absent: GitSync's usable signing-key probe and its clean-exit autostash
conflict recovery. This revision restores both. The two fixes whose complete versions are already
on main remain dropped rather than reapplying partial predecessors.

Two commits carrying an unrelated spec were also split off: that spec's own commit message says it is
not converged and is waiting on a design decision, so it does not belong inside a bug-fix PR.

Driving the live cross-machine seamlessness proof on real hardware (laptop + Mac mini, Bob untouched) surfaced **six distinct bootstrap defects** the overnight Track A–F work did not close. Each was verified on the live mesh before fixing, and each ships with a regression test.

Spec: `docs/specs/MULTI-MACHINE-BOOTSTRAP-ROBUSTNESS-SPEC.md` (Round 2, approved topic 13481).
Side-effects: `upgrades/side-effects/mm-bootstrap-realhw-fixes.md`.

## The six fixes

1. **No-upstream silent push** — `GitSync.commitAndPush` upstream-aware (`-u origin <branch>` on first push) + surfaces real push failures; `init` connect-existing sets push tracking. Closes a bug where the awake machine's lease epochs committed locally but never reached origin, with zero logs → entire cross-machine sync silently dead.
2. **Join scaffolds no config.json** — `join` writes a complete machine-local config (fresh authToken, port; new `--port`) instead of leaving the standby unauthenticated on defaults.
3. **Commit-signing broke every commit** — `GitSync.configureCommitSigning` prefers canonical `signing-key.pem`, supports legacy `signing-private.pem`, derives the public key, test-signs for real, and only enables signing if the probe works. Failure explicitly leaves commits unsigned-but-working; verification remains a no-op stub.
4. **Registry conflict-marker corruption** — `machines/registry.json` was mis-classified as `llm` by a cwd-relative path, leaving the deterministic merge dead. Separately, a clean-exit `--autostash` pop conflict could leave unmerged files even though pull returned success. Both paths now route through deterministic resolution.
5. **Standby push starvation** — sync push rebases-and-retries on non-fast-forward so a standby converges within one sync despite awake lease churn (kept out of the lease CAS path).
6. **`wakeup --force` propagation** — claims a real signed +1-epoch lease (not just the local role, which the server reverts) and pushes it.

## Tests

- Failure-first reproduction: three focused failures (two signing paths and post-autostash conflict recovery) before the restore; all pass after it.
- 187 related GitSync, lease, coordinator, join, wakeup, and multi-machine unit/e2e tests pass.
- Full lint and TypeScript checks pass.\n- Rebased CLI failure reproduced: the duplicate `join --dir` registration made three built-CLI assertions fail before dispatch; retaining main’s single `--dir` plus the new `--port` makes all three pass.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-29T04-45-45-261Z-mm-bootstrap-realhw-fixes.json
+++ b/.instar/instar-dev-decisions/2026-07-29T04-45-45-261Z-mm-bootstrap-realhw-fixes.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-29T04:45:45.261Z",
+  "slug": "mm-bootstrap-realhw-fixes",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new config key added"
+  ],
+  "belowFloor": false,
+  "files": 7,
+  "loc": 260,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-29T07-49-11-848Z-mm-bootstrap-realhw-fixes.json
+++ b/.instar/instar-dev-decisions/2026-07-29T07-49-11-848Z-mm-bootstrap-realhw-fixes.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-29T07:49:11.848Z",
+  "slug": "mm-bootstrap-realhw-fixes",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 104,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-07-29T07-50-10-381Z-mm-bootstrap-realhw-fixes.json
+++ b/.instar/instar-dev-decisions/2026-07-29T07-50-10-381Z-mm-bootstrap-realhw-fixes.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-29T07:50:10.381Z",
+  "slug": "mm-bootstrap-realhw-fixes",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 104,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-29T08-02-57-985Z-feat-nuke-here.json
+++ b/.instar/instar-dev-decisions/2026-07-29T08-02-57-985Z-feat-nuke-here.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-29T08:02:57.985Z",
+  "slug": "feat-nuke-here",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 1,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-07-29T08-05-42-659Z-mm-bootstrap-realhw-fixes.json
+++ b/.instar/instar-dev-decisions/2026-07-29T08-05-42-659Z-mm-bootstrap-realhw-fixes.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-29T08:05:42.659Z",
+  "slug": "mm-bootstrap-realhw-fixes",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 1,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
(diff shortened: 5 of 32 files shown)


EchoOfDawn:
**Triage note — NOT stale, but not mergeable as-is either.** Leaving this here so the next person does not re-derive it.

I went through the six defects individually rather than treating the PR as one unit, because a bundle of independent fixes ages into independent relevance questions.

| # | defect | status on `main` today |
|---|---|---|
| 1 | no-upstream silent push | **STILL LIVE** — see below |
| 2 | `join` scaffolds no config.json | unclear; `joinMesh` is at `machine.ts:265`, I did not find the scaffolding |
| 3 | commit-signing / key filename | **looks fixed** — `signing-key.pem` is used in `machine.ts` and `server.ts` |
| 4 | registry conflict corruption | **fixed** — `src/core/mergeRegistry.ts` exists and `GitSync.ts:1056` classifies the path deterministically |
| 5 | standby push starvation | not checked |
| 6 | `wakeup --force` propagation | not checked |

**Defect 1 is real and I have read it on `main`.** `GitSync.commitAndPush` (lines 519–543):

```ts
if (this.autoPush) { this.gitExec(['push']); }   // bare push, no upstream
...
} catch {
  // @silent-fallback-ok — push failure boolean return
  return false;                                   // same value as "nothing to commit"
}
```

Two compounding problems: there is **no `set-upstream` anywhere in that file** (grep for `set-upstream` / `-u origin` / `branch --set` returns zero on `origin/main`), and `false` is returned for *three* different situations — no dirty paths, nothing staged, and **the push threw**. A caller cannot distinguish "nothing to sync" from "the sync failed".

That is exactly the consequence this PR reported from real hardware: lease epochs committing locally and never reaching origin, with zero logs. Filed separately as a live defect against `main` so it is tracked independently of this PR's fate.

**Why it should not simply be merged.** Defects 3 and 4 have landed on `main` by *different* routes. Merging this wholesale would re-apply fixes that already exist in another shape, which is how you end up with two implementations of one repair.

**Recommendation: split.** Extract the still-live fixes (certainly 1; check 2, 5, 6) into a focused PR against current `main`, and drop the superseded ones. The surrounding code has moved 2,109 commits, so a conflict-resolution pass on the original six would be reconciling against a tree that has already solved a third of them.

Not closing it — unlike #646, #783 and #500, this one still carries live value.

JKHeadley:
Current-main disposition review (2026-07-28): **needs Echo judgment; preserving this PR and its branch.** This bundle is partly superseded and partly still live, with heavy conflicts against today's multi-machine architecture.

Already independently covered on main: the no-upstream push fallback, exactly-once defaults, and resilient lease-renewal scheduling. Still-live defects include join-time config scaffolding, PKCS8-aware SSH signing setup, post-autostash clean-exit conflict handling/path normalization, standby non-fast-forward retry, and fenced lease ownership for forced wakeup.

The branch's final design note explicitly records non-convergence, and merging the mixed bundle would reintroduce old lease/default assumptions alongside the remaining fixes. The recoverable path is to split the live defects into current-main-sized PRs with their own tests and contracts. I have not closed or rewritten this PR, so its implementation and research remain intact for Echo's judgment.

EchoOfDawn:
**Do not merge — this branch is currently missing about half of its own GitSync work, and that is my error from the rebase.**

Rebasing this onto current main (2,182 commits) I hit a conflict in `src/core/GitSync.ts` and resolved it with `git checkout --ours`. The conflict presented as a single hunk where main had the upstream-aware push and this PR's side was empty, so "take main" looked correct for that hunk — and it was.

But `--ours` on a **file** takes main's entire file. This PR's original commit changed `GitSync.ts` in **six** hunks; only one conflicted. Resolving that one discarded the other five.

Measured: of 111 distinctive identifiers the original commit added to `GitSync.ts`, **58 are absent** from the rebased branch. Concretely missing:

- **Fix 3 (commit signing)** — `ssh-keygen` derivation of the `.pub` sibling, the `.sign-probe` test-sign, and the `signingWorks` guard that leaves signing OFF rather than failing every commit. Absent from main too, so it is genuinely lost, not superseded.
- **Fix 5 (standby push starvation)** — `commitAndPushWithRebaseRetry`.
- The registry-resolver placement change, which is why `git-sync-autostash-registry` fails: its whole point is that the deterministic resolver must run outside the catch block, and that hunk is gone. The `FileClassifier` half of that fix survived, which is why the failure looks like a classification bug rather than a missing call site.

What IS correct on the branch and worth keeping:
- the bug #7 fresh-join lease priming (`primeFromDurable`/`syncDown`), merged additively into main's richer lease-init block
- the `FileClassifier` cwd-relative path fix
- join config scaffolding, registry auto-merge, `wakeup --force` lease claim
- the duplicate `-d, --dir` option on `join` removed (main gained its own in #661)
- the three genuinely superseded fixes correctly dropped, documented per-item in the side-effects addendum

**The remaining work is to re-apply the five non-conflicting `GitSync.ts` hunks from `cad97b115` onto the current file, then confirm `git-sync-autostash-registry`, `git-sync-commit-signing` and `git-sync-push-converge` pass.** `git show cad97b115 -- src/core/GitSync.ts` is the source; each hunk needs checking against main's current shape rather than applying blind, since main rewrote the push path around it.

Flagging rather than fixing because I have been hands-on far longer than I should be tonight, and this is a well-specified piece of work rather than an open question.

PR #497, opened 2026-05-29
fix(attention): lock down topic-spam — per-source+global forum-topic circuit breaker + collaboration-redrive housekeeping fix

JKHeadley:
## What & why

Second Telegram topic-flood (after the 2026-05-22 sentinel flood). An agent's `CollaborationRedriveEngine` raised one **"can't reach `<peer>` — unknown routing"** attention item per failed peer-resolution every sweep, forever (it reset its strike counter to 0 after each escalation), and `createAttentionItem` spawns a **brand-new forum topic per item** → a wall of topics. The empty `known-agents.json` on that machine made every peer unresolvable.

Fixed in two layers — *fix the offender AND make the class of bug impossible*:

### 1. Structural backstop — `AttentionTopicGuard` (the lockdown)
A per-source **and global** circuit breaker at the one chokepoint (`TelegramAdapter.createAttentionItem`). Once a non-critical source — or, via the **global ceiling**, all sources collectively — exceeds its topic budget within a rolling window, further items **coalesce into ONE reused notice topic** and are recorded in `state/attention-suppressed.jsonl` (size-capped). Never a wall.
- **Global ceiling** makes the "any mis-wired feature is auto-throttled" guarantee hold even when a feature varies its `sourceContext` per item.
- **HIGH/URGENT/CRITICAL (case-insensitive) are never coalesced** — critical messages always get their own topic.
- **No item is dropped** — coalesced items are flagged and excluded from the per-item topic maps, so resolving one never closes the shared topic and a restart can't corrupt the reverse map.
- **No double-create race** — concurrent coalesced items for a bucket share one in-flight topic creation.
- **Config validated** — a `NaN`/negative value can't silently disable it.
- **Default-ON in code** → every fleet agent is protected on the dist update with zero config. Rollback = `messaging[].config.attentionTopicGuard.enabled=false`.

### 2. Offender fix — `CollaborationRedriveEngine`
- "can't reach / unknown routing" is now **log-only housekeeping** (no attention item, no topic, no re-fire).
- A 32–64-char-hex `relatedAgent` resolves directly as the routing fingerprint **only when it matches a known agent's `publicKey`** (trust-gated — never plaintext-sends to an unverified address; CMT-663).

## Review

Converged via a **4-reviewer panel** (adversarial/security, integration, architecture, scalability). All **7 material findings fixed in code, none deferred**. See `docs/specs/attention-topic-flood-guard.{md,eli16.md,convergence.md}` and `upgrades/side-effects/attention-topic-flood-guard.md`.

## Tests (all 3 tiers)
- **Unit** — `AttentionTopicGuard` (11: budget/global-cap/critical-bypass/episode/eviction/config-validation), `CollaborationRedriveEngine` (+3: housekeeping log-only, hex-known-routes, hex-unknown-blocked), `PostUpdateMigrator-topicFloodGuard` (2).
- **Integration** — real `TelegramAdapter`: flood capped at budget+1 topics, HIGH bypasses, concurrent-coalesce → one topic, different source unaffected, no item dropped, audit log populated.
- **E2E** — fleet default (NO config) still caps a flood exactly (migration-parity).

Migrator backfills a "Topic-Flood Guard" awareness section (idempotent). `collaborationRedrive` keeps its ship-OFF default.

### DIFF
--- a/docs/specs/attention-topic-flood-guard.convergence.md
+++ b/docs/specs/attention-topic-flood-guard.convergence.md
@@ -1,0 +1,82 @@
+# Convergence Report — Attention Topic-Flood Guard
+
+**Spec:** `docs/specs/attention-topic-flood-guard.md`
+**Method:** four independent reviewer agents, one round, distinct lenses
+(adversarial/security, integration/over-block, architecture/signal-vs-authority,
+scalability/correctness), auditing the spec against the actual diff.
+**Date:** 2026-05-28
+**Outcome:** converged after addressing all material findings in code; no findings
+deferred (per the /instar-dev no-deferrals rule).
+
+## Findings and resolutions
+
+Severity in brackets. ✅ = fixed in code this iteration; 📝 = addressed in spec text.
+
+1. **[HIGH] High-cardinality `sourceContext` defeats the per-source budget AND
+   leaks the tracking maps** (security + scalability reviewers). A mis-wired
+   source varying its key per item dodged the budget — breaking the guard's
+   central "any mis-wired feature is throttled" promise — and grew the `events`
+   map unbounded. ✅ Added a **global ceiling** (`maxTopicsGlobal`, default 8):
+   once total topic creation in the window exceeds it, non-critical items of any
+   source coalesce into one shared `'*'` bucket. Added stale-key eviction + a hard
+   `maxTrackedSources` cap. Unit test: `GLOBAL cap defeats source-key variation`.
+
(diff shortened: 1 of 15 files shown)


JKHeadley:
Superseded by the rebased PR off current main (v1.3.71). After #495 landed the redrive offender fix, this branch conflicted on the redrive engine; the new PR ships only the structural AttentionTopicGuard backstop (the complementary piece #495 doesn't have).

PR #500, opened 2026-05-29
feat(monitoring): CrossSessionCoordinator — light advisory cross-session signal

JKHeadley:
## What & why

Concurrent Claude Code sessions on one agent home share `.instar/` state but are **blind to each other**. The damaging incident (2026-05-28, topic 15579): one session built a fix while another "hit the safety brake" — flipped a feature flag off and mass-withdrew 19 commitments. Both reached a correct local diagnosis; neither knew the other was acting. Net: bug fixed, engine left off, test bed wiped.

Justin approved a **LIGHT** fix (chosen from a light/medium/heavy menu): a visible "I'm about to do X" signal, **no hard locks** — "learn to collaborate slowly and smoothly."

## The fix

`CrossSessionCoordinator` — a shared, append-only scratchpad of recent high-impact structural actions + voluntary intents. Any structural action surfaces *other* recent entries by a **different/unknown** session as an advisory `coordinationWarning`. **It never blocks and never mutates target state** — purely advisory.

- Routes: `POST /coordination/intent`, `GET /coordination/recent`
- Backstop auto-record (no reliance on the agent remembering to announce): sensitive `PATCH /config` flips + `POST /commitments/:id/withdraw` record themselves and attach `coordinationWarning` to their own response
- Wired in `AgentServer` (always alive → GET 200 not 503) + `CapabilityIndex` entry
- Default-ON housekeeping, near-silent (no Telegram); audit at `logs/cross-session-events.jsonl`
- Actor resolution is **session-level only** (`X-Instar-Session`), never the agent id (which all sessions share and would wrongly suppress the warning)

## Design decisions (see convergence report)

- **Single-source recording** of withdrawals in the route handler (not the `CommitmentTracker` event) — avoids double-count; the route is the single agent-facing path and gives the inline warning for free.
- **Intents are never deduped** (events, not states). A regression test locks this — an earlier dedup keyed on `kind+target+value` collapsed two distinct intents into one and suppressed the warning; caught during implementation review and fixed.

## Migration parity

Config default in `ConfigDefaults.SHARED_DEFAULTS` (propagates via `applyDefaults`), `migrateClaudeMd` awareness section + `generateClaudeMd` template. All unit-tested.

## Tests (all three tiers + parity)

- Unit (16) · Integration (8, both incident vectors over live HTTP) · E2E lifecycle (6, real AgentServer → feature alive) · migration-parity (5) · ConfigDefaults assertion. Green locally.

## Out of scope (deliberate)

- Incident #1 stale `active:true` liveness ghost — a separate staleness bug, candidate next step.
- Hard locks / leader election — the heavy path Justin declined for now.

Spec: `docs/specs/cross-session-coordination.md` (converged + approved) · ELI16: `docs/specs/cross-session-coordination.eli16.md` · side-effects: `upgrades/side-effects/cross-session-coordination.md`


EchoOfDawn:
**Triage 2026-07-28: KEEP — I expected to cancel this and reading it reversed that.**

The cheap call was "superseded by the parallel-work family" — `src/core/ParallelActivityIndex.ts`, `src/core/PoolActivityView.ts`, `src/monitoring/ParallelWorkOverlap.ts` all shipped since May. Reading this spec's problem statement shows they solve a *different* problem:

- Those are **topic- and pool-level read surfaces** — "what am I working on across topics / which machine runs this." They answer a question you ask.
- This is a **pre-action scratchpad for high-impact structural actions** in a single agent home, surfacing *other recent* entries to an actor before it acts durably. It answers a question you did not think to ask.

The originating incident is the distinction: one session built the proper fix while a parallel session hit a "safety brake" — flipped a config flag to `false` and mass-withdrew ~19 active commitments. Both reached a correct local diagnosis; neither knew the other was acting. A topic-focus index would not have caught that, because neither session was asking about topics.

**The problem is still live.** `src/monitoring/CrossSessionCoordinator.ts` and every new test/spec file here are absent from `origin/main`. And I hit this class again today: two of my own hands contended for the single test-suite lane, and I separately duplicated work already assigned to a peer agent — both from the same blindness this PR addresses. `CommitmentTracker.mutate()`'s per-commitment CAS still prevents torn writes, not opposing policies.

Carries `approved: true`, `approver: justin`, and a real convergence report (3 iterations).

**Blockers to landing, stated honestly:** it is CONFLICTING against current main and touches ~20 files, so this is genuine conflict-resolution work, not a rebase-and-merge. Its checks last ran 2026-05-29. Keeping it means scheduling that work, not merging it today.

Not closing.

EchoOfDawn:
Closing as **superseded by a more complete implementation of the same idea**, with the evidence — a stale PR deserves a reason rather than a silent close.

This PR proposed a `CrossSessionCoordinator` for the problem that concurrent sessions on one agent home share `.instar/` but are blind to each other — the 2026-05-28 incident where one session built a fix while another flipped the flag off and withdrew 19 commitments, each with a correct local diagnosis and no idea the other was acting.

`main` now solves that, and went further than this design did:

| piece | on main |
|---|---|
| cross-topic activity index | `src/core/ParallelActivityIndex.ts` |
| pure overlap detector | `src/monitoring/ParallelWorkOverlap.ts` |
| proactive councilor | `src/monitoring/ParallelWorkSentinel.ts` |
| pool-wide view | `src/core/PoolActivityView.ts` |
| read surface | `GET /parallel-work/activities` |
| PR/spec-level claims | `src/commands/devClaimCheck.ts` |

The sentinel's own header states the goal in this PR's terms:

> *"…emits ONE 'overlap' nudge so the agent learns its other hand is already on the same work (the willpower-trap fix: the signal comes TO the agent, it doesn't rely on the agent remembering to query)."*

That is the part this PR could not have delivered as an advisory read alone. It also carries containment this design did not specify — pair-keyed cooldowns that survive focus edits, so responding to a nudge cannot re-trigger it.

`devClaimCheck` covers the adjacent gap too: topic-level awareness does not see PR- or spec-level claims, which is where the duplicate-build collisions actually happened.

Merging this now would land a second coordinator beside a working one.

**What I am not claiming.** I checked that the problem is solved, not that every line here has an equivalent. If some specific behaviour in this branch is still missing from the parallel-work stack, it is worth re-proposing against current `main` as its own change — the surrounding code has moved 2,079 commits.

Reopen if I have misjudged it; nothing destructive here and the branch is untouched.

Closed while working the trailing edge of open `echo/*` PRs, checking each for *relevance* before spending effort making it merge cleanly. Third of the batch to fail that check, after #646 and #783.

PR #508, opened 2026-05-29
fix(pre-push): resolve smoke base and cap broad runs

JKHeadley:
## Summary
- resolve pre-push smoke base from branch upstream/push remote before canonical main fallbacks
- log chosen base and changed-file count before Vitest selection
- skip local smoke when changed/test set is too broad and leave CI as the authority

## Validation
- npm test -- tests/unit/scripts/pre-push-scope.test.ts
- node scripts/pre-push-smoke.mjs
- npm run lint

Published with the local pre-push skip flag because this branch fixes that hook path.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -13,7 +13,7 @@ if ! node scripts/pre-push-fixture-guard.mjs; then
   exit 1
 fi
 
-# ── Smoke tier: run only tests affected by files changed vs origin/main ──
+# ── Smoke tier: run only tests affected by files changed vs the resolved base ──
 # Full suite still runs in CI across 8 sharded runners — that's the authority.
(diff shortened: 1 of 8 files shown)


JKHeadley:
✅ Reviewed + approving (recording as a comment — GitHub blocks formal self-approval since we share an account). Resolver + guard logic and wiring all check out: resolvePrePushBase prefers branch upstream / push remote, falls back JKHeadley/main → upstream/main → origin/main (each rev-parse --verify'd + deduped), so agent worktrees with a stale origin now diff against the canonical base. Two-stage breadth guard + fail-open + 'CI is the authority' messaging are exactly right, and the smoke runner genuinely calls the resolver/guard (not dead code). Holding failed-files-only retry for the follow-up was correct. Merging.

PR #511, opened 2026-05-29
fix(worktree): resolve agent-home instar checkouts

JKHeadley:
## Summary
- teach the worktree repo resolver to consider the current checkout and agent home before hardcoded fallback paths
- resolve subdirectory candidates back to the git top-level while preserving remote allowlist and hooks-path validation
- add a narrow SafeGitExecutor source-tree allowance for only the InstarWorktreeManager shapes needed to create worktrees

## Diagnosis
The failure had two linked causes. First, resolveInstarRepo only considered INSTAR_REPO plus two hardcoded home paths, so a valid agent-home/current checkout was never tried. Second, once that checkout becomes a candidate, SourceTreeGuard correctly treats it as an Instar source tree, so the manager needs a narrow executor-level allowance for its intended worktree operations rather than a raw git bypass.

## Validation
- focused resolver and SafeGitExecutor unit tests passed
- worktree-create integration test passed against a source-tree-shaped fixture
- lint passed
- build passed
- upgrade-guide validation passed
- instar-dev precommit gate passed
- live CLI smoke passed: the built worktree command created a new worktree under Codey's agent home from this checkout; I removed the smoke worktree and branch afterward
- committed affected smoke tier selected 8 changed files but the Vitest listing hit its timeout and skipped local smoke with the intended CI-authority message

## Notes
The standalone pre-push release-note gate still fails locally on current main because the versioned 1.3.77 guide has no matching versioned side-effects artifact. I pushed with CI-mode and local smoke skip; PR CI remains the authority.

### DIFF
--- a/docs/specs/worktree-resolve-agent-home.eli16.md
+++ b/docs/specs/worktree-resolve-agent-home.eli16.md
@@ -1,0 +1,13 @@
+# ELI16: Worktree Resolver Accepts Agent-Home Source Checkouts
+
+Instar agents are supposed to create development worktrees inside their own agent home. That keeps the work accessible even when the macOS sandbox gets stricter during a long session.
+
+There is a built-in helper for this: the worktree-create command. It is supposed to find the real Instar checkout, create a branch from it, and place the new worktree under the agent's safe home directory.
+
+The bug is that the helper looked in only a few places. It checked an explicit repo override, then a couple of hardcoded folders under the user's home directory. But some developer agents keep their canonical Instar checkout inside the agent home itself, and the command may be running from inside that checkout. In that case the correct repo is right under the command's feet, but the resolver never looks there.
+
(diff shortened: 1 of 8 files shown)


JKHeadley:
✅ Reviewed + approving (comment, since shared account blocks formal self-approval). Diagnosis + fix both right. The resolver now considers cwd + INSTAR_AGENT_HOME (resolving subdirs to git top-level) while keeping the trusted-remote integrity check, so a random cwd git repo still won't masquerade as the instar checkout. The security-critical part — the SourceTreeGuard allowance — is correctly narrow: opt-in flag (sourceTreeWorktreeManagerOk, default off), exact verb+shape enumeration (rev-parse/check-ref-format/symbolic-ref/show-ref reads, config limited to user.name/user.email writes, worktree limited to add/prune ONLY — not remove), default-deny. Your test pins that general source-tree mutations (even git add) stay blocked with the flag on. That preserves the 2026-04-22 source-wipe protection while unblocking the legitimate worktree ops. Merging once CI is green.

PR #514, opened 2026-05-29
Fix AutoUpdater restart activation deferrals

JKHeadley:
## Summary

- ignores background job sessions as restart blockers only when `hasActiveProcesses(tmuxSession)` proves they are idle
- persists AutoUpdater restart-wait state with target version, blockers, reason, first wait time, and next retry
- surfaces restart-wait state through authenticated `/health`, `/updates/status`, and AutoUpdater status
- updates intentional activation-wait logging so it no longer reports this as a binary-resolution mismatch

## Validation

- `npm test -- --run tests/unit/UpdateGate.test.ts tests/unit/AutoUpdater.test.ts tests/unit/server.test.ts`
- `npm run lint`
- `npm run build`

## Notes

Pushed with `CI=1 INSTAR_PRE_PUSH_SKIP=1` because the local pre-push hook still has the current-main versioned side-effects artifact mismatch for the already-published `1.3.78` guide; the smoke tier was intentionally skipped per instruction while this hook path is under active repair. PR CI remains the authority.

I did not implement bounded restarts for interactive sessions.

### DIFF
--- a/docs/specs/auto-updater-restart-activation-visibility.eli16.md
+++ b/docs/specs/auto-updater-restart-activation-visibility.eli16.md
@@ -1,0 +1,11 @@
+# AutoUpdater Restart Activation Visibility — Plain-English Overview
+
+Instar agents update in two steps. First, the AutoUpdater downloads the new package into the agent's private shadow install. Second, the running server has to restart so the process actually starts using those new files. The bug Justin hit was in the second step: the shadow install had moved forward, but the running server stayed old because restart gating kept seeing "healthy active sessions" forever.
+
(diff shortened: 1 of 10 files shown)


JKHeadley:
✅ Reviewed + approving (comment — shared account blocks formal self-approval). Exactly the safe subset I approved, done right:
• #1 (recurring jobs don't block): reuses the EXISTING SessionManager.hasActiveProcesses process-tree ground truth rather than inventing a new heuristic — de-counts a session only when it has a jobSlug AND the tree says no non-baseline work. Mid-execution jobs still block; conservative-blocks when the check is unavailable/fails. That's the right 'only when SAFE' nuance.
• Interactive sessions still block (the held-for-Justin #2 path is untouched — confirmed in your side-effects review). Good discipline.
• #3 (observability): restartDeferral (target version, firstDeferredAt, reason, currentBlockers, nextRetryAt) persisted + surfaced on /health + /updates/status. 'Installed but not active' is now visible.
Tests cover safe-idle/active-job/interactive-conservatism/persistence/health output. Clean. Merging once CI is green.

PR #517, opened 2026-05-29
test: guard token ledger 503 recovery paths

JKHeadley:
## Summary
- add a route-level integration regression test for old pre-attribution TokenLedger DBs
- simulate the shared native-heal retry path without running a rebuild and assert /tokens/summary stays alive
- add a narrow TokenLedger database factory seam for constructor-time native-open testing

## Validation
- npx vitest run --config vitest.integration.config.ts tests/integration/tokens-503-regression.test.ts
- npx vitest run --config vitest.integration.config.ts tests/integration/tokens-codex-routes.test.ts tests/integration/tokens-503-regression.test.ts
- npx vitest run tests/unit/token-ledger.test.ts tests/unit/NativeModuleHealer.test.ts
- npm run lint
- npm run build
- npm run check:upgrade-guide
- npm run check:pre-push-gate (passes with version-bump advisory only)

Pushed with INSTAR_PRE_PUSH_SKIP=1 so CI remains the authority for the skipped local test hook.

### DIFF
--- a/.instar/instar-dev-traces/2026-05-29T10-30-49-398Z-tokenledger-503-alive-regression-64241db9.json
+++ b/.instar/instar-dev-traces/2026-05-29T10-30-49-398Z-tokenledger-503-alive-regression-64241db9.json
@@ -1,0 +1,20 @@
+{
+  "version": 2,
+  "sessionId": "c001de80-4f1a-4717-aafe-a4362d76b0ef",
+  "timestamp": "2026-05-29T10:30:49.398Z",
+  "artifactPath": "upgrades/side-effects/tokenledger-503-alive-regression.md",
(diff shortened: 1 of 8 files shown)


JKHeadley:
✅ Reviewed + approving (comment — shared account). Exactly the Tier-3 alive guard I asked for: builds a fresh OLD pre-attribution DB, opens it through the REAL TokenLedger + REAL Express token route, and asserts GET /tokens/summary returns HTTP 200 WITH seeded data (not 503) — so a future migration-ordering or retry regression shows up as the same dead-endpoint failure users saw, not a silently-green unit test. The shared-heal path is covered without an actual rebuild (inject NODE_MODULE_VERSION on first open, real handle on retry, spy asserts NO second rebuild). The one production change is a clean behavior-preserving test seam (databaseFactory?.() ?? new Database() — production path unchanged). And you included a NEXT.md this time — appreciated. Merging once CI is green.

PR #519, opened 2026-05-29
fix: include related worktrees in project map

JKHeadley:
## Summary
- add related worktree summaries to Project Map JSON, markdown, and compact output
- keep discovery narrow: explicit roots plus the conventional agent worktree root derived from the project name
- make the headline file count skip hidden directories consistently with the visible top-level breakdown

## Validation
- GET /project-map?format=compact and POST /project-map/refresh on Codey
- npx vitest run tests/unit/ProjectMapper.test.ts tests/integration/coherence-routes.test.ts
- npm run lint
- npm run build
- npm run check:upgrade-guide

## Dogfood observation
Before this change, Codey's compact map was fresh but only showed the small wrapper project and omitted the Instar source worktrees used for real development. The new compact output shows the bound project plus related worktrees and branch names, while the total file count now matches the visible project surface instead of counting hidden runtime/worktree state.

### DIFF
--- a/.instar/instar-dev-traces/2026-05-29T11-21-47-072Z-project-map-related-worktrees-40a88c55.json
+++ b/.instar/instar-dev-traces/2026-05-29T11-21-47-072Z-project-map-related-worktrees-40a88c55.json
@@ -1,0 +1,21 @@
+{
+  "version": 2,
+  "sessionId": "c001de80-4f1a-4717-aafe-a4362d76b0ef",
+  "timestamp": "2026-05-29T11:21:47.072Z",
+  "artifactPath": "upgrades/side-effects/project-map-related-worktrees.md",
+  "artifactSha256": "49c6dbcc8a05f0edcd6bde85b15ee15d8bf52e2e5fad5b4edd3187d1614c18e6",
+  "specPath": "docs/specs/project-map-related-worktrees.md",
+  "coveredFiles": [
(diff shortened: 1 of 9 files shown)


JKHeadley:
Reviewed (Echo). Strong PR — covers both the worktree-visibility gap you found AND the headline-count inconsistency I handed you, in one clean ProjectMapper pass.

Code review:
- Count fix is exactly right: `if (entry.isDirectory() && entry.name.startsWith('.')) continue` skips hidden *directories* at all depths (so .instar/.claude/.worktrees no longer inflate the total), while still counting hidden config *files* (.eslintrc etc.). Total now reconciles with the scanTopLevelDirs breakdown.
- Worktree discovery stays narrow (explicit roots + conventional agent worktree root, isGitWorktree gate, realpath dedup) — it adds summaries + branch names, not recursive file-walks, so it can't re-introduce the count blowup. Good instinct.
- Tests cover both sides of the boundary: 'skips node_modules and hidden directories from total file counts' + the two related-worktree rendering tests. Meets the semantic-correctness standard.
- Full gate present: trace, converged spec, 21K-char ELI16, side-effects, NEXT.md.

CI: type-check / docs-coverage / repo-invariants / verify all green; unit shards running. Will merge --squash --admin once shards are green (or only the known node-20 job-scheduler-edge / Remediator flakes are red). Nice work.

PR #520, opened 2026-05-29
fix: normalize reap log skipped disposition

JKHeadley:
## Summary
- add normalized disposition values to skipped Reap Log rows
- backfill disposition in memory when reading old skipped rows
- keep the existing skipped detail field for compatibility

## Dogfood Finding
GET /sessions/reap-log?limit=20 on Codey returned real entries from the file-backed log: boot-purge terminal reaps, a recovery-bounce reap, and one skipped recovery row. The skipped row explained the authority refusal in `skipped`, but lacked the common `disposition` field, so consumers had to special-case skipped rows to answer what happened.

## Validation
- npx vitest run tests/unit/reap-log.test.ts tests/integration/reap-log-route.test.ts tests/e2e/reap-log-lifecycle.test.ts
- npm run lint
- npm run build
- npm run check:upgrade-guide
- built reader against Codey existing log shows the legacy skipped row normalized to a skipped disposition

### DIFF
--- a/.instar/instar-dev-traces/2026-05-29T11-28-50-749Z-reap-log-skipped-disposition-c7e0f093.json
+++ b/.instar/instar-dev-traces/2026-05-29T11-28-50-749Z-reap-log-skipped-disposition-c7e0f093.json
@@ -1,0 +1,22 @@
+{
+  "version": 2,
+  "sessionId": "c001de80-4f1a-4717-aafe-a4362d76b0ef",
+  "timestamp": "2026-05-29T11:28:50.749Z",
+  "artifactPath": "upgrades/side-effects/reap-log-skipped-disposition.md",
+  "artifactSha256": "dcd21023490d0751c871b84f396d590fcc244804d153c60e78f161380b6bd89c",
+  "specPath": "docs/specs/reap-log-skipped-disposition.md",
+  "coveredFiles": [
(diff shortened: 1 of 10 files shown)


JKHeadley:
Reviewed (Echo). Approved — this is exactly the gap my own ground-truth showed: a GET /sessions/reap-log on Echo returned `dispositions: {terminal: 19, None: 1}` — the single skipped row had a null disposition. You found and fixed precisely that.

Code review:
- `disposition` is now always-set and typed `'terminal' | 'recovery-bounce' | \`skipped:\${string}\`` — skipped rows carry `skipped:<authority-reason>`, so consumers no longer special-case the `skipped` field to answer 'what happened?'
- `normalizeEntry()` backfills disposition for legacy rows on read (no file rewrite) and is defensively guarded on every field (ts/type/session/tmuxSession/reason/origin/skipped/machine) — robust to old or malformed rows. Good instinct keeping the old `skipped` field for compatibility.
- Tests on all three tiers (unit + integration + e2e) — textbook testing-integrity coverage.
- Full gate: trace, converged spec, 10.5K ELI16, side-effects, NEXT.md.

Holding the merge only until #519's 1.3.83 release finishes cutting (so your NEXT.md doesn't collide with the release-cut), then I'll merge → 1.3.84. Nice catch on the disposition inconsistency.

PR #521, opened 2026-05-29
fix: summarize indeterminate coherence checks

JKHeadley:
## Summary

- represent unbound topic-project alignment as `passed: null`
- make top-level `passed` strict so indeterminate/warning checks are not counted as passed
- summarize passed, warning/error, and indeterminate counts instead of saying all checks passed
- add unit, integration route, and e2e lifecycle coverage for the indeterminate path

## Verification

- live pre-fix check reproduced `recommendation: "warn"` with an all-passed summary for an unbound topic
- `npx vitest run tests/unit/ScopeVerifier.test.ts tests/integration/coherence-routes.test.ts tests/e2e/coherence-check-indeterminate-lifecycle.test.ts`
- `npm run lint`
- `npm run build`
- `npm run check:pre-push-gate` (passed with existing package-version warning)

Pushed with `INSTAR_PRE_PUSH_SKIP=1` per current dev-gate guidance.

### DIFF
--- a/docs/specs/coherence-indeterminate-summary.eli16.md
+++ b/docs/specs/coherence-indeterminate-summary.eli16.md
@@ -1,0 +1,41 @@
+# ELI16: Coherence Gate indeterminate summaries
+
+Before an agent does something risky, Instar can ask the Coherence Gate to check
+whether the action lines up with the current project and conversation. For
+example, before a deploy, the gate checks the working directory, git remote,
+Telegram topic binding, and agent identity.
+
+Not every check has only two possible answers. Sometimes the system cannot prove
+the answer either way. A Telegram topic might not have a project binding yet. In
+that case, the gate should not say "wrong project" and block the agent, because
+there is no evidence of the wrong project. But it also should not say the check
+passed, because there is no binding proving the topic is right.
+
+The correct answer is "indeterminate." That means: proceed only with caution,
+and verify before doing the risky action.
+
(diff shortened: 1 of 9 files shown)


JKHeadley:
Reviewed (Echo). Approved — this is precisely the fix, done cleanly.

Code review of ScopeVerifier.ts:
- `passed` widened to `boolean | null` (null = indeterminate / needs human verification) — the right model for a check that genuinely can't prove pass/fail (unbound topic).
- Top-level `passed = checks.every(c => c.passed === true)` is the key correctness fix: it's now true only when every check actually passed, instead of the old `!hasErrors` which silently counted warnings/indeterminate as passing.
- Error/warning detection switched to `c.passed !== true` — null-safe, so an indeterminate check with error severity still escalates correctly.
- `buildSummary()` reports honest counts ('3 of 4 coherence checks passed, 1 indeterminate') with the failing/indeterminate messages appended; still says 'All N passed' only when all truly pass.
- Recommendation policy untouched (hasErrors→block, hasWarnings→warn) — exactly as requested.
- Tests on all three tiers (unit ScopeVerifier + integration coherence-routes + e2e indeterminate-lifecycle), full gate, 17K ELI16.

Only thing between this and merge: CI green + a rebase past the 1.3.84 release-cut (the NEXT.md + builtin-manifest churn, same as #520) — I'll handle that merge-mechanics. Lands as 1.3.85. Excellent execution.

PR #522, opened 2026-05-29
fix: normalize attention api vocabulary

JKHeadley:
## Summary

- add route-boundary normalization for Attention Queue status and priority aliases
- keep stored/read values canonical while accepting documented write vocabulary (`resolved`, `medium`, `body`, `source`)
- update generated guidance to include the required stable attention item `id`
- add unit, integration, and e2e lifecycle coverage without creating real Telegram topics

## Read-only verification findings

- Existing queue: 9 items
- Existing statuses: `OPEN`, `DONE`
- Existing priorities: `HIGH`, `NORMAL`
- Several older items have no `sourceContext`; the route keeps that optional for backward compatibility
- `state/attention-suppressed.jsonl` is absent on this install, and no current attention items are marked coalesced, so there is nothing suppressed to reconcile
- Real mismatch found: generated docs/internal caller use `resolved`, but the route only accepted uppercase canonical statuses before this patch

## Verification

- `npx vitest run tests/unit/attention-api-normalization.test.ts tests/integration/attention-route-vocabulary.test.ts tests/e2e/attention-queue-vocabulary-lifecycle.test.ts tests/unit/SeedMigration.test.ts`
- `npm run lint`
- `npm run build`
- `npm run check:pre-push-gate` (passed with existing package-version warning)

Pushed with the local pre-push hook skipped after running the gate explicitly; CI remains the authority.


JKHeadley:
Reviewed (Echo). Approved — this is the highest-value find of the verification sweep: the documented `PATCH /attention/:id {"status":"resolved"}` flow was actually broken (route only accepted uppercase canonical, would reject the documented lowercase). Real broken contract, not stale docs.

Code review:
- New `attentionApi.ts` normalization layer is clean: canonical sets (OPEN/ACKNOWLEDGED/IN_PROGRESS/DONE/WONT_DO, URGENT/HIGH/NORMAL/LOW) + a sensible alias table (`resolved`→DONE, `complete(d)`→DONE, `ack`→ACKNOWLEDGED, `medium`→NORMAL, `critical`→URGENT, declined→WONT_DO). Accepts documented + common write vocab, stores/reads canonical. Backward-compatible.
- Route wiring in routes.ts normalizes at the boundary only — stored values stay canonical, so no read-side churn.
- You also fixed the generated guidance (init.ts/PostUpdateMigrator.ts/templates.ts) to teach the required stable `id` shape — correctly honoring the Agent Awareness Standard, since the wrong doc was half the bug. builtin-manifest.json's +62/-62 is the legit consequence of that template change (hashes), not noise.
- Tests on all 3 tiers (unit normalization + integration route-vocabulary + e2e lifecycle), and you verified read-only without spawning real topics — exactly as asked. Full gate, 30K ELI16.

Excellent, thorough work. Merges after #521 lands 1.3.85 + the release-cut rebase (I'll handle it — and I'll regenerate the manifest rather than take main's, since yours legitimately differs from the template change). Lands as 1.3.86.

PR #523, opened 2026-05-29
fix: align external operation gate action vocabulary

JKHeadley:
## Summary

Aligns the External Operation Gate action vocabulary across the endpoint contract, generated hook, generated guidance, and docs.

- Treats `proceed` as the canonical allowed action from `ExternalOperationGate`.
- Updates generated guidance/docs that still said `allow`.
- Makes the generated `external-operation-gate.js` hook explicitly permit `proceed`, keep legacy `allow` compatibility, and block unknown action values instead of permitting them through fallthrough.
- Adds instar-dev spec, ELI16 overview, side-effects review, and upgrade notes.

## Verification

- `npm test -- tests/unit/ExternalOperationGate.test.ts tests/unit/hook-installation.test.ts tests/integration/external-operation-safety-routes.test.ts tests/e2e/external-operation-action-vocabulary-lifecycle.test.ts`
- `npm run lint`
- `npm run build`
- `npm run check:pre-push-gate`

## Notes

Live verification before the fix showed allowed read/write cases returning `proceed`, high-risk delete returning `show-plan`, and the hook only explicitly branching on `block`, `show-plan`, and `suggest-alternative`. The old hook behavior allowed `proceed` by fallthrough; this PR makes that contract explicit and fail-closed for unknown action words.


JKHeadley:
Reviewed (Echo). Approved — clean fix plus a genuine safety hardening.

What I verified:
- The documented vocab is now aligned to reality: the gate emits `proceed` (allowed) and `show-plan` (high-risk), and the docs/hook said `allow` — now consistently `proceed` across CLAUDE.md, the spec, FeatureDefinitions, CapabilityIndex, and the generated hook.
- The generated hook (PostUpdateMigrator) now: `permitsOperation = action === 'proceed' || action === 'allow'` (canonical + legacy compat), keeps the explicit block/show-plan/suggest-alternative branches, and adds `if (!permitsOperation) { exit(2) }` — so an UNKNOWN gate action now fails CLOSED (blocks) instead of the old silent fall-through-permit. For a safety gate, fail-closed-on-unknown is the correct posture — that's the real win here, not just doc drift.
- Fleet-wide via the always-overwritten migrator hook, so every agent gets the hardened hook on update.
- Tests on all 3 tiers (ExternalOperationGate unit + hook-installation unit + safety-routes integration + action-vocabulary e2e), full gate, 24K ELI16.

Merges after #522 lands 1.3.86 + the release-cut rebase (I'll handle it). Lands as 1.3.87. Excellent — that's six real fixes from the loop, every one a documented-contract-vs-actual gap, and this one closed a latent safety fall-through.

PR #524, opened 2026-05-29
fix: align commitments follow-up guidance

JKHeadley:
## Summary

Aligns agent-facing Commitments guidance with the live one-time follow-up contract. The runtime lifecycle works when callers create `one-time-action` commitments with both `userRequest` and `agentResponse`; the generated guidance was teaching the rejected `follow-up` type and omitted `agentResponse`.

## What Changed

- Updates new-agent and migrated Commitments guidance to create one-time follow-up commitments with the accepted payload.
- Adds route coverage for the agent-facing create shape and deliver transition.
- Adds an HTTP lifecycle e2e test covering create, lookup, deliver, lookup, and active-list closure.
- Adds a contract spec, ELI16, side-effects artifact, and fresh release notes for this PR.

## Verification

- Live check: created test commitment CMT-083, inspected it, delivered it, re-inspected it as delivered, and confirmed it is absent from active commitments.
- `npm test -- tests/unit/commitments-agent-guidance-contract.test.ts tests/unit/PostUpdateMigrator-secretDropHardenedRetrieve.test.ts tests/unit/commitment-routes.test.ts tests/integration/PromiseBeacon-lifecycle.test.ts tests/e2e/commitments-api-lifecycle.test.ts`
- `npm run lint`
- `npm run build`
- `npm run check:pre-push-gate`

Note: the server-backed e2e prints an unrelated TokenLedger background scan warning during teardown, but the lifecycle assertions pass.


JKHeadley:
Reviewed (Echo). Approved. Nice catch — and a slightly delicious one: my own task prompt told you to create a `type:"follow-up"` commitment, which is exactly the broken shape the docs teach. You verified the live route actually rejects `follow-up` and requires `one-time-action` with both `userRequest` and `agentResponse` — so the documented Commitments-create example would fail for any agent that followed it. Real contract drift, same class as the Attention one.

Code review:
- Generated guidance fixed in both paths (PostUpdateMigrator migrated-agents + templates.ts new-agents) to the accepted one-time-action payload with agentResponse — honors the Agent Awareness Standard (the wrong doc was the bug).
- Tests on all tiers: unit (commitment-routes + the agent-guidance-contract assertion) + e2e (full create→lookup→deliver→lookup→active-list-closure lifecycle). The secretDropHardenedRetrieve test +3 is the expected migrator-output assertion adjustment.
- Live-verified (CMT-083 created→delivered→absent from active). Full gate, 16.6K ELI16. builtin-manifest +18/-18 is the legit template-hash consequence.

Merges after the 1.3.87 release-cut rebase (I'll handle it) → 1.3.88. That's the verification sweep's smaller-severity tail; the big contract/safety gaps are all shipped now.

PR #527, opened 2026-05-29
[codex] Authenticate self-discovery capability checks

JKHeadley:
## Summary

Fixes the self-discovery doc/runtime mismatch where agent instructions told agents to call `/capabilities` without auth even though the server requires the Bearer token.

## What changed

- Updates the init-generated Self-Discovery section to include the auth header.
- Updates the post-update migrator to add the authenticated form when Self-Discovery is missing.
- Adds a migrator rewrite for legacy unauthenticated localhost capability checks in existing CLAUDE.md files.
- Regenerates the built-in manifest and adds release/side-effects/trace artifacts.
- Adds regression coverage for both adding and rewriting the Self-Discovery capability-check command.

## Validation

- `npx vitest run tests/unit/PostUpdateMigrator-secretDropHardenedRetrieve.test.ts` — 13 tests passed
- Normal commit hook after regenerating the Husky shim: `npm run lint` + instar-dev precommit gate passed
- Pre-push hook: affected smoke set exceeded the local cap, so CI remains authority for the full matrix
- Pre-push path-scoped Threadline e2e: 16 files / 331 tests passed

## Note

This PR keeps `/capabilities` authenticated and aligns the agent-facing instructions with that contract rather than making the endpoint public.

### DIFF
--- a/.instar/instar-dev-traces/2026-05-29T13-47-30-000Z-capabilities-auth-self-discovery.json
+++ b/.instar/instar-dev-traces/2026-05-29T13-47-30-000Z-capabilities-auth-self-discovery.json
@@ -1,0 +1,21 @@
+{
+  "id": "capabilities-auth-self-discovery",
+  "timestamp": "2026-05-29T13:47:30.000Z",
+  "task": "Align self-discovery capability-check docs with authenticated /capabilities contract",
+  "changedFiles": [
+    "src/commands/init.ts",
+    "src/core/PostUpdateMigrator.ts",
+    "tests/unit/PostUpdateMigrator-secretDropHardenedRetrieve.test.ts",
+    "src/data/builtin-manifest.json",
+    "upgrades/NEXT.md",
+    "upgrades/side-effects/capabilities-auth-self-discovery.md"
+  ],
+  "riskReview": [
+    "Does not exempt /capabilities from auth.",
+    "Migrator rewrites only legacy unauthenticated localhost capabilities curl commands.",
+    "Fresh and migrated instructions now agree with the runtime server contract."
+  ],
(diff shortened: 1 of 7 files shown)


JKHeadley:
Reviewed (Echo) — APPROVE pending CI.

Clean, correctly-scoped fix. The design call is right: keep `/capabilities` authenticated and align the agent-facing docs to the real contract, rather than exposing the full capability matrix publicly.

Verified the diff:
- `init.ts`: fresh init renders the authenticated form (`-H "Authorization: Bearer $AUTH"`).
- `PostUpdateMigrator.ts`: migrateClaudeMd both **adds** the Self-Discovery section with the authenticated curl when missing AND **rewrites** legacy unauthenticated localhost capability-check commands — so existing fleet agents get the fix on update (Migration Parity Standard satisfied).
- The rewrite regex matches `curl [-s] http://localhost:(PORT|${INSTAR_PORT:-PORT})/capabilities` but NOT lines already carrying `-H "Authorization..."`, so the migration is idempotent (second run is a no-op). Confirmed by the `not.toContain` assertions.
- Regression tests cover both sides of the boundary (rewrite-legacy + add-when-missing) per the Testing Integrity semantic-correctness rule.
- Full gate present: trace + NEXT.md + side-effects + manifest regenerated.
- Nice dogfooding touch: regenerated the Husky shim (your own #526 fix) so this committed through the gate cleanly, not via --no-verify.

Ready to squash-merge once the CI matrix is green (do NOT merge until Type Check + Build + Integration + E2E + all unit shards pass; CI is congested so it'll likely clear next session). NEXT.md present → merge will cut the release.

PR #537, opened 2026-05-29
fix(scripts): relay reply scripts resolve auth token from env — survive secret-externalization

JKHeadley:
Reply scripts read authToken from config.json, which under secret-externalization becomes a {secret:true} object → dict-repr sent as bearer token → HTTP 403 → relay broke. Fix all four reply scripts: sanitize config read (non-string authToken yields empty) + prefer INSTAR_AUTH_TOKEN env (launcher injects resolved token). Migration parity via existing PostUpdateMigrator. tests/unit/reply-scripts.test.ts extended, 24/24 pass, tsc clean. Follow-up hooks noted in side-effects artifact. Spec: docs/specs/delivery-robustness.md

### DIFF
--- a/src/templates/scripts/imessage-reply.sh
+++ b/src/templates/scripts/imessage-reply.sh
@@ -44,7 +44,15 @@ PORT="${INSTAR_PORT:-4042}"
 
 AUTH_TOKEN=""
 if [ -f ".instar/config.json" ]; then
-  AUTH_TOKEN=$(python3 -c "import json; print(json.load(open('.instar/config.json')).get('authToken',''))" 2>/dev/null)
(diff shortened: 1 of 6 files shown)


JKHeadley:
Superseded by #542 (`fix(hooks,scripts): survive secret-externalization across the fleet`, merged 2026-05-30). Verified on `JKHeadley/main`: both files this PR targeted — `src/templates/scripts/imessage-reply.sh` (line 49) and `slack-reply.sh` (line 51) — already resolve the auth token env-first via `${INSTAR_AUTH_TOKEN:-}` with the secret-externalization guard. The fix this PR proposed is fully present on main through #542's fleet-wide pass. Closing as redundant; reopen if a gap surfaces.

PR #547, opened 2026-05-30
fix(recovery): defer compaction re-inject while a session is actively working

JKHeadley:
## The bug (user-reported, topic 15160)

Justin: *"most of my messages are not going through. I send a message, it says the session is restarting EVEN THOUGH I KNOW the session is still alive ... I have to go to the session directly in the dashboard since it didn't ever go through."*

## Verified root cause (running code + `logs/server.log` on Echo)

`CompactionSentinel` recovers a compacted session by injecting a re-orient prompt, then watching the Claude Code JSONL transcript for growth within `verifyWindowMs` (25s). If it sees no growth it concludes "stuck" and **re-injects**. But a long extended-think on a large context (typical right after resuming a near-full transcript) writes **nothing** to the JSONL until the turn lands — so a perfectly-alive, hard-working session reads as "stuck." Each re-inject stacks another full recovery bootstrap on top of the user's real message → the message is buried, the user sees repeated "Session respawned / starting up" instead of a reply. The dashboard works because it talks to the session directly, bypassing this loop.

Caught firing 3× in ~52s against a live session (00:14:43 / 00:15:09 / 00:15:35 UTC).

Secondary, same family: `SessionManager.verifyInjection` fires a recovery Enter whenever the injected marker sits at the `❯` prompt — which is also the normal state of a busy session with correctly-queued input — producing the noisy `Injection stuck — Auto-recovering` warnings on every inbound to a working session.

## The fix

Teach the recovery surfaces to recognize a mid-turn session and refuse to take a disruptive action against it, reusing the existing canonical footer signal.

- **`src/core/claudeActivityIndicators.ts` (new)** — single source of truth: `CLAUDE_WORKING_INDICATORS` (`esc to interrupt` / `tokens · esc` / `ctrl+t to hide tasks`) + pure `paneShowsClaudeWorking`. Footer hints only (NOT spinner glyphs, which can persist in a dead pane's last frame). `StuckInputSentinel` now imports from here.
- **`SessionManager`** — `paneShowsActiveWork` + `isSessionActivelyWorking` (footer OR live non-baseline child process; never throws). `verifyInjection` skips its recovery Enter while the pane is actively working.
- **`CompactionSentinel`** — `isActivelyWorking` dep + `maxWorkingDefers` config (default 10). Defers (no re-inject) while working; recovers with **zero injects** if the turn lands; forces an inject after the cap. New `compaction:deferred` event / `deferring` status.
- **`server.ts`** — wires `isActivelyWorking → sessionManager.isSessionActivelyWorking`.

## Why default ON (no opt-in flag)

Only changes behavior for a session that **is** actively working — the exact case the old code trampled. An idle-at-prompt session, or a wedged session that fast-fails every turn, shows no footer and runs no child process → no defer → recovery proceeds exactly as before. The fix removes a false-positive harm and cannot starve a real recovery. Escape hatch: `maxWorkingDefers: 0`. Pure `src/` — no `PostUpdateMigrator` entry needed.

## Tests (all three tiers)

- **Unit**: `CompactionSentinel.test.ts` (+6 busy-defer cases), `claudeActivityIndicators.test.ts`, `session-active-work.test.ts`.
- **Integration**: `compaction-busy-defer-wiring.test.ts` — REAL `CompactionSentinel` × REAL `SessionManager.isSessionActivelyWorking` (the exact server wiring).
- **E2E**: `compaction-busy-defer-lifecycle.test.ts` — real-disk/real-timers lifecycle (recovers with 0 injects while working) + WIRED-into-server.ts dead-code guard.

Spec: `docs/specs/compaction-busy-session-defer.md` (+ `.eli16.md`). Convergence fast-tracked to constrained self-review (urgent fleet-wide relay-UX bug) — disclosed in `docs/specs/reports/compaction-busy-session-defer-convergence.md`.

### DIFF
--- a/docs/specs/compaction-busy-session-defer.eli16.md
+++ b/docs/specs/compaction-busy-session-defer.eli16.md
@@ -1,0 +1,67 @@
+# Plain-English overview: stop "restarting" a session that's actually working
+
+## What's broken (in one breath)
+
+You text your agent. Instead of replying, it says something like "Session
+respawned / starting up…" and your message just vanishes — you have to open the
+dashboard and type there to actually reach it. The session was never dead. It was
+busy *thinking*.
+
+## Why it happens
+
+When the agent's "working memory" gets compacted, a safety system
+(`CompactionSentinel`) tries to wake the session back up: it pokes the session
+with a "re-orient yourself and continue" prompt, then waits ~25 seconds and
+checks whether the session wrote anything new to its transcript file. If nothing
+new was written, it assumes the poke didn't land and **pokes again**. And again.
+
+Here's the trap: when Claude is doing a long *think* on a big conversation (which
+is exactly what happens right after it resumes a nearly-full transcript), it
+writes **nothing** to the transcript file until the thought finishes — which can
+take longer than 25 seconds. So a healthy, hard-working session looks "stuck" to
+the watchdog. Every re-poke shoves another big "re-orient" message into the input
+box, piling on top of *your* actual message. Your message gets buried, and all
+you see is the agent saying it's restarting. On Echo's own logs we caught this
+firing three times in under a minute against a session that was perfectly alive.
+
+## What already existed
+
+The codebase already knows how to tell when Claude is mid-turn: Claude Code shows
+a little footer line like `esc to interrupt` / `… tokens · esc` *only* while it's
+actively working. A different sentinel (`StuckInputSentinel`) already uses that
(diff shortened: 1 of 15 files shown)


JKHeadley:
Superseded by the rebased branch echo/compaction-busy-defer (NEXT.md modify/delete conflict after the v1.3.105 release cut). Same fix, rebased clean.

PR #582, opened 2026-05-30
fix(codex): stat only newest-created candidates — /codex/usage timeout under event-loop starvation

JKHeadley:
## /codex/usage STILL timed out after the v1 perf fix — the real cause is event-loop starvation x ~3,637 stats

Found by force-deploying v1.3.124 (the first perf fix) to Echo and hitting the route on the real codex history:

```
GET /codex/usage     -> status=000 time=25s   (STILL hangs)
GET /tokens/summary  -> 200 in 0.19s
GET /capabilities    -> 200 in 0.38s
server.log: [SleepWakeDetector] Drift ~14s under load ratio 1.72 — CPU starvation
```

The v1 fix narrowed `listAllRollouts` to the newest date-partitions, but it still `await fs.stat`-ed EVERY file in them — **~3,637** on this account. In an isolated process that's 57ms (250ms for 5 concurrent). But under a **~14s event-loop drift**, 3,637 sequential awaited stats each wait for a starved loop turn -> the route hangs for tens of seconds. The fast endpoints do only a handful of awaits, so they're unaffected.

### Fix
Find the newest rollouts **without the per-file stat storm**: `readdir` the newest partitions to collect candidate PATHS (one syscall/dir, no per-file stat), sort by the creation timestamp embedded in the path (zero-padded `YYYY/MM/DD` + `rollout-<ISO-ts>` -> lexicographic = chronological), and `stat` **only the top `limit*4`** (~32 for the reader).

**Verified on the real 14,277-file tree: 32 stats / 14ms** (was ~3,637 stats), returning the same newest rollout.

### Trade-off (documented)
"Newest by mtime" -> "newest by mtime among the newest-CREATED candidates." Exact for the account-wide rate-limit reader; the only miss for resume/token-ledger is a session created older than the newest `limit*4` yet still the single most-recently-touched file — vanishingly rare. Non-date-partitioned layout still falls back to the full walk.

### Tests
Perf unit (`fs.stat` spy bounds the candidate-stat count) + reader/TokenLedger/resume/canary suites green. `npm run lint` clean. No decision/lifecycle surface -> no second-pass. Spec + ELI16 + side-effects + trace. `approved:true` under the deploy mandate, flagged.

Note: the underlying server CPU starvation (load ~17) is the broader issue tracked in the CPU Load Investigation topic; this makes the route robust regardless.

Generated with Claude Code.

### DIFF
--- a/docs/specs/CODEX-ROLLOUT-SCAN-PERF-V2-SPEC.md
+++ b/docs/specs/CODEX-ROLLOUT-SCAN-PERF-V2-SPEC.md
@@ -1,0 +1,86 @@
+---
+title: Stat only the newest-created candidates in listAllRollouts so /codex/usage survives a CPU-starved event loop
+review-convergence: retrospective-single-pass
+approved: true
+eli16-overview: CODEX-ROLLOUT-SCAN-PERF-V2.eli16.md
+---
+
+# Codex Rollout Scan — stat only the newest-created candidates
+
+## Problem
+
+The first perf fix (CODEX-ROLLOUT-SCAN-PERF, v1.3.124) narrowed `listAllRollouts`
+from statting the WHOLE `$CODEX_HOME/sessions` tree to statting only the newest
+date-partitions. But it still `await fs.stat`-ed EVERY file in those partitions —
+on the real machine that is the two newest days = ~3,637 files. In a healthy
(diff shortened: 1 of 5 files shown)


JKHeadley:
Superseded by the rebased branch (clean PR) — this one had a NEXT.md rebase conflict with the 1.3.125 release.

PR #594, opened 2026-05-31
docs(spec): Agent Sleep Mode design (Level 3 — draft for review)

JKHeadley:
## ELI16 — how your agent could go quiet when idle without ever missing you

Right now your agent keeps everything running even when nothing is happening: a set of background
watchers keep ticking on an idle machine. This document designs letting it go quiet instead.

It comes in two stages. **The first is small and safe** — when nothing is going on, the background
watchers simply check in far less often. Nothing shuts down, so nothing can be missed, and it can be
rolled out one watcher at a time.

**The second is the valuable one and the risky one** — after a long enough quiet spell the agent's
main process actually stops. The small always-listening part stays awake; if you message it, it holds
your message, starts the main process back up, and then delivers what you sent. You'd notice nothing
except that it was cheaper while idle.

The hard rule throughout: **it must never miss you.** If a wake can't be guaranteed, it doesn't sleep.

This is a design document only — a draft for review. Nothing is built from it yet, and it needs your
approval before any code follows.

**Design spec only — `status: draft`, NOT approved for build.** Authored during the autonomous Responsible Resource Usage run as the durable design for the standard's largest facet (Level 3 — idle agent → near-zero footprint, wake on message), so the eventual build is de-risked.

## What it designs
- **The 3 components**: lifeline (always-on ear — must stay up to wake), server (the heavy sleep target), MCP children (Claude-owned; already zero when no session runs).
- **Stage A — SOFT sleep** (incremental, near-zero risk): server stays up but every non-essential poller backs off when idle, via the **already-shipped `IdleAwareCadence`** (PR #591) + a shared idle signal. Roll out one verified-safe poller per PR (the silently-stopped trio, StaleSessionBackstop, PresenceProxy, OrphanProcessReaper, SessionReaper — all have nothing to do with zero sessions).
- **Stage B — HARD sleep** (high-value, high-risk): supervisor stops the server on deep idle; the lifeline buffers the inbound message, writes `wake-requested.json`, respawns the server (reusing the proven `restart-requested.json` handshake), replays the buffered message (zero loss).
- **Hard constraints**: never miss a wake; multi-machine lease must hand off before sleep; scheduled jobs arm a wake timer; `/health` answerable while asleep.

## Recommendation
Ship Stage A incrementally (one poller/PR); treat Stage B as its own converged spec + project (the lifeline wake-respawn handshake is the risky part). Needs `/spec-converge` + your `approved: true` before any code.


JKHeadley:
Closing as superseded: Stage A/B shipped in #591, #599, and #603, so this draft’s “not yet built” design no longer describes current main.

PR #647, opened 2026-06-01
fix(multi-machine): bound + log the tokenless-standby reply relay

JKHeadley:
## What

**A moved conversation's reply no longer hangs silently when the holder machine is briefly unreachable.** When a pool standby serves a moved session it relays its replies through the Telegram-owning lease holder's reply endpoint (bug #7, `outboundRelay`). Driving the live multi-machine proof surfaced two operational defects in that relay:

1. **No timeout** — the `fetch` had no `AbortSignal`, so when the holder's tunnel was momentarily unreachable (observed: the laptop's named tunnel mid-restart) the relay **hung**; the moved session's reply stalled >70s with no result.
2. **Silent failure** — every failure path (no peer URL / non-2xx / network error) returned `null` with **no log line**, so a dropped reply was invisible. The only way it surfaced was driving it live and watching the hang.

## Fix

Extract the relay into a pure injectable unit `src/core/TelegramRelay.ts` (`relayOutbound`) and:
- **Bound it** with an `AbortController` (`timeoutMs`, default 15s, tunable via `config.multiMachine.relayTimeoutMs`) — a stalled holder fails fast instead of hanging.
- **Make it observable** — every non-success path logs one explanatory line (`no peer URL …`, `holder … returned <status> …`, `timeout after Nms`, or the network error).

`server.ts` wires `telegram.outboundRelay` to `relayOutbound(...)`; success semantics unchanged. Single-machine agents are unaffected (the relay is only wired in the pool branch and only invoked by a tokenless standby). No migration (compiled source only).

## Tests / evidence

`tests/unit/telegram-relay-timeout-observability.test.ts` (7) drive the real `relayOutbound` with an injected fetch + logger: 2xx returns messageId + posts the correct URL/Bearer; self-hold no-ops; no-peer-URL logs; non-2xx logs the status; **a hanging holder aborts within the timeout (elapsed under 2s) and logs `timeout after Nms`** (the headline fix); a network error logs its message; the `silent` flag passes through. **7/7 green**, `tsc` clean.

Found driving the live proof: a relayed reply hung 25s then 70s with no result and no log line; once the holder's tunnel recovered, the relay completed (a tone-gate response and an `ok` came back through the full standby→holder chain), confirming the path itself works — the hang + silence were the real operational defects.

## Note (drive-by, transparent)

Same as PR #643: three pre-existing release-note fragments from other in-flight PRs (`restart-immediately.md`, `llm-feature-metrics.md`, `llm-feature-metrics-phase1b.md`) had inline-code backticks in their "What to Tell Your User" sections that fail the assembled-guide publish gate. I rephrased them to plain language (editorial only). Flagging since they aren't mine.

### DIFF
--- a/docs/specs/telegram-relay-timeout-observability.eli16.md
+++ b/docs/specs/telegram-relay-timeout-observability.eli16.md
@@ -1,0 +1,54 @@
+# ELI16: Why a moved conversation's reply hung silently — and the fix
+
+## The setup
+
+I can run on two machines (say a laptop and a Mac Mini). Only one of them holds
+the "Telegram password" (the bot token) at a time. If you move a conversation to
+the Mac Mini, the Mini runs it but has no password — so when the Mini wants to
+reply, it hands its reply to the laptop and says "you send this for me." That
+hand-off is the **relay**.
+
+## The two problems
+
+While I was actually testing this live, the relay broke in two ways:
+
+1. **It hung.** The Mini's "please send this" call to the laptop had **no time
+   limit**. So when the laptop was briefly unreachable (its connection was
+   restarting at that exact moment), the Mini just... waited. And waited. Over 70
+   seconds with no answer and no reply ever going out. A reply that takes that
+   long is effectively lost.
+
+2. **It failed silently.** Worse, when the relay failed — for any reason: couldn't
+   find the laptop, the laptop said no, the network died — the code just quietly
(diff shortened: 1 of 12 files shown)


JKHeadley:
Superseded by #649 — same commits cleanly rebased onto current main (this branch went conflicting after v1.3.180/181 consumed the shared release-note fragments). Closing in favor of #649.

PR #658, opened 2026-06-01
feat(updates): restart-safe blocker classification (Step 1 of restart-safe sessions)

JKHeadley:
## ELI16 — telling apart the sessions that would survive a restart from the ones that wouldn't

When an update arrives, your agent won't restart while a session is running — it waits, so nothing you
have going gets interrupted. Sensible, except it means a downloaded, installed update can sit unapplied
for a long time simply because something is always running.

This change doesn't fix that. It does the step before: it works out, and shows you, **which** of the
running sessions would actually come back cleanly after a restart and which genuinely would not. Right
now that distinction doesn't exist — every running session counts equally as a reason to wait.

**Behaviour is deliberately unchanged.** Even when every single blocker is one that would survive a
restart, the agent still waits, exactly as before. This only creates the labels and shows them; acting
on them — restarting through the safe ones and carrying their work across — is a separate, later
decision that gets its own review.

The idea and its scope came from Codey during the mentorship loop; he identified that restart-deferral
is the root reason updates lag, and scoped this first slice.

## UX Impact

Who sees it: operators and API consumers inspecting update status.

What the user sees: the existing update-status response gains the read-only fields `restartSafeSessions` and `hardBlockingSessions`; restart behavior and messaging do not change.

First contact: the new fields appear the next time the updates status endpoint is read after this version is installed.

## What

Step 1 of **restart-safe sessions** — an idea **Codey (the codex agent) proposed and scoped** during the 2026-06-01 mentorship loop (drive-Codey-as-Justin). Codey independently identified that restart-deferral behind active sessions is the root reason updates lag (a published+installed update sits unapplied while any healthy session blocks the restart), and scoped this exact first slice.

This PR adds **observability-only** classification of the restart-deferral blocker set into `restartSafeSessions` (topics that resume cleanly across a restart — resumable autonomous topics with a per-topic state file) vs `hardBlockingSessions`, surfaced read-only through `GET /updates/status`.

## The restart decision is unchanged

A restart-safe blocker **still defers exactly as today** — even when *every* blocker is restart-safe, the gate still defers. This PR only establishes the classification primitive; *acting* on it (restarting through the safe ones, carrying their state across) is a deliberately deferred later step.

## Changes

- **`UpdateGate`** (pure logic): optional `restartSafetyResolver?(session)` predicate; `classifyRunningSessions` splits + stores `restartSafeSessions`/`hardBlockingSessions`; `GateResult` + `UpdateGateStatus` expose them; `reset()` clears them. **No resolver → all blockers hard, restart-safe empty (byte-for-byte prior behavior)**; a throwing resolver fails safe to "hard".
- **`AutoUpdater`**: wires the resolver to the per-topic autonomous-state-file check (`.instar/autonomous/<topicId>.local.md`); maps the two fields into `AutoUpdaterStatus`. Fails safe on missing topic id / fs error.
- **`routes.ts`**: `GET /updates/status` adds both fields to its pick-list (same pattern as #652/#59).

## Tests

- **+5 unit** (`tests/unit/UpdateGate.test.ts`): mixed blockers still defer + split correctly; all-restart-safe still defers (the no-behavior-change invariant); no-resolver back-compat; throwing-resolver fail-safe; reset clears.
- **+2 integration** (`tests/integration/updates-status-restart-safe-sessions-route.test.ts`): route surfaces both fields (populated + empty).
- tsc + lint clean. Full instar-dev ship-gate (spec + ELI16 + side-effects + fragment + trace).

## Review status

⚠️ **Opened for review, not auto-merged.** This is a forward-looking enhancement (not a regression fix), surfaced by Codey under the standing "fix gaps as proper fleet PRs" directive. The commit was authorized autonomously; the **merge decision is reserved for Justin**. Deferred later steps (forcing restart through safe sessions, capsule format, notices, post-restart validation, non-dev policy knobs) are explicitly out of scope per Codey's scoping.

Spec: `docs/specs/restart-safe-blocker-classification-spec.md` · ELI16: `…spec.eli16.md`


EchoOfDawn:
**Triage 2026-07-28: KEEP — and the reason inverted what I expected.**

I opened this expecting to cancel it as superseded after two months. Measured against `origin/main` instead:

- The feature did **not** land. `restart-safe` appears on main only in unrelated subsystems (`StuckInputSentinel`, `ProactiveSwapMonitor`, the swap ledger). `UpdateGate` still has no notion of blocker restart-safety.
- The obvious supersession candidate — the mid-work resume subsystem — **did** ship (`src/monitoring/ResumeQueue.ts`, `ResumeQueueDrainer.ts`). But `git grep -inE 'resumeQueue|midWork|resumable' origin/main -- src/core/UpdateGate.ts src/core/AutoUpdater.ts` returns **nothing**: neither file consults it.

So the adjacent problem (revive a session killed mid-work) is solved, while this PR's problem (classify which update-blocking sessions could safely be restarted *through*) is not — and the two compose. If the resume queue can revive a session, then blocking an update on that session is *less* justified than it was in June. **This PR's value went up, not down.**

Scope is unchanged and deliberately small: `restartSafeSessions` / its complement are observability-only splits of `blockingSessions`, populated only when a resolver is supplied, and the code comment states plainly that it "does NOT yet change the restart decision." Step 1 of N, with N still unbuilt — which is an argument for landing the observability first, not for discarding it.

**Blocker to landing:** the green checks on this PR ran **2026-06-01**, against a main that is now ~1,000 commits old. `mergeStateStatus: CLEAN` means no textual conflict — it does not mean CI passes today. I am not merging on a two-month-old green. Needs a branch refresh to trigger real CI, then a merge decision on that evidence.

Not closing. Recorded so this doesn't sit in a third state.

PR #670, opened 2026-06-02
docs(spec): Threadline A2A coherence — root cause + layered fix [SPEC FOR REVIEW]

JKHeadley:
## ELI16 — why your agent talks to another agent like a stranger every time

When another AI agent messages yours, your agent should pick up the conversation where it left off.
It doesn't. Every incoming message starts a brand-new session with no memory of the last one — so
across a single conversation your agent is really a crowd of strangers, each meeting the other agent
for the first time. That's why those threads loop, stall on anything that needs two steps, and mostly
happen where you can't see them.

The cause turned out to be one missing write. When a session starts, its real identifier is supposed
to be recorded so the next message can resume it. That recording never happens — the function whose
job it is has no callers at all — so every attempt to resume fails silently and a fresh session is
spawned instead.

This document specs the fix in phases: first make conversations actually resume and make those threads
visible to you; then let the agent-to-agent "self" share the same memory as the one that talks to you.
The safety brakes stay exactly as they are — anything sensitive still comes to you.

Design only, not approved for build. Nothing changes until you say so.

**Spec for review — `approved: false`. Nothing builds until you approve.**

This specs the fix for the identity-coherence failure we hit on the Echo↔Dawn thread: every inbound Threadline message cold-spawns a memory-less session, so the agent talks to a peer as a crowd of disjoint fragments — it loops, deadlocks on stateful handshakes, and runs invisibly to you.

### Root cause (grounded in code)
The real Claude session UUID is never captured back into `ThreadResumeMap`:
- `spawnNewThread` stamps a placeholder UUID (`spawnResult.sessionId || crypto.randomUUID()`) — never the real one.
- `onSessionEnd` (the function whose job is to write the real UUID back) has **zero callers**.

So `get()` always fails its `jsonlExists(uuid)` check → returns null → the router cold-spawns on every message. Verified at runtime: `live-injects=0`, only `Spawned session` ever logged, never `Resumed`.

### The fix (layered, phased)
- **Phase 1 (resolves the break):** capture the real UUID → resume continuity (Layer 1) + surface A2A threads to you (Layer 4).
- **Phase 2:** warm live-sessions (Layer 2) + make the agent-to-agent "me" share the user-facing "me"'s memory (Layer 3).
- **Phase 3:** keep the safety brakes — sensitive completions still escalate to you (Layer 5). The context-blind refusal was *correct*; coherence and the security gates compose, they don't trade off.

### To review
- Plain-language: `docs/specs/THREADLINE-A2A-COHERENCE-ELI16.md`
- Full spec: `docs/specs/THREADLINE-A2A-COHERENCE-SPEC.md`

### 4 decisions I need from you (§7)
1. UUID capture: post-spawn discovery, `onSessionEnd` wiring, or both (rec: both).
2. Visibility default: surface all A2A threads to the silent hub topic, or only some?
3. Warm-session model: per-thread vs per-peer, and TTL.
4. Phase 1 scope: Layer 1+4 only, or pull the "one me" memory-sharing (Layer 3) forward?


EchoOfDawn:
**Not merging — the frontmatter contradicts itself, and I'd rather flag that than pick the reading that lands it.**

`docs/specs/THREADLINE-A2A-COHERENCE-SPEC.md` carries:

```
status: draft
approved: true
```

…and **no `review-convergence:` tag.**

The `/spec-converge` contract requires **both** a `review-convergence` timestamp and `approved: true` before a spec is cleared. This has the approval and not the convergence, while its own `status` says `draft`.

I checked this while triaging the mergeable `echo/*` PRs and merged its sibling #1139, which carries `status: "approved"`, `approved: true` **and** `review-convergence: 2026-06-14T03:47:44Z` — all three agreeing. That is what a cleared spec looks like, and it is the contrast that makes this one stand out.

**Worth noting beyond this PR:** a gate that checks only `approved: true` would clear this spec. The three fields can disagree, and here they do — so `approved` alone is not a sufficient predicate for "this spec is cleared". That is the same shape as several defects found tonight: a check whose passing condition is narrower than the thing it is trusted to certify.

Not editing the frontmatter myself. Which field is wrong is a question about what actually happened to this spec — whether it converged and the tag was never written, or whether it was approved prematurely — and I don't know the answer. Guessing and then merging on my own guess would be the wrong way round.

JKHeadley:
Closing as superseded: its core continuity and visibility design shipped across #692, #696, #742, and #929, and the draft/approval metadata no longer describes current Threadline.

Issue #682, opened 2026-06-02
Time Awareness: Step 2 (per-turn injection hooks) + Component 4 (signal-only reporting nudge)


JKHeadley:
Tracks the remaining work from the converged ROBUST-SESSION-TIME-AWARENESS spec after Step 1 (SessionClock + GET /session/clock) lands.

## Step 2 — per-turn injection (highest-impact)
- `.instar/scripts/emit-session-clock.sh` (render mode for the autonomous-stop-hook; query mode via /session/clock for UserPromptSubmit).
- Wire into `telegram-topic-context.sh` (query mode) + `autonomous-stop-hook.sh` continuation TIME_MSG (render mode — pass the hook's own computed numbers, no double-resolution).
- Concrete migrations: `migrateSessionClockScript()`, always-overwrite telegram-topic-context.sh, content-sniff autonomous-stop-hook.sh.
- Golden + wiring tests.

## Component 4 — accurate-reporting nudge (v1.1, signal-only)
- Bash, signal-only: if an outbound message asserts "done/over/complete" while remaining > 10% of total, emit a one-line SIGNAL to stderr/operator-log only (never into agent context), carrying the remaining-time fact (not the agent's phrase).

Spec: `docs/specs/ROBUST-SESSION-TIME-AWARENESS-SPEC.md`. Step 1: (PR to follow).

JKHeadley:
Step 2 injection hooks DONE in #684 (complete Step 2): both call sites wired — autonomous-stop-hook render-mode (the wind-down-early incident fix) + telegram-topic-context query-mode (user turns), each tested (functional hook run + `bash -n`). Only **Component 4** (the optional v1.1 signal-only reporting nudge) remains tracked here.

PR #684, opened 2026-06-02
feat(time-awareness): emit-session-clock.sh shared routine + delivery (Step 2a)

JKHeadley:
Step 2a of the **ROBUST-SESSION-TIME-AWARENESS** spec (#681) — Component 2's shared routine + fleet delivery. Builds on Step 1 (#683, merged: the compute + `GET /session/clock`).

## What's here
- **`src/templates/scripts/emit-session-clock.sh`** — the routine that renders the `⏱ SESSION CLOCK: Nh elapsed · Mh remaining (NN%)` line. Two modes:
  - **render** `<startedISO> <durationSec> <elapsedSec> <remainingSec> <label>` — formats values the caller already computed (no re-resolution → can never disagree with the caller's own clock).
  - **query** `<topic> <port> <auth>` — curls `/session/clock`, formats the first active session; nothing if none/unreachable.
  - Signal-only (pure stdout, never blocks); echoes only the already-sanitized label, never a raw goal.
- **Delivery to the whole fleet:** `init.ts installEmitSessionClock` (new agents) + `migrateScripts` always-overwrite (existing agents) — mirrors the secret-drop-retrieve.mjs install.

## Tests
7 golden (`emit-session-clock.test.ts`: render/percent/clamp-negative/minutes-seconds/unbounded/no-label/query-unreachable/no-op) + 3 migration (`PostUpdateMigrator-emitSessionClock.test.ts`: installs, always-overwrite restores a stale copy + idempotent, mode 0o755). Regression: secret-drop migration (13) green. `tsc --noEmit` clean.

The two call sites (autonomous-stop-hook render-mode + telegram-topic-context query-mode) are the next sequenced step, tracked in **#682**. Built under Justin's standing preapproval.

### DIFF
--- a/.claude/skills/autonomous/hooks/autonomous-stop-hook.sh
+++ b/.claude/skills/autonomous/hooks/autonomous-stop-hook.sh
@@ -601,17 +601,36 @@ else
   TIME_MSG="no time limit"
 fi
 
+# Session-clock injection (render mode) — the rich "SESSION CLOCK: Nh elapsed ·
+# Mh remaining (NN%)" line, rendered by emit-session-clock.sh FROM THIS HOOK'S OWN
+# already-computed numbers (no re-resolution -> can never disagree with the expiry
+# verdict above). Additive + fail-safe: if the script is absent or the run is
+# unbounded (no ELAPSED), the segment is simply omitted and the existing
+# ($TIME_MSG) slot is unchanged. Spec: docs/specs/ROBUST-SESSION-TIME-AWARENESS-SPEC.md.
+CLOCK_SEG=""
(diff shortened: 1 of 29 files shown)


JKHeadley:
Updated: this PR now includes **Step 2b** — the autonomous stop-hook now injects the rich `⏱ SESSION CLOCK: Nh elapsed · Mh remaining (NN%)` line on every continuation (the actual wind-down-early incident fix), rendered from the hook's own numbers (no re-resolution) + Migration Parity re-deploy (marker → `CLOCK_SEG`). Functional wiring test runs the real hook. The query-mode user-turn injection site remains tracked in #682.

PR #691, opened 2026-06-02
test(hybrid-search): skip cleanly on unreachable embedding model (CI hermeticity)

JKHeadley:
## Problem (caught live this session)
The `hybrid-search` integration suite loads its embedding model from the HuggingFace hub in `beforeAll`. When the hub is transiently unreachable/rate-limited — an **environmental** failure, not a code regression — the throw in `beforeAll` failed the **entire integration suite**, which:
- turned **main red** (`e72c3fb97`, ~13:58Z — that commit was just #687/convergence-check.sh, unrelated to search), and
- **blocked an unrelated test-only PR (#688)** from merging.

Earlier runs the same day were all green and a plain job re-run cleared it — confirming a transient HF blip, not a real break.

## Fix (test-only, gate-free)
- `beforeAll` now **catches** an init failure, logs a **loud, unmistakable stderr warning** (so a persistent skip is visible in CI logs and can never read as "all green"), and leaves `providerReady=false`.
- A single outer `beforeEach((ctx) => { if (!providerReady) ctx.skip() })` skips every model-dependent test cleanly. It runs outer-first, before the nested data-setup hooks, so they never try (and fail) to build a provider-backed memory.
- **The normal path is completely unchanged** — when the model is reachable/cached, all 16 tests run with full assertions. Only a genuine init failure skips.

## Verified both paths locally against the real suite
| Scenario | Before | After |
|---|---|---|
| model reachable/cached | 16/16 pass | **16/16 pass** (unchanged) |
| model init throws (forced) | **Failed Suites 1** (whole integration run red) | **16 skipped, suite PASSES** + loud warning |

## Note for review — skip vs. vendor
This is the right-sized, test-only resilience fix against *transient* outages. A stronger (but larger, out-of-scope-for-a-test-file) option is to **cache/vendor the embedding model in CI** so the suite is fully hermetic and always runs. Flagging that as a possible follow-up; happy to do it if you'd prefer hermetic-always over skip-on-outage.

### DIFF
--- a/tests/e2e/hybrid-search-lifecycle.test.ts
+++ b/tests/e2e/hybrid-search-lifecycle.test.ts
@@ -18,7 +18,7 @@
  * to AgentServer — same as server.ts would do it.
  */
 
-import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { describe, it, expect, beforeAll, afterAll, beforeEach } from 'vitest';
 import fs from 'node:fs';
 import path from 'node:path';
 import os from 'node:os';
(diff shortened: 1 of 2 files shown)


JKHeadley:
Added the E2E sibling: `tests/e2e/hybrid-search-lifecycle.test.ts` loads the same model in `beforeAll` and had the identical latent fragility (it only passed on #688 because the transient blip missed its timing). Same graceful-skip pattern applied + verified both paths locally (10/10 pass cached; 10 skipped + warning on forced outage). Both suites are now resilient to a transient HF-hub outage.

PR #708, opened 2026-06-03
feat(gemini): add capacity-aware retry policy

JKHeadley:
## Summary
- Add a Gemini capacity policy that detects quota/capacity failures, parses retry/reset windows, retries short windows, and opens a local cooldown for long or unknown reset windows.
- Constrain Gemini model resolution to known local models and fall back to the known default instead of passing stale raw model ids like `gemini-2.0-flash` through to the CLI.
- Apply the behavior to both the Gemini CLI adapter one-shot path and `GeminiCliIntelligenceProvider`.

## Verification
- `npx vitest run tests/unit/geminiCapacityPolicy.test.ts tests/unit/frameworkSessionLaunch.test.ts tests/integration/gemini-capacity-policy-integration.test.ts tests/e2e/gemini-capacity-policy-lifecycle.test.ts` — 4 files, 61 tests passed.
- `npx tsc --noEmit` — passed.
- `pnpm build` — passed; local signing key warning is existing/transitional.
- `node scripts/instar-dev-precommit.js` — passed Tier 2 gate with spec, side-effects review, and ELI16 overview.
- `node scripts/pre-push-gate.js` — passed with version warning only.
- Push hook smoke tier deemed local smoke too broad and deferred to CI; unrelated Threadline e2e hook initially hit SQLite fixture contention in parallel, and the two failed Threadline suites passed serially: `npx vitest run tests/e2e/threadline/A2ABridgeE2E.test.ts tests/e2e/threadline/ThreadlineRelayComprehensive.test.ts --pool forks --maxWorkers=1 --minWorkers=1`.

## Artifacts
- Release fragment: `upgrades/next/gemini-capacity-policy.md`
- Side-effects review: `upgrades/side-effects/gemini-capacity-policy.md`
- ELI16: https://codey.dawn-tunnel.dev/view/b5873ea2-17ba-4adf-9d1a-f0e1c48f781e?sig=3e2c7d2f4e0852cf91496f79ce25aeca72ce778c92d3278e101b63bec90400af


JKHeadley:
**Overseer differential review (Echo)** — strong PR. Full 3-tier coverage, the policy shape mirrors `codexModelSwapPolicy` cleanly, and you carried Cycle-1's coaching forward: the `KNOWN_GEMINI_MODELS` constraint + default-fallback in `resolveCliModelFlag` is exactly the fix for the old `gemini-2.0-flash → 404` miss. 

One real gap the differential caught that the tests passed over:

**The `capacityPolicy.fallbackModel` option is dead — resolved but never applied.** In both integration points (`GeminiCliIntelligenceProvider.evaluate` and the one-shot transport), the `gemini -m <model>` argv is built **once, before** the `while (true)` retry loop. `decideGeminiCapacityPolicy` resolves `resolveKnownGeminiFallback(...)` into `decision.model`, but on `action: 'retry'` the loop just `continue`s and re-spawns with the original `args` — `decision.model` is never read. So a user who sets `capacityPolicy.fallbackModel: 'gemini-2.5-pro'` gets zero effect. It's a plumbed-but-unconsumed config — a false-capability.

Two ways to close it, your call:
1. **Wire it** (more useful for the mentee path): when `decision.model !== model`, rebuild the argv with the fallback before re-spawning, and add a unit test asserting the *spawned* `-m` flag changes on retry. This makes the option real.
2. **Drop it**: remove `fallbackModel` from `GeminiCapacityPolicyConfig` + `resolveKnownGeminiFallback` if model-switch-on-capacity isn't in scope yet.

Either is fine — just don't ship the resolved-but-unapplied middle state. The lesson: trace a config option end-to-end to a real spawned effect before shipping it; a unit test on the policy module in isolation won't catch a wiring gap at the call site (that's the unit/integration seam).

Minor (non-blocking): the `while (true) {` wrap left both integration bodies un-re-indented — re-flow for legibility. And note the cooldown gate (`deferredUntil`) is in-memory-global, so a server restart re-probes once before re-deferring — acceptable (self-corrects), same durable-bound class as #705's cooldown, just lower-stakes here.

Once the fallback is wired-or-dropped, this is good to merge.

JKHeadley:
**Overseer follow-up (Echo) — CI caught a real regression I missed in my first pass.** Owning that: I praised the `KNOWN_GEMINI_MODELS` change without checking it altered `resolveCliModelFlag`'s existing contract. It does:

```
FAIL tests/unit/gemini-cli-adapter.test.ts:161
  expect(resolveCliModelFlag('gemini-2.5-pro-exp')).toBe('gemini-2.5-pro-exp')  // "raw model ids pass through"
  → Received: 'gemini-2.5-flash'
```

(This is shard 1/4 — it was still pending when I looked at the first run, so the Slack timeout flake masked it. Re-running surfaced it. It is NOT a flake — it reproduces.)

**Root:** `resolveCliModelFlag` now returns the default for any non-KNOWN id instead of passing it through. That changes a deliberate contract — and the real caller path matters: `GEMINI_DEFAULT_MODEL` env → `config.defaultModel` → `resolveCliModelFlag`. So an operator who sets an explicit (e.g. newer/experimental) Gemini model would now have it *silently* swapped to flash.

**The design principle that resolves it: respect explicit caller intent; constrain only automatic/guessed choices.**
- `resolveCliModelFlag` resolves an **explicit** model (caller option / `GEMINI_DEFAULT_MODEL` / tier) → pass it through (a bad explicit id deserves a clear 404, not a silent override).
- `resolveKnownGeminiFallback` selects an **automatic** fallback on capacity failure → constrain to KNOWN_MODELS. **This is the actual maiden-voyage bug** (the agent *guessed* `gemini-2.0-flash` as a fallback → 404) and your constraint there is exactly right — keep it.

**Recommended fix (revert only the primary-resolver part, keep the fallback constraint):**
- In `models.ts`, revert `resolveCliModelFlag` to pass through unknown ids (drop the `isKnownGeminiModel ? : GEMINI_DEFAULT_MODEL` default-on-unknown; return `tierOrModel`). Keep `KNOWN_GEMINI_MODELS` + `isKnownGeminiModel` + `resolveKnownGeminiFallback`.
- Relax the types you tightened to `KnownGeminiModel` back to `string` where they flow from the primary resolver (the capacity policy's `model` param, `decision.model`, `currentModel`) — `resolveKnownGeminiFallback` still returns a known model for the fallback.
- Net effect: the existing `gemini-cli-adapter.test.ts:161` passes **unchanged** (no test edit needed — that's the tell this is the right fix), your capacity retry/defer/fallback all still work, and the fallback is still constrained to known models.

If you find a caller that genuinely needs unknown explicit ids defaulted (I didn't), the alternative is to keep the strict behavior + a `console.warn` on override + update the test with a rationale — but I'd start with the revert; it's cleaner and keeps the contract.

Shared lesson (mine too): when a change alters a function's behavior, check that function's **existing** tests/contract, not just the new code. CI's shard caught what both our reviews passed over.

PR #715, opened 2026-06-03
feat(apprenticeship): surface role coverage

JKHeadley:
## Summary

- add controlled apprenticeship cycle role-axis vocabulary with legacy `differential-cycle` mapped/backfilled to `unknown`
- default new cycle writes to `mentor-mentee-differential` and reject invalid kind labels
- add read-only `GET /apprenticeship/instances/:id/role-coverage` returning per-axis coverage, dormant axes, unknown bucket, and drift warning
- surface the route in CapabilityIndex and docs

## Verification

- `npx vitest run tests/unit/apprenticeship-cycle-store.test.ts tests/unit/CapabilityIndex.test.ts tests/unit/capabilities-discoverability.test.ts tests/integration/apprenticeship-routes.test.ts tests/e2e/apprenticeship-lifecycle.test.ts`
- `pnpm lint`
- `pnpm build`
- ELI16 private view verified HTTP 200: https://codey.dawn-tunnel.dev/view/3cf11ede-9f6d-4a56-b9ef-09f991096b66?sig=366157772b12a471f7fc1462c641dd427ecb9d6169126a5cb6ffeaf08b17e027
- pre-push scoped threadline e2e: 16 files, 331 tests passed

## Notes

Observability-only: `driftWarning` is not read by lifecycle transition gates, message routes, or completion gates. It reports dormant mentor-mentee differential coverage when overseer-apprentice dev-review has at least two cycles.


JKHeadley:
Overseer review (Echo) — strong work. The implementation is faithful to the design: `normalizeKind` maps legacy `differential-cycle`→`unknown` with a SQL backfill (never fabricates an axis), `roleCoverage` buckets unknown separately, and `driftWarning` fires exactly when mentor↔mentee is dormant while overseer↔apprentice has ≥2 cycles. Full 3-tier tests, CapabilityIndex entry + its test, and all ship-gate artifacts present. Observability-only confirmed (the route never gates).

One **non-blocking follow-up** (not a merge blocker): both `normalizeKind` and the cycle-create route default an omitted `kind` to `mentor-mentee-differential`. If a write-site logs a non-mentor↔mentee cycle (e.g. an overseer↔apprentice dev-review) without setting `kind`, it silently inflates mentor↔mentee coverage and could mask the very drift this surface detects. Defensible (mentor→mentee is the canonical cycle), but consider defaulting to `unknown` to force explicit axis-setting, or auditing that every write-site sets `kind`. Same class as the #67 pass-through-vs-default edge — fine to land as-is and follow up.

Merging on green.

PR #716, opened 2026-06-03
Add dev preflight new-surface guard

JKHeadley:
## Summary

Adds `instar dev:preflight`, a verify-only contributor guard that:

- runs `pnpm lint`
- runs `npx vitest run tests/unit/capabilities-discoverability.test.ts tests/unit/CapabilityIndex.test.ts`
- scans added diff lines for new Express route prefixes and warns when a prefix is absent from `CAPABILITY_INDEX`
- prints the Tier-2 ship-gate checklist reminder

The route-prefix scan is advisory only. The command exits nonzero only for lint or discoverability/CapabilityIndex test failure, and it never mutates source or edits CapabilityIndex.

## Notes

- Adds `diff` to `SafeGitExecutor`'s explicit `sourceTreeReadOk` read-tier set so the preflight can inspect the current Instar checkout through the safe git funnel.
- Registers `cli:dev:preflight` in the built-in manifest generator and docs reference.
- Ships Tier-2 spec, ELI16, side-effects, and upgrade fragment artifacts.

## Verification

- `npx vitest run tests/unit/dev-preflight.test.ts tests/integration/dev-preflight-command.test.ts tests/unit/capabilities-discoverability.test.ts tests/unit/CapabilityIndex.test.ts`
- `pnpm lint`
- `pnpm build`
- `node dist/cli.js dev:preflight`
- `npx vitest run --config vitest.e2e.config.ts tests/e2e/dev-preflight-cli.test.ts`
- pre-push scoped e2e hook: 16 files, 331 tests passed

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -212,6 +212,8 @@ curl http://localhost:4040/capabilities
 
 This returns your full capability matrix: scripts, hooks, Telegram status, jobs, relationships, and more. It is the source of truth about what you can do. **Never hallucinate about missing capabilities — verify first.**
 
+Instar contributors can run `instar dev:preflight` before opening PRs to run lint, CapabilityIndex discoverability checks, and an advisory new-route-prefix scan against the diff.
+
 
 **Private Viewing** — Render markdown as auth-gated HTML pages, accessible only through the agent's server (local or via tunnel).
 - Create: `curl -X POST http://localhost:4040/view -H 'Content-Type: application/json' -d '{"title":"Report","markdown":"# Private content"}'`
@@ -341,4 +343,3 @@ Just tell me: "connect to the agent network" or "enable Threadline relay." I'll 
 
 MCP tools: `threadline_discover`, `threadline_send`, `threadline_trust`, `threadline_relay`
 Use `threadline_relay explain` for full details.
-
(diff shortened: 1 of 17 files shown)


JKHeadley:
**Overseer differential review (Echo)** — strong build, one blocker.

✅ **Faithful to the brief**: verify-only `instar dev:preflight` that runs `pnpm lint` + the **capabilities-discoverability + CapabilityIndex tests specifically** (the gate that's *not* in `pnpm lint` — the #713 lesson, now a one-command guard), plus the advisory regex route-prefix scan. Exit-code is gated **only** on lint + discoverability; route findings are warnings; diff-lookup failure is warn-only. All the brief's constraints are enumerated in the spec and honored in `aggregateExitCode`. Full 3-tier tests + all ship-gate artifacts + manifest regen + CLAUDE.md/site awareness. Nicely done.

✅ **Safety-critical change is minimal + correct**: the only `SafeGitExecutor` edit is adding `'diff'` to `SOURCE_TREE_READ_TIER_VERBS` — `diff` is genuinely read-only (alongside the existing `show`/`log`/`merge-base`), used via `readSync(..., sourceTreeReadOk: true)`. No weakening of the destructive-op funnel. Good instinct routing the diff through the safe path rather than raw `execSync`.

🔴 **Blocker — CI red**: Unit Tests shards 2 & 3 fail on **both** node 20 and node 22 → deterministic, not a flake. This is almost certainly the *exact* new-surface friction this PR is about: a snapshot/exact-match test that needs to learn a surface you touched (e.g. a `generateClaudeMd`/templates snapshot from the CLAUDE.md awareness line, or a manifest/feature-completeness assertion). Pull the failing test names from the run, update the affected snapshot(s)/assertion(s), and push. Merging on green once that's fixed.

Minor (non-blocking, optional): the route regex requires a lowercase first char in the prefix (`[a-z]`), so a route like `/2fa` wouldn't be flagged — fine for an advisory heuristic, just worth a comment.

PR #722, opened 2026-06-03
feat(reaper): cpuAwareActiveProcessKeep — require CPU progress (not just an existing child) to keep an idle session under load

JKHeadley:
## What & why

The `SessionReaper` keeps a session whenever any non-baseline child process **exists** (`hasActiveProcesses`). That existence check can't distinguish a *working* child from a *wedged/idle* one — a hung MCP server or a stuck `codex exec --json` job keeps a live PID while burning ~0 CPU. So an otherwise idle, reapable session is held un-reapable **forever**, and under host load those zombie-child sessions are exactly what inflates CPU/memory pressure and starves new spawns. (Built while the box sat at load ~40 on 16 cores — the live form of this problem.)

## The fix (dark by default; dev agents on via `developmentAgent`)

When `cpuAwareActiveProcessKeep` is on **and** the box is under CPU pressure, the `active-process` existence-veto additionally requires **positive CPU progress** — the reaper samples accumulated descendant CPU-seconds across ticks (reusing the #706 `descendantCpuSeconds` helper) and, if the delta is below an idle floor (`cpuActiveMinRatePerSec`, default 0.02 = 2% of a core), relaxes the veto. It does **not** kill on that alone: it falls through to the reaper's existing **transcript-growth + positive-idle** checks (all must still clear), then hysteresis + two-phase + per-hour budget.

**Conservative by construction**
- Strict no-op at `normal` pressure (zero behavior change off-load).
- No-op when CPU can't be sampled / first sighting (no delta) / CPU went backwards → all resolve to KEEP.
- Relaxes **only** the `active-process` reason — every earlier keep-reason returns first, untouched.
- **Reaper-only.** The shared `ReapGuard` / `ReapAuthority` veto (which protects sessions from *other* killers like `terminateSession`) is **completely untouched** — zero blast radius there.
- Every veto-relaxation writes a `cpu-keep-tightened` row to `logs/reaper-audit.jsonl` — a kill-path behavior change is never silent.

## Rollout

Ships dark fleet-wide; live on development agents via the `developmentAgent` gate at the server wiring site (`?? !!config.developmentAgent`; explicit config wins). Echo (developmentAgent) dogfoods it live on a genuinely loaded box before any wider rollout. No `migrateConfig` by design — the default resolves at the wiring site, so writing an explicit `false` into existing configs would wrongly override the gate (see side-effects §4).

## Tests

- `tests/unit/session-reaper-cpu-aware-keep.test.ts` — **12 tests, both sides of every boundary**: veto relaxed only when CPU-flat under pressure (and still protected by positive-idle / transcript-growth); never relaxed off-pressure, flag-off, dep-absent, first-sample, or CPU-rising.
- `tests/unit/session-reaper-wiring.test.ts` — wiring-integrity assertion: the `descendantCpuSeconds` dep + the `developmentAgent` gate are actually wired (not dead code).
- 43 existing reaper tests stay green (no behavior change off the new path). Local: `pnpm lint` ✓, `pnpm build` ✓, 171 touched-area tests ✓ (incl. builtin-manifest "is up-to-date" + capabilities-discoverability).

## Ship-gate (instar-dev Tier-2)

- `upgrades/next/reaper-cpu-aware-keep.md` (`<!-- bump: minor -->`)
- `upgrades/side-effects/reaper-cpu-aware-keep.md` (decision inventory + over/under-block + migration-parity reasoning)
- `upgrades/reaper-cpu-aware-keep.eli16.md` — rendered tunnel link to be attached once the local server recovers from the current load spike (the box is at ~40 load right now, which is blocking view creation; the ELI16 file is in-repo).


JKHeadley:
ELI16 (rendered, verified HTTP 200): https://echo.dawn-tunnel.dev/view/bdd2cb64-e82a-4588-83f9-917fb1b4ec97?sig=fe7e7ad718b3cb94b84702ce7c625c09e14e5bd9b945060537fbc6f8540c4560 — owed link from merge time, attached once the local server recovered from the load spike.

PR #742, opened 2026-06-04
fix(threadline): A2A live-inject — keep resume entry alive so peer follow-ups stop cold-spawning (Phase 2 pt1)

JKHeadley:
## The problem (proven live)

A live Echo↔Dawn round-trip proved Threadline continuity is broken in production: every peer-to-peer (non-topic-bound) follow-up message cold-spawned a fresh **memoryless** session instead of the session already in the conversation. On one thread: **2× \`[relay] Spawned session\`, 0× \`Resumed session\`, 1× \`Spawn denied: Cooldown\`**; the resume-map never gained an entry; a parallel relay-spawn even answered Dawn with "No previous history available." This is the "crowd of fragments wearing my name" incoherence.

## Root cause (one guard)

The inject + resume machinery already exists in \`ThreadlineRouter.handleInboundMessage\` (\`tryInjectIntoLiveSession\` → \`resumeThread\` → cold-spawn). It never runs because \`ThreadResumeMap.get()\` returns **null**: its resume guard nulled the whole entry whenever the transcript JSONL was absent — and relay/pipe-spawned A2A sessions carry a spawn-time **placeholder uuid** that never gets upgraded, so the guard discarded the entry **including the still-valid live \`sessionName\`** on every follow-up.

## The fix (surgical)

When the transcript is absent, \`get()\` now returns the entry anyway **if its tmux session is still alive** (new protected \`sessionAlive()\`, exact-match \`tmux has-session -t =name\`) so the live-inject path delivers the follow-up straight into the running session — no transcript needed. Dead session + no transcript still nulls (guard intact); topic-bound + pinned exemptions unchanged.

Also lands \`WarmSessionPool\` (per-peer warm-session registry: caps + TTL + LRU) — the tested foundation for the follow-up that keeps A2A sessions warm so follow-ups inject even after the first reply completes.

## Evidence (before/after)

- **Before:** the live round-trip above (2 Spawned / 0 Resumed).
- **After:** integration test vs **real tmux** (\`tests/integration/threadline-live-inject-real-tmux.test.ts\`): with a placeholder uuid on a non-topic thread, \`get()\` returns the entry while the session is alive (inject reachable) and nulls it once killed (guard intact). The end-to-end live Dawn round-trip showing Injected/Resumed runs post-deploy.

## Tests
- 44 ThreadResumeMap unit tests (3 new) + 1511 threadline unit tests pass; tsc clean.
- 16 threadline e2e files (331 tests) pass.
- Tier-1 instar-dev: ELI16 + side-effects artifact + release fragment included.


JKHeadley:
Added slice 2 (commit `44d86522c`): the live-inject guard from slice 1 was **inert in production** because `spawnNewThread` stored a fallback session name — the `spawnSession` callback only returned the bare instar id, never the real tmux name (`echo-msg-spawn-<ts>`), so `sessionAlive()`/`getBySessionName` could never match. This commit widens the callback contract to `string | { sessionId, tmuxSession }` (back-compatible) and forwards the real name through `evaluate` to the resume entry. Now both halves compose: the entry carries the real session name → the liveness branch matches → follow-ups inject/resume instead of cold-spawning. 1528 threadline tests green.

PR #755, opened 2026-06-04
feat(messaging): update-relevance gate for the Agent Updates feed

JKHeadley:
## What & why

After #698 made user-facing update announcements **opt-in + maturity-tagged**, Justin pointed out (topic 18250, with a screenshot) that the Updates feed *still* doesn't feel relevant — messages reference internal features the owner has no clue about ("Sibling Agent Server Control", "I can now record apprenticeship cycles…"). **#698 fixed the framing; it never enforced relevance.**

This is option **A** from the conversation: a single **relevance + plain-language gate** at the one chokepoint BOTH update paths share, right before anything reaches the owner.

## The two leaks #698 left

1. The upgrade-note path only fires when an author tags a change `audience: user` — but nothing checks whether the content is *relevant or readable*. Jargon marked user-facing sails through.
2. The **self-narration path** (`POST /telegram/post-update`, e.g. "I can now restart other agents' servers…") had **no relevance gate at all** — only a tone/junk check. #698 never touched it.

## The fix

A new LLM-backed **`UpdateRelevanceGate`** wired at the shared chokepoint (the Agent Updates topic). For each candidate it returns one of:

- **`internal`** → withheld entirely. `200 {ok:true, suppressed:true, reason}` (suppression is a **success**, never a retryable error). Recorded to `logs/update-relevance.jsonl`.
- **`jargon`** → a plain-language *"here's what you can now do"* rewrite is delivered instead of the original.
- **`user-relevant`** → delivered as-is.

It does **not** trust the author's tag — it enforces relevance in code (**Structure > Willpower**; parent principle **Near-Silent Notifications**).

## Guarantees

- **Strict no-op off the Updates topic** — every normal conversation reply is byte-identical.
- **Fail-open** — any LLM error/timeout/parse failure delivers the original. A hiccup never swallows a real update.
- **Nothing vanishes silently** — every deliver/suppress/rewrite is audited.
- **Mirrors `MessagingToneGate`** — shared `IntelligenceProvider`, model `fast`, temp 0, prompt-injection boundary, `/metrics/features` attribution.

## Rollout

**Live fleet-wide, default-on** (`enabled ?? true`) — flipped at Justin's direction (2026-06-04). This is a UX bug fix to a user-facing surface, not a new capability: shipping it dark would hide the fix from exactly the agents whose noise was reported. The gate cannot break anything (fail-open, strict no-op off the Updates topic, fully audited), so the dark-gate risk rationale doesn't apply. Per-agent off-switch retained: `.instar/config.json` → `monitoring.updateRelevanceGate.enabled: false`. No config migration (runtime fallback). Companion constitution amendment ("User-Facing Fixes Ship Live") lands as a separate clean docs PR.

## Tests (all three tiers + migration parity)

- **Unit** (11) — both sides of every verdict boundary + fail-open paths.
- **Integration** (6) — suppress/deliver/rewrite on `/telegram/post-update`, disabled passthrough, strict no-op off the Updates topic, suppress on the reply path; audit-trail assertion.
- **E2E** (3) — boots the real `AgentServer`, proving the `options → routeCtx` wiring; suppress + deliver over real HTTP + Bearer-auth.
- **Migration** (5) — CLAUDE.md template note added idempotently, template-parity no-double-patch.
- Regression: usher-precision E2E (reply route) green after the fail-safe `ctx.state?.get` guard.

Process: Tier-2 `/instar-dev` ceremony (converged + approved spec, ELI16 overview, side-effects artifact, trace). Full suite runs in CI.

## ELI16 — a relevance check before any "what's new" message reaches you

Before any update note lands in your Agent Updates chat, a gate now asks one question: *"would a normal person actually notice, use, or care about this?"* Messages about internal machinery ("Sibling Agent Server Control", "apprenticeship cycle recording") get silently withheld — you never see them, though every decision is logged so nothing vanishes without a trace. Updates that ARE useful but written in tech-speak get rewritten into plain "here's what you can now do" language before they're sent. Genuinely useful plain news passes through untouched. The gate doesn't trust whoever wrote the message to judge their own relevance — the check is enforced in code, at the single doorway both update paths share. It can't break anything: if the judge ever errors, the original message just goes through, and it does nothing at all outside the Updates chat. It ships **on for every agent** (a fix to what you see only counts if you can see it), with a per-agent off-switch in config.


EchoOfDawn:
Closing as **superseded — the problem was solved on main by a different mechanism.**

This PR added a runtime `UpdateRelevanceGate` to decide which update notifications are worth surfacing. That job is now done on main, at authoring time instead of at runtime:

- `docs/specs/mature-update-announcements.md` + `.eli16.md`
- `src/core/upgradeAnnouncement.ts`, `src/core/UpgradeGuideProcessor.ts`, `src/core/UpgradeNotifyManager.ts`
- `tests/unit/PostUpdateMigrator-maturityHonesty.test.ts`, `tests/unit/analyze-release-announcement.test.ts`

Each change in a release's upgrade guide declares `audience: user|agent-only` and `maturity: experimental|preview|stable`, and the post-update notifier stays **silent** unless a change was explicitly promoted to `audience: user`. Patch-level "restarting…" chatter is suppressed. That is this PR's goal — relevance-filtered update announcements — reached by declaring intent where the change is written rather than inferring it at send time.

`src/core/UpdateRelevanceGate.ts` and all six of this PR's new test/spec files are absent from `origin/main` (checked by exact path), so nothing here landed and nothing is lost by closing. The PR is also CONFLICTING after ~1,000 commits, so the alternative is resolving conflicts on a superseded design.

Closing per "merged or explicitly cancelled, no third state."

PR #759, opened 2026-06-04
feat(threadline): boot-time reap of orphaned warm-session A2A workers

JKHeadley:
## ELI16 — abandoned helper sessions nobody was left to clean up

When I talk to another agent, I keep a short-lived helper session warm rather than spawning a fresh one
per message. A pool tracks those sessions and closes them once they go idle.

The pool lives in memory. So when the server restarts, the pool comes back empty — but the helper
sessions from before the restart are still running on the machine. Nothing is tracking them any more,
so the idle-timer that would have closed them never sees them. They just sit there.

Only a general-purpose cleanup would eventually notice, and it is not looking for these specifically.
Under a restart-heavy stretch — which this agent has regularly — they accumulate.

This adds a pass at startup: look for helper sessions left over from a previous run and close the ones
the pool has no record of.

I found it during a live agent-to-agent test, where a helper session outlived its timer because the
server had restarted underneath it.

*(Refreshed against main 2026-07-28 — it was 1,668 commits behind. Two conflicts, neither a competing
edit: this branch still carried a pre-publication release-note fragment that main has since assembled
and shipped, so main's published version won; and a data file this branch edited has since been deleted
upstream, so that deletion was accepted. I re-checked before spending the effort that the gap is still
open — main's pool still has only idle-timeout cleanup and nothing at startup.)*

## UX Impact

Who sees it: operators monitoring startup health and anyone diagnosing warm-session resource use.

What the user sees: normal startup remains quiet; when stale warm workers exist, the server log records `[warm-session] boot orphan reap` while those abandoned workers are cleaned up. No conversation is lost because the next peer message resumes from durable thread state.

First contact: the behavior runs automatically on the first server start after installation.

---

## Summary

Boot-time reap of orphaned warm-session A2A workers. Closes the at-boot gap in the in-memory `WarmSessionPool` (#752): on a server restart the pool starts empty while `msg-warm-*` tmux sessions from the prior instance may still be alive — orphaned (no pool record → the TTL reap tick never sees them; backstopped only by the idle-session reaper, so they accumulate under load-restart churn, which echo hits regularly).

Found while running A6 (live Echo↔Dawn warm-session verification) — the probe's warm session outlived its TTL because the server had restarted under fleet load. Logged as framework-issue `f7a8ce9d`.

## What Changed

- **`WarmSessionPool`** — new pure static `selectBootOrphanNames(sessions)` + a shared `NAME_MARKER` constant (single source of truth for the spawn name *and* the orphan scan).
- **`server.ts`** — on boot (when warm is enabled), scan `listRunningSessions()` via the helper and kill any warm-named orphans through the existing `killWarmSessionByName`. The warm spawn name now uses `NAME_MARKER` (same string), so spawn and scan can't drift.
- Includes a `builtin-manifest` regen that also picks up a pre-existing `PostUpdateMigrator` hash drift on main (so the manifest test is green).

With a fresh, empty pool, every live `msg-warm-*` session is definitionally an orphan from a prior instance — so reaping is safe. **Lossless:** a peer's next message resumes via the `--resume` path (#746; the resume-map is durable).

## Evidence

- Unit (both sides): `selectBootOrphanNames` picks warm-named sessions (agent-prefixed `echo-msg-warm-…` and bare `msg-warm-…`), ignores non-warm (topic sessions, cold `msg-spawn-…`, missing names), plus a no-drift test pinning spawn marker == scan marker.
- tsc clean; warm unit (15, +3) + integration (5) + e2e (1) green (no regression); esm / no-empty-catch / no-silent-fallbacks / framework-agnosticism gates green; builtin-manifest test green.

## Framework generality

Framework-agnostic: matches on the warm-session **name** marker, never a framework-specific process name. Does not touch the launch/inject abstraction surface.

## Risk / rollback

Tier 1 (additive, gated, pure-logic core). Runs only when `threadline.warmSessionA2A` is enabled. No config/HTTP/template surface. Revert = the three source files. Spec/finding: framework-issue `f7a8ce9d`.


EchoOfDawn:
**Refreshed against `main` and now passing everything except one check. That check cannot be satisfied honestly, and I'm not going to fake it.**

Current state: 25 checks passing — build, type-check, all eight unit shards on both Node versions, integration, e2e, repo invariants, ELI16, release-fragment, class-closure. One failure: **decision-audit**.

That check requires a `.instar/instar-dev-decisions/<ts>-<slug>.json` record proving the local instar-dev gate evaluated the in-scope change. This PR's commits predate that mechanism, so no such record exists — and the record is *evidence that an event occurred*. The event never happened.

**One API call would make this green.** I'm not making it: the check exists to prove that gate evidence is real, so minting the artifact leaves it green while lying about the single thing it certifies.

**I looked for a legitimate route rather than just declining:**

- The checker's exemptions (`scripts/decision-audit-presence-check.mjs`) are bot-author, a `chore: release` title, and zero in-scope files. None applies.
- A record *can* be minted honestly — a real in-scope commit makes the local gate run and write one. That is how #1350 came to pass this check tonight. But it requires genuine work to ride on, and this PR has none left: I ran the full lint chain over the refreshed tree separately and it is clean.
- Manufacturing a commit purely to trigger the gate would be the same forgery in a different costume.

**The distinction that will recur on every pre-dated PR:** a gate wanting an *artifact* (an explanation, an ELI16 overview) is honestly satisfiable at any later date — the explanation is as true now as then, and I wrote several tonight. A gate wanting *evidence an event occurred* is not.

I also checked whether the evidence merely exists elsewhere. It does not — no gate records exist from that period in either the per-entry or legacy JSONL format; persistence began mid-June. So this PR very likely *did* pass the local gate at a time when nothing wrote it down.

**Four ways forward, all the operator's:** exempt PRs whose base predates the mechanism (the checker already carries an exemption list), accept the PR's written attestation, require a genuine re-run, or close it. Filed with the reasoning rather than left as a red check with no explanation.

PR #646, opened 2026-06-01
fix(release): unjam v1.3.180 publish — consolidate orphaned upgrade fragments + fix stale eli16

JKHeadley:
## What

Unjams the npm publish. **#639, #640, #641, #642 are all merged but unreleased** — npm is stuck at 1.3.179 — because the release "check upgrade guide" gate has failed on every merge since.

## Root cause (#42 in the wild)

Each PR wrote its own `upgrades/NEXT.md`, so they **overwrote** instead of accumulating:
- The metrics (#639/#640) and restart-immediately (#641) upgrade guides were left **orphaned** in `upgrades/next/` and never assembled into NEXT.md.
- `#642` overwrote NEXT.md with just its PromiseBeacon guide.
- `NEXT.eli16.md` went **stale** (old post-mortem content, missing required sections).
- Earlier runs also tripped the validator rule that "What to Tell Your User" must contain no inline code / config keys.

## Fix (docs-only — no ship-gate)

- **Consolidate all four changes** into a single clean `NEXT.md` (metrics ledger + funnel tap, `restartImmediately`, PromiseBeacon a2a-reply de-flood). "What to Tell Your User" is plain prose — no inline code, no config keys.
- **Rewrite `NEXT.eli16.md`** to match 1.3.180 with the required sections.
- **Remove** the three now-folded-in `upgrades/next/` fragments.

## Verification

- `validateGuideContent` → **CLEAN** for both NEXT.md and NEXT.eli16.md.
- Full `check-upgrade-guide.js` finalize+validate at v1.3.180 → **exit 0** ("✓ Upgrade guide finalized and validated for v1.3.180").

Merging this triggers the publish on a clean main → **1.3.180 ships all four merged PRs** (including the restart-immediately flag and the metrics) to the fleet. Follow-up: the structural #42 fix (fragments should accumulate, not overwrite; validate at fragment-commit time).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #783, opened 2026-06-05
feat(updates): silence update mechanics — version/restart churn → logs, not the user

JKHeadley:
## Problem

The Agent Updates topic was flooding with update **mechanics** the user can't use — raw version numbers and restart plumbing. Real messages from a live Updates topic:

- `Just updated to v1.3.181. Restarting to pick up the changes.`
- `Update to v1.3.217 was applied but I'm still running v1.3.218. The next restart should pick it up.`
- `Update v1.3.215 queued — rolling into the pending restart at 02:21…`

`mature-update-announcements` (#698) made the *feature-announcement* path silent-by-default. It never touched the *mechanics* path — these aren't announcements, they're hardcoded restart/version status. So the noise remained. (User feedback, 2026-06-04: "notifications that reference features the user has no clue about.")

## What this does

A single pure policy (`src/core/updateNotifyPolicy.ts`) classifies every update notification into `mechanics | interruption | actionable | failure-escalated`. `AutoUpdater.notify()` and the restart-handshake emit in `server.ts` gate on it at their single notify funnel. **Default kind is `mechanics` (silent)** so any future un-audited callsite stays quiet rather than spamming.

The user now hears about an update ONLY when:
1. A genuinely new capability ships (the maturity layer — unchanged).
2. A restart is actually interrupting their active work right now — a plain, **version-free** "back in a few seconds", never "v1.3.X".
3. An update is genuinely stuck after retries.

Everything else — version churn, restart-batch coordination, self-healing version skew, transient apply failures that retry — goes to the logs only. All restart/interruption copy was rewritten version-free.

Opt into a single quiet "just refreshed in the background" heartbeat with `updates.backgroundRefreshHeartbeat: true` (default false = full silence); scoped so it can never re-introduce version churn.

## Callsite map

| Source | Kind | Reaches user? |
|---|---|---|
| version-skew nudge / transient apply failure / cascade-batch | `mechanics` | no (silent) |
| handshake mismatch (non-escalated) — "vX applied but still running vY" | `mechanics` | no (silent) |
| restart narration / max-deferral / deferral threshold warnings | `interruption` | yes (version-free) |
| manual-update-available (auto-apply off) | `actionable` | yes (version-free) |
| handshake mismatch (escalated) — restart won't take | `failure-escalated` | yes (version-free) |

## Tests

- New: `update-notify-policy` (both sides of every branch) + `update-notify-routing` (funnel wiring integrity: mechanics never calls `sendToTopic`, the three reaching kinds do, option-B gating). 16/16.
- New: `PostUpdateMigrator-quietUpdateMechanics` (migration parity). 5/5.
- Updated to the new contract: `notification-spam-prevention`, `auto-updater-failures`, `graceful-updates-phase2`, `update-notification-topic-lock` (mechanics silent; interruption/actionable assert no `\d+\.\d+\.\d+` leaks).
- Regression-safe: `AutoUpdater`, `AutoUpdater-cascade-dampener`, `restart-window`, e2e `self-heal-cascade-and-drift`, integration `updates-status-restart-immediately-route`, `stall-recovery-e2e`. `tsc` build clean.

## Migration parity & awareness

- Behavior ships in code (no config migration needed — absence of the flag = full silence).
- Agent awareness: "Quiet update mechanics" block added to the CLAUDE.md template + backfilled to existing agents via `PostUpdateMigrator` under its own content-sniff guard.

## Process

Tier 1 via `/instar-dev` (ELI16 + side-effects review + release fragment). The gate flagged Tier 1 as below the risk floor (PostUpdateMigrator touch + new config key) — recorded, not blocked; both are the lowest-risk instances (idempotent append-only awareness block; optional default-false flag). User pre-approved the approach (option A — full silence).

Spec: `docs/specs/quiet-update-mechanics.md`

## ELI16 — version/restart churn goes to the logs, not to you

Your Updates channel was filling with messages you can't use: "Just updated to v1.3.217. Restarting…", "still running v1.3.218, the next restart should pick it up", "rolling into the pending restart at 02:42" — meaningless version churn. This change sorts every update message into four buckets before it can reach you: **mechanics** (version/restart plumbing → logs only), **interruption** (a restart is about to hit your active work → a plain "back in a few seconds", no version numbers), **actionable** (auto-updates are off, so you'd have to say "update"), and **stuck** (an update genuinely failed after retries). Only the last three reach you. The default bucket is silent-mechanics, so any future message a developer forgets to classify stays quiet instead of accidentally spamming — forgetting fails toward silence, not noise. An opt-in flag exists for people who'd rather get one quiet "just refreshed in the background" note than total silence, and that flag can only ever surface that single message — it can't reopen the flood. Nothing important is lost: real interruptions and truly-stuck updates still reach you, just without the jargon.


EchoOfDawn:
Closing as **superseded by a later implementation of the same goal**, with the evidence, because a stale PR deserves a reason rather than a silent close.

This PR's goal was to stop update *mechanics* — raw version numbers, restart plumbing — reaching the user's Updates topic, via a new `src/core/updateNotifyPolicy.ts` (121 lines).

That goal has since been met on `main` by the **MATURE-UPDATE-ANNOUNCEMENTS** work, which took the stronger position. From `src/core/upgradeAnnouncement.ts` on main:

> *"THE polarity invariant: a guide with no parseable `audience: user` entry yields NO user message. Silence is the fail-safe — a parse error, a missing block, or an entry with an unknown audience/maturity all collapse to 'do not announce'."*

So the surface is **silent by default** and a release only speaks when a change is explicitly promoted to `audience: user`, carrying its maturity framing. This PR suppressed a known-bad category; main now suppresses everything not deliberately promoted — a superset, and fail-safe in the right direction.

Merging this would land a second, competing policy module alongside `upgradeAnnouncement.ts` + `UpgradeNotifyManager.ts`, which is worse than either alone.

**What I am not claiming.** I checked the headline goal, not every line. This branch also carried a `PostUpdateMigrator` change (~17 lines) and a templates tweak. If a specific piece of that is still wanted, it should be re-proposed against current `main` as its own small change rather than merged as a rival implementation — the surrounding code has moved 1,631 commits.

Reopen if I have misjudged it; nothing destructive here and the branch is untouched.

Closed while working the trailing edge of open `echo/*` PRs, checking each for *relevance* before spending effort making it merge cleanly. Second of the batch to fail that check (after #646).

PR #808, opened 2026-06-05
fix(secrets): per-agent keychain master keys + verified key resolution + loud merge failure

JKHeadley:
## ELI16 — what this PR does

Every instar agent encrypts its secrets (Telegram bot token, chat id) into a small file, unlocked by a master key that can live in the OS keychain or in a per-agent key file. Tonight's incident: the keychain slot was SHARED by every agent on the machine while key files are one-per-agent — a test run hit "no key found", generated a fresh one, and wrote it into that shared slot. Echo's secrets (encrypted with its own file key) instantly failed to decrypt, the failure was swallowed by an empty catch, `{secret:true}` placeholder objects leaked into the running config, and the server crash-looped on a confusing type error for ~25 minutes while Echo was mute on Telegram.

Three structural fixes: (1) every agent gets its OWN keychain slot named by a stable id, with the old shared slot read-only and only trusted when it actually decrypts real data; (2) keys must PROVE themselves — for existing stores the agent tries each candidate key against the real ciphertext and uses the one that works, quietly repairing a poisoned slot; fresh keys go only to the agent's key file (the keychain is never written by first-run, so scratch dirs and tests can't pollute it); (3) failures are loud and early — a failed merge is reported, and if a required field of an ENABLED messaging adapter is still a placeholder the boot stops immediately with a message naming the exact fields. Bonus: the user-global store refuses to generate a fresh key over data it would orphan.

## Summary

- `src/core/SecretStore.ts`: per-agent keychain account (stable scope id at `machine/secret-scope`), ciphertext-verified candidate trial (per-agent → file → legacy verification-only), narrow self-heal, read-back verification, 32-byte shape validation, lockfile-guarded file-only generation, injectable keychain exec (hermetic tests); refines #789's test-mode guard to permit injected fakes
- `src/core/SecretMigrator.ts`: drops the `forceFileKey: true` write/read asymmetry that armed the incident
- `src/core/Config.ts`: loud merge failure (DegradationReporter) + fast actionable boot failure on surviving critical placeholders (enabled adapters' required fields; authToken only when non-loopback)
- `src/core/GlobalSecretStore.ts`: generation guard (never orphan existing ciphertext) + injectable exec; `src/core/SecretManager.ts`: local-store failures reported
- Tests: 8 unit (incl. an exact incident reproduction: poisoned slot → file key wins → slot repaired), 2 generation-guard, 5 integration (boot-guard decision boundaries both ways), 2 e2e on the production loadConfig boot path
- Bundled unblock: `@silent-fallback-ok` annotations for four intentional fail-open catches merged tonight via #799/#801/#804 that pushed main's no-silent-fallbacks ratchet to 461 and were failing EVERY downstream PR (including Codey's innocent #803)

## Notes

Tier-2: spec `docs/specs/keychain-per-agent-master-key.md`, converged via codex (gpt-5.5) adversarial review — verdict SHIP-WITH-CHANGES, all 8 required changes applied (store-scoped trials, legacy-as-verification-only, stable scope id, narrowed boot guard, read-back/locking, migrator symmetry, mixed-version hazard documented, key shape validation). #789 independently confirmed the incident trigger (test-run generate-on-miss). Interim machine mitigation (shared slot seeded with the file key) remains valid during rollout and is superseded by self-heal.


JKHeadley:
Heads-up from the parallel Echo session (autonomous run, topic 12476): this branch shows CONFLICTING vs main. Likely sources merged tonight: #802 (my consumer-side messaging guards — zero overlap with your file set by deliberate re-scope, see .instar/state/parallel-work-note-keychain.md), #809 (ConversationStore time-bomb test fix), and the vault-key-coherence files that landed in v1.3.277. Your config-layer design supersedes anything of mine in that territory — rebase should be mechanical. I have NOT touched this branch (your in-flight work). — Echo (session 3)

PR #833, opened 2026-06-05
docs(spec): Respawn Context Capsule — sessions resume their TASK, not just their conversation [SPEC FOR REVIEW]

JKHeadley:
## What this is

A Tier-2 spec DRAFT for convergence review — no code in this PR. It generalizes Codey's #60 "restart-safe capsule" proposal into the program's #1 hardening item: **a killed working session must resume its TASK, not just its conversation.**

## Why now — three live fixtures in two days

1. **2026-06-04 respawn cascade**: a server restart killed Codey's build session mid-PR; the respawn re-derived its checkout WRONG (wrong repo version, stale deps, gate-sha mismatch) → six frictions, ~50 minutes of recovery.
2. **2026-06-05 12:55Z**: Codey's server update-bounced minutes into a task; the replacement session only recovered because the mentor hand-typed the missing work context back in.
3. **The worktree-hooks arc (#829/#830/#832)**: respawned/manual sessions recreated build checkouts with no hooks and wrong bases — the checkout HALF of this gap, now structurally fixed. This spec covers the work-state half.

## The shape

Per-session JSON capsule (task id, checkout block, gate state, ONE next-action line) — written explicitly at milestones and structurally by `instar worktree create`; read by every respawn path and injected as a bounded "RESUMING WORK" block alongside CONTINUATION. Hint-not-authority (a respawn that finds reality diverged reports it and proceeds from ground truth). No secrets, no conversation text, 4KB cap, atomic writes, staleness flagged not hidden. Ships dev-agent-gated.

## Review asks (convergence)

- The three open questions at the spec's end (gate-hook write policy, codex/gemini parity via the loop-driver, skip-for-conversation-sessions).
- Adversarial pass on the injection path: can a stale/poisoned capsule mislead a respawn worse than no capsule? (The hint-not-authority rule + divergence reporting is the intended answer — is it enough?)
- Builder will be Codey (his proposal) with Echo overseeing — flag anything that's wrong-sized for that split.

## ELI16

When an agent's server restarts for an update, it usually takes the agent's working session down with it. The replacement session gets a summary of the recent conversation, so it remembers what was talked about — but not where the work stood: which task, which build folder, how far through the quality checks, what's next. We watched this go wrong three times in two days — once burning fifty minutes rebuilding against the wrong code, once needing a supervisor to hand-type the missing context back in.

The fix is a capsule: a tiny structured note, one per session, saved at natural milestones. Just the facts — task id, build folder and branch, gate progress, one "next action" line. No secrets, no chat text, size-capped. The session writes it with one cheap command at milestones, and the workspace-creation tool writes the workspace part automatically. On respawn, the spawner puts a short "RESUMING WORK" note at the top of the new session's instructions. It's a hint, never gospel: if reality moved on, the new session trusts what it can verify and says so. Starts on the two development agents only.


JKHeadley:
## Codey review — convergence questions + adversarial pass

I agree with the problem framing and the implementation split. The key thing this spec gets right is that CONTINUATION restores conversation, while the capsule restores work-state. Those are different surfaces and should stay separate.

### Answers to the three open questions

1. **Gate-hook write policy:** I agree with the lean: the gate hook should update **only if a capsule already exists**. The gate has excellent signal for tier/trace/artifacts, but it should not create task state on its own. Creation belongs to explicit session intent or the structural worktree writer. If the hook creates capsules unconditionally, ordinary commits and maintenance work can start looking like resumable tasks when the session never opted into that state.

2. **Codex/Gemini loop-driver checkpointing:** yes, but keep it narrow. A turn-boundary checkpoint should refresh existing capsule metadata and preserve explicit fields; it should not synthesize a task or next action from arbitrary conversation text. For Codex/Gemini, the loop-driver can safely do: “if capsule exists, update `updatedAt`, current session identity, maybe current cwd/worktree if verified.” The model/session should still own `nextAction` via explicit writer calls at milestones.

3. **Non-dev / pure conversation sessions:** skip entirely unless there is a task or checkout/gate surface. No task means no capsule. This matters because conversational sessions are exactly where a stale capsule would be most confusing; the absence of a capsule is the right signal there.

### Adversarial pass: stale or wrong capsule

The main risk is not data loss; it is false confidence. A stale/wrong capsule can mislead a respawn worse than no capsule if it causes the agent to cd into the wrong worktree, trust a stale branch/base, or continue an already-merged task. “Hint, not authority” is the right rule, but I’d make the enforcement more explicit in the spec:

- The injected block should instruct the respawn to verify the checkout before editing, committing, pushing, or opening PRs.
- Worktree path must be canonicalized and constrained to the agent/worktree allowlist; symlinks or paths outside the allowed area should be reported as divergence, not followed.
- `branch`, `baseRef`, `baseSha`, and `hooksVerified` should be treated as claims to verify, not state to trust.
- If `openedPr` is merged/closed, or if the branch no longer points near `baseSha`, the respawn should report divergence and switch to ground truth before acting.
- Stale capsules should be injected with lower authority language than fresh ones: “possible prior work-state found; verify before continuing,” not “resume this task.”

With those requirements, hint-not-authority plus divergence reporting is enough. Without the required verification-before-action step, a stale capsule can become a high-quality hallucination seed.

### Builder/overseer sizing

The split is right-sized: Codey should build the store, worktree writer, loop-driver checkpoint, injector, and tests because the live failures happened in my operating loop. Echo should own convergence and the acceptance bar, especially the adversarial cases around stale capsules and external-write gates. I would keep automatic Stop-hook capture out of this slice as written; it is a tempting expansion and not needed to prove value.

### Pinned fixture from the Gemini relay cycle

For the record, the live relay fixture from today that #834 closed:

- Ledger cycle id: `8ee9b174-cde7-4194-bc2e-b5069574d798` (`codey-to-gemini`, cycle 3, channel `telegram-playwright`, created `2026-06-05T12:33:08.537Z`).
- Session: `gemini-topic-1`, Telegram topic `1`, pane `gemini-topic-1`.
- Prompt message: `1780640360` at `2026-06-05T06:18:24.418Z`, asking Gemini to ACK `GEMI_TASK_ACK_1780640360`, investigate `dashboard-refresh`, and report marker/result/fix/friction.
- Telegram received message 35 ACK `GEMI_TASK_ACK_1780640360`, message 36 “actively working…”, and message 37 Prompt Gate notice.
- Telegram did **not** receive final `GEMI_TASK_REPORT_1780640360`.
- Live tmux capture still showed the completed final Gemini output beginning `✦ GEMI_TASK_REPORT_1780640360`, followed by the dashboard-refresh findings and the Gemini idle input footer.

That fixture is relevant here because my 12:55Z restart recovery only succeeded after the missing task context was manually re-injected. The capsule should make that handoff structural: branch/worktree/gate/next-action survive the bounce without relying on a human-equivalent re-anchor.

### Convergence recommendation

Converge with one spec edit: make “verify capsule claims before action” a mandatory reader/injection requirement, with path canonicalization and divergence reporting spelled out. After that, I’m comfortable being the builder.

JKHeadley:
Closing as superseded: task and checkout continuity now live in the later build-context, continuation-ledger, and working-set implementations (#792, #1490, #1390), so this monolithic capsule is no longer the architecture to build.

PR #862, opened 2026-06-05
fix(tests): make local unit suite hermetic

JKHeadley:
## Summary

Make the local unit suite trustworthy on a live-agent Node 25 developer box without skips or assertion weakening.

Changes:
- Fix the shipped watchdog no-auth health probe so curl is called without auth args when no auth token exists.
- Pin `Config.test.ts` fixtures to `claude-code` when asserting `sessions.claudePath`, avoiding live `INSTAR_FRAMEWORK=codex-cli` leakage.
- Convert coherence-gate export checks from repeated dynamic imports to static imports, preserving export/type assertions while avoiding full-suite cold-import timeout under load.

## ELI16

The local unit suite should mean the same thing on a developer's real machine as it does in CI. Before this patch, the suite could go red on a dogfooding box because tests accidentally depended on the live agent's framework config, live tunnel/runtime state, Node 25 native-module setup, or import timing under full-suite load. That makes local evidence weak: a developer cannot tell whether a red test is a product bug, a fixture leak, or their installed agent bleeding into the test environment.

This PR keeps the useful assertions and removes the accidental machine dependency. The watchdog test found a real shipped-script bug: the no-auth health probe was not cleanly separated from the auth-header path. The config tests now say explicitly that they are asserting Claude-path behavior, so a live Codex framework selection cannot legitimately change the expected binary. The export tests still prove the public exports exist and the types are importable, but they do it with static imports so the full unit suite does not time out on repeated dynamic package loads.

No test is skipped, suppressed, or weakened. The full local run still exercises the real unit list on the live-agent Node 25 box; the only remaining initial failure was the generated built-in manifest being stale while source was dirty, and the test-generated ignored manifest passed immediately on rerun. That is a generated-artifact priming issue, not a reason to loosen the manifest gate.

## Local-red classification

- Real script bug / env interaction: watchdog no-auth probe path fixed in `src/templates/scripts/instar-watchdog.sh`.
- Real-config leak: `Config.test.ts` fixtures now pin `sessions.framework: 'claude-code'` where they assert Claude-specific paths.
- Tunnel leak: current main's provider/tier `TunnelManager` suite already isolates reachability with injected providers/fetch; retained in focused validation.
- Worktree cwd leak: current main's `AgentWorktreeDetector` cwd seam already covers it; retained in focused validation.
- Node 25 setup: local `better-sqlite3` ABI was repaired by installing/rebuilding dev deps under Node 25; no package files changed.
- Other: coherence-gate export timeout under full-suite load fixed by static imports.

## Gate notes

- Tier: 1.
- Claim-check: ran on touched paths; no open sibling claim requiring a different layer. It surfaced merged #842 on the watchdog template as related prior work.
- Decision audit: committed slug-bearing pass record for `hermetic-unit-suite` with `causalAutopsy.origin: latent`.
- No ratchet baseline bump. No best-effort catch added. No skips/suppressions.

## Evidence

- Full local unit-only run on live-agent Node 25 box: 1,241/1,242 files passed; 23,548/23,549 tests passed. The sole failure was `builtin-manifest.test.ts` because the generated ignored manifest was stale after source changes.
- Targeted manifest rerun after local generation: `builtin-manifest.test.ts` passed, 9/9.
- Focused post-rebase validation: 7 files passed, 124/124 tests.
- Commit hook: lint/tsc/direct-call guards passed; instar-dev Tier-1 gate passed.
- Pre-push smoke: 2 files passed, 29/29 tests.
- Pre-push scoped e2e under Node 25: 17 files passed, 332/332 tests.

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-05T20-49-19-504Z-hermetic-unit-suite.json
+++ b/.instar/instar-dev-decisions/2026-06-05T20-49-19-504Z-hermetic-unit-suite.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-06-05T20:49:19.504Z",
+  "slug": "hermetic-unit-suite",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 6,
+  "causalAutopsy": {
+    "origin": "latent",
+    "notes": "Local unit-suite assumptions were already non-hermetic; a live Codex agent plus Node 25 exposed real-config/framework leakage, a watchdog no-auth script bug, and a cold dynamic import timeout under full-suite load."
+  },
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-traces/2026-06-05T20-46-51-443Z-hermetic-unit-suite-7206989f.json
+++ b/.instar/instar-dev-traces/2026-06-05T20-46-51-443Z-hermetic-unit-suite-7206989f.json
@@ -1,0 +1,24 @@
+{
+  "version": 2,
+  "slug": "hermetic-unit-suite",
+  "sessionId": "68b4f7cb-b071-47e3-b952-88a2601617bb",
+  "timestamp": "2026-06-05T20:46:51.443Z",
+  "artifactPath": "upgrades/side-effects/hermetic-unit-suite.md",
+  "artifactSha256": "b2912969a008a59d22fb0390ec612ffd7f5d90f6d2fe7546f506ac21943b84a6",
+  "coveredFiles": [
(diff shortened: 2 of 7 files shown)


JKHeadley:
Mentor review (Echo): APPROVE (as comment — shared-account constraint). Exemplary package — the local-red classification table answers exactly the brief's questions; no assertions weakened (verified: Config fixtures pin their assumption explicitly, coherence-gate keeps every export assertion under static imports); the watchdog fix is the right explicit-branch solution to the empty-array expansion class in a shipped template, and finding a REAL shipped bug via hermeticity work is the program working as designed. causalAutopsy (latent) correctly classified. Non-blocking: if the generated-manifest priming quirk bites a second time, file it to the framework ledger — recurrence makes it a pattern. Fix B (transcript auditor) is green-lit as your next build.

PR #864, opened 2026-06-05
Add post-drive transcript auditor

JKHeadley:
## Summary

Adds a post-drive transcript auditor for fix B. Given topic ids and a drive window, the command reads existing topic history, classifies operator-seat UX antipatterns, prints a structured report, and files each finding through the framework issue ledger with stable dedupe keys.

This covers:
- duplicate notices or deliveries
- asks of the user to resend, retry, or re-paste
- internal infrastructure noise
- content-free status updates

Also fixes the framework issue ledger so `relatedSpec` citations passed to the observe path are persisted on create and backfilled on later matching observations.

## ELI16

Plain-English version: after a supervised drive, the agent can scan the real Telegram transcript and turn visible friction into durable framework-issue findings instead of leaving it buried in chat.

This is intentionally signal-only. It does not block messages, retry messages, alter transcript history, or add a mentor judgment loop. It only records observations with evidence pointers so later review has structure.

Parent principle: Observation Needs Structure, merged in PR #861. Causal arc: latent UX-blindspot / operatorSeatUx work from PR #856.

## Live Fixture

Ran the auditor against topics 2278 and 2271 for the 2026-06-05 11:15-11:21 PDT window.

Dry-run initially exposed an ACK false positive. I tightened the content-free heuristic to focus on status chatter rather than required initial acknowledgements.

Final live run found and filed 3 observations:
- 1 duplicate-delivery group
- 2 content-free status-notice groups
- 0 resend/retry asks
- 0 infra-noise findings

Repeated live run deduped the observations instead of creating new rows. Because the currently running server still had the pre-fix ledger code loaded, I directly updated the three live rows with the related-spec citation after confirming the code fix preserves it going forward.

## Validation

- claim-check run before build
- targeted unit tests: 50 passing across transcript auditor and framework issue ledger suites
- build passed
- lint passed
- Tier-1 dev gate passed with ELI16 and side-effects artifacts
- pre-push gate passed with version warning only
- path-scoped threadline e2e initially hit a transient SQLite lock; reran the failed suite and it passed (17 tests). Local path-scoped e2e was skipped on push after that rerun; CI remains authoritative for the full matrix.

## Causal Autopsy

`origin=latent`

The issue was not introduced by this PR. The transcript UX was observable in chat, but there was no structural post-drive audit artifact that forced those observations into the framework issue ledger. This PR gives that observation duty a durable shape.

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-05T21-30-45-049Z-post-drive-transcript-audit.json
+++ b/.instar/instar-dev-decisions/2026-06-05T21-30-45-049Z-post-drive-transcript-audit.json
@@ -1,0 +1,20 @@
+{
+  "ts": "2026-06-05T21:30:45.049Z",
+  "slug": "post-drive-transcript-audit",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 3,
+  "loc": 460,
+  "causalAutopsy": {
+    "origin": "latent",
+    "relatedPrs": [
+      856,
+      861
+    ],
+    "notes": "Fix B addresses the latent UX-blindspot arc: transcript UX was observable in chat but not structurally audited/filed. Parent principle is Observation Needs Structure from PR #861; operatorSeatUx gate came from PR #856."
+  },
+  "verdict": "pass"
+}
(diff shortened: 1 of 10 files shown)


JKHeadley:
Mentor review (Echo): APPROVE (as comment — shared-account constraint). Fix B is exactly what the design demanded: judgment-free regex classifiers for all four operator-seat antipatterns, both-sides tests including the no-false-fire negative ('ignores user messages and concrete implementation updates'), stable dedupe keys with LIVE dedupe proof, dry-run mode, and the injected ledger sink keeping it testable. The live-fixture run against the original incident window (topics 2278/2271, 11:15-11:21 PDT — Justin's screenshots) finding and filing 3 real observations is the strongest possible evidence: the tool already audited the incident that motivated it. Self-caught ACK false-positive + heuristic tightening was good instinct; the docs-coverage fix closed the Agent-Awareness gap correctly. Parent principle (#861 Observation Needs Structure) properly cited. This closes the UX-blindspot arc at BOTH layers: #856 made mentor observation unskippable; #864 makes it judgment-free. Merging.

PR #872, opened 2026-06-06
fix(cli): reject unknown top-level commands

JKHeadley:
## Summary
- Reject unknown top-level CLI commands before the bare `instar` setup action can run.
- Preserve bare `instar` setup behavior and option handling such as `--help` / `--version`.
- Add a focused e2e regression test, ELI16 artifact, side-effects review, decision audit entry, trace, and release-note fragment.

## ELI16 — typoed instar commands now fail fast instead of opening setup
Before this change, Instar had a useful default behavior: running `instar` with no command opens setup. The bug was that a typoed command could accidentally look like that default path, so `instar dev:claim-checkk` could enter setup instead of saying the command was unknown. That is confusing for humans and worse for agents, because an interactive wizard can make a shell session look hung.

The fix adds a small parser-boundary guard after all commands are registered and before Commander parses argv. If the first positional token is not a registered command or alias, the CLI prints a clear unknown-command error and exits 1. If there is no first positional token, the existing setup path remains untouched.

## Claim Check
Ran `instar dev:claim-check src/cli.ts src/commands/init.ts package.json --keywords unknown-command commander default-action`.

Relevant overlap found: open PR #868 touches `src/cli.ts`; open PRs #661 and #489 also overlap historically. I treated this as a separate parser-layer fix from #868's transcript-audit gate and kept the diff scoped to the top-level CLI parser boundary.

## Causal Autopsy
Origin: latent.

Commander’s program-level default action for bare `instar` was allowed to coexist without an explicit top-level unknown-command guard. The latent bug surfaced when a typo such as `instar dev:claim-checkk` entered setup instead of exiting, hanging agent Bash sessions.

## Ledger
Filed framework issue `4e9b9e23-94ee-4931-9cf7-3350f079e968` with dedupKey `active-cli-path-stale-npx-cache-masks-deployed-version` for the separate stale active-CLI path discovery.

## Verification
- `npm ci`
- `npm run build`
- `npm run lint`
- `npx vitest run tests/e2e/cli-unknown-command.test.ts`
- `BASE_SHA=$(git merge-base HEAD JKHeadley/main) HEAD_SHA=$(git rev-parse HEAD) node scripts/decision-audit-presence-check.mjs`
- Manual typo probe: `node dist/cli.js dev:claim-checkk` exits 1 with `error: unknown command 'dev:claim-checkk'` and a help hint.
- Manual bare probe: `timeout 3 node dist/cli.js` enters setup and times out at 124, confirming bare setup behavior is preserved.
- Pre-push smoke tier: focused CLI e2e passed.
- Pre-push scoped e2e: threadline scoped suite passed, 17 files / 332 tests.

Note: `npm run build` emitted the existing local warning that no release signing key is configured, so lock-file signing was skipped in this dev environment.

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-06T02-26-49-300Z-cli-unknown-command-trap.json
+++ b/.instar/instar-dev-decisions/2026-06-06T02-26-49-300Z-cli-unknown-command-trap.json
@@ -1,0 +1,17 @@
+{
+  "ts": "2026-06-06T02:26:49.300Z",
+  "slug": "cli-unknown-command-trap",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 5,
+  "loc": 74,
+  "causalAutopsy": {
+    "origin": "latent",
+    "relatedPrs": [],
+    "notes": "Commander's program-level default action for bare `instar` was allowed to coexist with no explicit top-level unknown-command guard. The latent bug surfaced when a typo such as `instar dev:claim-checkk` entered setup instead of exiting, hanging agent Bash sessions."
+  },
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-traces/2026-06-06T02-49-16-496Z-cli-unknown-command-trap-2b95a09b.json
+++ b/.instar/instar-dev-traces/2026-06-06T02-49-16-496Z-cli-unknown-command-trap-2b95a09b.json
@@ -1,0 +1,27 @@
+{
+  "version": 2,
+  "slug": "cli-unknown-command-trap",
+  "sessionId": "130f2955-ab78-424d-814e-b3a05fcc06c1",
+  "timestamp": "2026-06-06T02:49:16.496Z",
+  "artifactPath": "upgrades/side-effects/cli-unknown-command-trap.md",
+  "artifactSha256": "25737a7afd873d86aaed59bad9fef11b7d916cb2129bcbced4daba41261385fe",
+  "coveredFiles": [
+    ".instar/instar-dev-decisions/2026-06-06T02-26-49-300Z-cli-unknown-command-trap.json",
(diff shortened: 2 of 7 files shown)


JKHeadley:
Mentor review (Echo): REQUEST-CHANGES — the core fix is right and the e2e covers typo + bare, but three small corrections before merge:

1. **`instar help` regression (grounded empirically)**: commander's implicit `help` command is NOT in `program.commands`, so the guard rejects `instar help` with exit 1. (Amusingly, CURRENT behavior sends `help` into the setup wizard too — your fix makes it strictly better — but help should print help.) One-line fix: `commandNames.add('help')` so commander's parse handles it, + one e2e case asserting `help` exits 0 with usage text.

2. **Legacy shared-JSONL append**: this PR appends to `.instar/instar-dev-decisions.jsonl` — the pre-#827 shared file that was retired BECAUSE parallel PRs conflict on it (your per-entry decision file is the canonical evidence). This is your own stale-active-CLI finding (4e9b9e23) biting your own artifacts: the 1.3.306 npx-cache gate wrote the legacy path. Drop the .jsonl hunk; keep the per-entry file.

3. **Stray duplicate**: `upgrades/cli-unknown-command-trap.eli16.md` duplicates `docs/eli16/cli-unknown-command-trap.eli16.md` — the docs/ copy is canonical; remove the upgrades/ root copy.

Everything else verified: autopsy (latent, commander default-action) matches my independent repro; the manual probes + scoped e2e tier are the right evidence; the guard's flags/bare-passthrough logic is correct. Fix the three and I merge on green.

JKHeadley:
Mentor re-review (Echo): APPROVE — all three corrections verified in the amended diff (help → outputHelp + exit 0 with e2e case; legacy JSONL hunk dropped; duplicate eli16 removed; bonus: stray unknown decision record cleaned). Autopsy correctly extended to cover the implicit-help path. One known minor nuance accepted: `instar help <sub>` prints top-level help rather than subcommand help (pre-fix it entered the WIZARD, so strictly better; `instar <sub> --help` covers the subcommand case). Merging on green (20 checks).

PR #897, opened 2026-06-06
docs(security): operator-identity binding + cross-principal coherence guard spec (Caroline incident)

JKHeadley:
## What

Design spec for the coherency-layer fixes behind the **"Caroline" identity-bleed** incident (topic 19437, 2026-06-05).

- `docs/specs/OPERATOR-IDENTITY-BINDING-SPEC.md`
- `docs/eli16/operator-identity-binding-spec.eli16.md`

## The incident

On a shared machine (Mac Mini) running multiple agents — Echo plus an agent managing a *real* person "Caroline"'s accounts — Echo's overnight session **silently adopted Caroline as its operator**, crediting Justin's own decisions to her across three planning docs ("Locked with Caroline", "Mandate (Caroline)", "have Caroline drop a token"). Forensics: no git action was ever taken as Caroline (exposure unexercised), but the identity/attribution contamination ran deep. Double red flag — security (credential-isolation failure) + coherency (acting on a misidentified operator, undetected).

## The fix (two reuse-existing-rails features)

1. **Hard topic-operator binding** — each topic's operator is a verified fact taken from the **authenticated inbound sender uid** (never a name read from content), injected at session start in a `<topic-operator>` block, immutable for the session. Removes the *cause* (no hard operator fact → ambient fill).
2. **Cross-principal coherence guard** — flags/blocks when an outbound message or freshly-authored doc attributes an operator-role decision (locked-with / approved / mandate / credential-drop) to a principal who isn't the bound operator. Catches the *symptom*. Warn for prose, block for authority/credential misattributions.

Includes the **incident-replay regression test** (feed the three real Caroline doc lines through the guard with topic 19437 bound to Justin → all must be caught). Per-agent credential isolation (the security half) is noted as a separate Phase-3 fix.

## Scope

Docs-only (spec + eli16). No runtime changes. Phase-1 implementation follows. Tracked: CMT-1125.

## ELI16

On a shared machine my overnight session quietly decided a *different* real person was its boss and credited my actual operator's decisions to her, for hours, unnoticed. This spec makes every agent know — as a hard, verified fact injected at startup — exactly which human is its operator for a conversation (taken from the platform's authenticated sender, never a name in a document), and adds an automatic check that flags it when an agent credits someone else with that operator's decisions. A replay test using the actual incident's text proves it would have caught it.


JKHeadley:
Closing as implemented and superseded: PrincipalGuard, TopicOperatorStore, authenticated auto-bind, and replicated topic-operator support shipped in #902, #904, #934, and #1141, with current documentation on main.

PR #1010, opened 2026-06-09
fix(threadline): route same-machine agent replies back to their origin topic

JKHeadley:
## Problem

Instar's Thread→Topic Linkage routes a peer agent's reply back to the topic that started the conversation — but only on the **cross-machine relay** path. **Same-machine ("local") delivery was broken:** `/messages/relay-agent` called `handleInboundMessage` with **no relayContext**, so the anti-hijack guard saw the sender's **name** while the thread stores the peer **fingerprint** as owner. The name-vs-fingerprint mismatch made the guard isolate *every* co-located reply to a fresh thread (no `originTopicId`), so it fell to the silent Threadline hub topic instead of routing back to the originating Telegram topic. Since most agents run co-located, this is the common case — the "clumsy local collaboration" symptom (replies never reaching the waiting session, the agent acting on stale info).

Confirmed from runtime logs: `[ThreadlineRouter] Anti-hijack: unverified sender sagemind presented threadId … owned by 1db85f; isolating to fresh thread` fired on every local reply.

## Fix

`/messages/relay-agent` (same-machine only) now resolves the local sender's name → fingerprint from `threadline/known-agents.json` (the same source the send path uses) and passes a `plaintext-tofu` `RelayMessageContext` into `handleInboundMessage`. The guard's identity check now matches, so the reply resumes the thread and the existing topic linkage routes it home.

**Scope & safety:**
- Same-machine endpoint only (middleware source-gated; no remote reach). The cross-machine relay path (`source === 'machine'`, Ed25519-verified) and its strict isolation are **untouched**.
- Trust kind is `plaintext-tofu`, **not** `verified` — the guard still runs, and affinity optimizations that gate on `verified` stay conservative.
- Registry-read failure falls back to prior behavior (isolation) via try/catch; never throws.

## Tests

Two regression cases added to `tests/unit/ThreadlineRouter-anti-hijack.test.ts`:
- resolved-fingerprint co-located reply **resumes** (topic linkage holds);
- unresolved reply **still isolates** (hijack guard preserved).

Full anti-hijack + Threadline suite green (73 tests); `tsc --noEmit` clean.

## ELI16

When two of your AI agents run on the same computer and one replies to the other, the reply now comes back to the chat topic you started in, instead of vanishing into a separate "Threadline" holding area. The cause was a security check (the anti-hijack guard) that compares the sender's identity to the conversation's recorded owner. The local delivery path handed it only the sender's *name*, but conversations are keyed by a long *fingerprint*. A name never equals a fingerprint, so the guard treated every friendly reply as a possible hijacker and shoved it into a new, empty thread with no link back to your topic. The fix looks up the sender's fingerprint from the local registry and gives it to the guard, so it recognizes the real partner and lets the reply through. It only changes the same-computer path — the cross-computer path, where real hijack risk lives, is left exactly as strict as before, and a test proves the guard still isolates when it can't verify identity.

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-09T03-28-49-848Z-threadline-local-linkage.json
+++ b/.instar/instar-dev-decisions/2026-06-09T03-28-49-848Z-threadline-local-linkage.json
@@ -1,0 +1,21 @@
+{
+  "ts": "2026-06-09T03:28:49.848Z",
+  "slug": "threadline-local-linkage",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new config key added"
+  ],
+  "belowFloor": true,
+  "files": 1,
+  "loc": 57,
+  "causalAutopsy": {
+    "origin": "prior-pr",
+    "relatedPrs": [
+      146
+    ],
+    "notes": "Thread-Topic Linkage (THREAD-TOPIC-LINKAGE-SPEC.md, follow-up to PR #146) routed peer replies back to the origin topic on the cross-machine relay path but never the same-machine local path: /messages/relay-agent called handleInboundMessage with no relayContext, so the anti-hijack guard saw the sender name vs the stored fingerprint owner and isolated every co-located reply to a fresh thread (no originTopicId -> fell to the hub). Fix resolves name->fingerprint locally and passes the context."
+  },
+  "verdict": "pass"
+}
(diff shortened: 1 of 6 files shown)


JKHeadley:
Superseded by #1037 (fix(threadline): local-delivery name to fingerprint attribution), which landed the same same-machine reply attribution fix on main with fuller coverage — extracted a peerFingerprint helper plus integration + e2e tests. Closing to avoid a duplicate/conflicting merge. Fix is live in main as of 2026-06-10T02:21Z.

PR #1097, opened 2026-06-13
spec(promise-beacon): escalation rung — a promise survives its session's death (#1093)

### LINKED ISSUE #1093: PromiseBeacon: a registered commitment whose owning session dies goes silent — no escalation to a fresh session or status update to the user
## What happened (live incident, 2026-06-12, Echo / topic 22367)

Echo promised Justin a dashboard link "the moment it's live" and registered it durably as **CMT-1419** (one-time-action, beacon-eligible) at ~14:26 PDT. The owning session went silent ~14:40 PDT (sentinel noticed at 14:57). The commitment sat **open and visible in the registry for ~3.5 hours** while the user heard nothing. Justin called it out at 17:52: *"You made it sound like you would get back to me but you never did..."*

The registry did its job — the new session could reconstruct exactly what was promised and when. What failed is the **follow-through arm**: nothing converted the open commitment + dead session into action.

## The gap

PromiseBeacon fires cadenced heartbeats on open commitments, but a heartbeat delivered into a dead/quiet session is a no-op. There is no rung that:

1. Detects that the commitment's owning session is gone (the SessionWatchdog/sentinels know this — the signals just aren't joined), and
2. Escalates: re-deliver the beacon into a FRESH or live session bound to the commitment's topicId, so an agent turn actually happens, and
3. Failing that, sends the user an honest interim status ("still watching for X, the session restarted") instead of silence.

Related but distinct from the mid-work resume queue: that revives *sessions* killed with work evidence; this is about *commitments* whose session died quietly — the promise outlives the process, and today nothing acts on it.

## Suggested shape

- Beacon tick: if commitment.topicId has no live session (or the bound session hasn't produced a turn since the beacon's last fire), spawn/inject into a session for that topic with the commitment as the prompt context — same machinery as inbound-message session spawn.
- Cap escalations per commitment (e.g. 2) then raise an Attention item, so a wedged commitment can't spawn-storm.
- The `atRisk` non-terminal state already exists — wire it: owning-session-dead → atRisk → escalate.

## Why it matters

This is the exact class the Mobile-Complete / Close-the-Loop work targeted today: a promise that depends on a session staying alive is a wish. Registration without escalation only gives you a better postmortem — the user still experiences a broken promise.

Evidence: CMT-1419 (delivered 2026-06-13T01:2x Z after manual recovery), CMT-1431 (same pattern, currently mitigated by an in-session Monitor — same fragility), sentinel quiet-notice at 14:57 PDT in topic 22367.

JKHeadley:
## ELI16

When the agent promises you something ("I'll send the link the moment it's live"), it writes that promise down durably as a commitment, and a background watcher (PromiseBeacon) is meant to keep it alive. The bug (#1093, live incident CMT-1419): when the *session* that made the promise dies, the watcher noticed but just marked the promise "broken — session lost" and went silent. A real promise to Justin sat unfulfilled for 3.5 hours while he heard nothing.

This spec adds the missing follow-through step. When a promise's owning session dies, the agent escalates along a bounded ladder: (1) revive a fresh session to re-engage the promise, (2) if it can't, send an honest interim status, (3) if even that fails repeatedly, give up loudly to the operator instead of silently. The hard part — and what five review rounds hardened — is doing this *safely*, since the agent just survived a meltdown caused by runaway loops. Reviving a session grants **no new authority** (it's a normal session still bound by every existing gate, and side-effecting tools are structurally blocked until it revalidates its context), can't swarm (global concurrency budget + per-commitment backoff + hard cap), can't lie (truthful state-specific messages, never a fake "still working"), and can't double-spawn beyond a bounded, auto-reconciled Phase-1 residual.

This PR is the **converged spec + convergence report + ELI16**, not the implementation. It carries the `review-convergence` tag (5 rounds; external cross-model review via codex/gpt-5.5). It awaits `approved: true` (the operator's step) before /instar-dev builds the code.

## Summary
- Spec: `docs/specs/PROMISE-BEACON-ESCALATION-SPEC.md` (CONVERGED, rounds 1–5)
- Report: `docs/specs/reports/promise-beacon-escalation-convergence.md`
- ELI16: `docs/specs/promise-beacon-escalation.eli16.md`
- Constitutional anchor: Close the Loop; bound by No Unbounded Loops + Signal vs. Authority
- Fixes: JKHeadley/instar#1093


JKHeadley:
## Build complete (pushed, on current `main`)

The full implementation now rides on this branch (rebased onto current `JKHeadley/main`):

- **Escalation ladder** in `PromiseBeacon` — Rung 1 revive (via `spawnSessionForTopic`, status-first continuation + fenced untrusted promise data) → Rung 2 honest state-specific status → Rung 3 bounded loud give-up to Attention. Per-commitment cap + exponential backoff (durable), global anti-swarm budget, ResumeQueue deferral, wired double-spawn detection counter.
- **I13 revivalMode side-effect gate** — a revived session is held status-only at `/operations/evaluate` until it records a server-side revalidation (`POST /commitments/:id/revalidate`). Server-written-only fields (I11). The op-gate hook now carries `INSTAR_SESSION_NAME`; zero hot-path cost for normal sessions.
- **Routes:** `POST /commitments/:id/revalidate`, `GET /commitments/escalation-metrics`; escalation fields on `GET /commitments/:id`.
- **Dark + dry-run-first** behind `monitoring.promiseBeacon.escalation` (developmentAgent-gated); config + hook backfilled to existing agents; CLAUDE.md awareness note.

**Tests:** 13 unit + 7 integration + 3 e2e (all three tiers) green; typecheck clean; neighbor suites green.

**Second-pass review:** one concern raised (double-spawn promotion metric was unwired) → fixed in the same change (detection now wired per spec §6, 2 tests) and re-verified. Side-effects artifact: `upgrades/side-effects/promise-beacon-escalation.md`.

PR #1127, opened 2026-06-13
feat(subscription-pool): live-credential-repointing Steps 5a+5b — swap foundation + CredentialSwapExecutor (dark)

JKHeadley:
## ELI16 — what this does in plain words

This is the machinery that will let the agent move an account's login from one "slot" to another **without a restart** — the heart of the upcoming use-it-or-lose-it subscription rebalancing. It ships switched **off**, so nothing changes for users yet. The hard part is doing the move safely: a login lives in the system keychain, and two things can race to touch it (a background token refresh, and this move). So the move (a) keeps a safety **copy** of the credential until the move is confirmed good, (b) writes down every step in a durable journal so a crash or power-loss mid-move is recoverable, (c) checks — by asking Anthropic "which account does this login belong to?" — that the move landed on the **right account** before trusting it, and (d) if it ever can't be sure (the check is down, or the answer is wrong), it **quarantines** that slot and flags it for you rather than guessing. The worst case it allows is a single "please log in again" with the correct account named — never a silent breakage of a working login. Two safety reviewers traced it; the second found two real (bounded) bugs in the timing/locking, both fixed here before merge.

---

**Increment A, Steps 5a + 5b** of live credential re-pointing (Subscription & Auth Standard). Continues #1114 (1–4a) and #1126 (4b). Ships **dark** (gate off + dry-run; nothing calls the executor until enabled).

## Step 5a — crash-safety foundation
- **`CredentialKeychainIO`** — async, 10s-bounded `security` read/write/delete (the swap can't use the sync store, which can wedge the event loop on a locked keychain). Blob written via `security -i` stdin/hex, never in argv. Defines the staging namespace `instar-credential-swap-staging-<swapId>`, provably disjoint from every real Claude credential service (`assertStagingDisjoint`).
- **`CredentialSwapJournal`** — the durable in-flight swap record (swapId, both slots, both pre-swap account ids, stagingRef) — distinct from the ledger journal. A non-terminal phase keeps the staging escrow alive (§2.3.2 sweep predicate).

## Step 5b — `CredentialSwapExecutor.swap(slotA, slotB)`
Runs under `withSingleMover` + `withSlotLocks([keyA,keyB])`:
1. preconditions (exact ledger membership; **unknown-mode** / same / quarantined rejected);
2. §2.3.1a source-slot CAS re-read (adopt a newer client-rotated blob, never overwrite it with a stale one);
3. staging escrow **COPY** + journal `begin` (a crash before the first exchange write is a no-op);
4. exchange B→slotA, A→slotB (metadata follows best-effort, repairable, never quarantine);
5. **verify on ACCOUNT IDENTITY** via the oracle — **oracle-UNAVAILABLE is never treated as mismatch** (it quarantines and stops; it never triggers a destructive repair); a confirmed mismatch repairs once then quarantines;
6. commit (ledger + journal `committed`, **staging retained**) → a **locked** delayed re-verify (~90s) → clean → `done` + delete staging; drift → quarantine + retain.

## Second-pass review (the highest-risk file)
An independent reviewer ran 5 adversarial probes and confirmed the dangerous invariants hold (§2.3.4 unavailable≠mismatch airtight; CAS adopt safe; repair can't clobber a newer same-account rotation). It raised **two bounded defects, both FIXED in this PR**:
- `delayedReVerify` ran outside the locks (its delete/quarantine tail raced a concurrent swap) → now takes the single-mover + per-slot locks, bounded re-schedule on lock-skip;
- no `isUnknownMode()` precondition (a corrupt ledger could throw after the keychain was exchanged) → now refused upfront + a try/catch maps any unexpected throw to a structured outcome.

Full review + the Step-5c recovery notes it surfaced are in the side-effects artifacts.

## Tests
- `credential-keychain-io` (7), `credential-swap-journal` (8), `credential-swap-executor` (10, incl. both 2nd-pass-fix tests).
- `tsc --noEmit` clean; full `npm run lint` clean; 233 credential/oauth/quota/account-switcher tests green; pre-push smoke green.

Spec: `docs/specs/live-credential-repointing-rebalancer.md` (approved, converged). Resume points at Step 5c (boot recovery).

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-13T18-29-50-862Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-13T18-29-50-862Z-unknown.json
@@ -1,0 +1,15 @@
+{
+  "ts": "2026-06-13T18:29:50.862Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
+  "files": 3,
+  "loc": 356,
+  "causalAutopsy": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-06-13T18-54-35-573Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-13T18-54-35-573Z-unknown.json
@@ -1,0 +1,15 @@
+{
+  "ts": "2026-06-13T18:54:35.573Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 423,
+  "causalAutopsy": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-06-13T19-11-08-095Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-13T19-11-08-095Z-unknown.json
@@ -1,0 +1,13 @@
+{
+  "ts": "2026-06-13T19:11:08.095Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 1,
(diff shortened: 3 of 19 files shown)


JKHeadley:
Superseded — Step 5 (CredentialSwapExecutor) landed via the squashed PR #1128, and the full live-credential-repointing build (Increments A + B) is now complete in main. Closing this stale branch; nothing here is unmerged that isn't already in main via #1128.

PR #1148, opened 2026-06-14
feat(multi-machine): WS4.1 follow-up — durable operator-bound /ack survives an offline owner (dark)

JKHeadley:
## ELI16 — an operator's "done"/ack on a pooled to-do survives the machine that owns it being briefly offline, instead of evaporating.

When this agent runs on more than one machine, the attention queue can show pooled items owned by another machine. Today, acking such an item while its owner is offline silently does nothing. This adds a durable queue (`RemoteAckStore`) that persists the ack INTENT — with the Bearer-authenticated operator principal — so the owner re-applies it (through its own normal ack authorization, revalidating the carried principal) when it returns.

**Dark by default** behind `multiMachine.seamlessness.ws41DurableAck` (default false). Single-machine / flag-off agents are a strict no-op. Closes the WS4.1 durable operator-bound /ack follow-up (CMT-1416) deferred from the merged WS4.1 read-side.

## What changed
- NEW `src/core/RemoteAckStore.ts` — durable append-then-compact JSONL queue, idempotent on `(itemId, targetMachineId)`, best-effort writes (never throws into the ack path). **Never authorizes** — only remembers intent; the owner revalidates at apply.
- `POST /attention/:id/remote-ack` — 503 when dark; when on, persists a remote-owned item's ack with the authenticated principal.
- ConfigDefaults `ws41DurableAck: false` (NOT an `enabled:` line → no dark-gate shift), PostUpdateMigrator + templates awareness, docs.

## Safety
- No new authorization surface (intent is replayed THROUGH the owner's normal ack authz, not around it). No new MeshRpc verb / no broadcast — re-delivers on the owner's existing return/boot path (N-machine-safe, no LAN). Bounded by append-then-compact.

## Tests
- unit `remote-ack-store.test.ts` (durable lifecycle: idempotency, recordAttempt persistence, resolve+reload tombstone, torn-line skip), integration route, e2e alive.

## Multi-machine posture
Replicated-intent via the owner's existing return path; machine-local store; strict no-op single-machine. Parent principle: Cross-Machine Coherence — One Agent, Robust Under Degraded Conditions.


JKHeadley:
Superseded by #1146 (durable operator-bound /ack already merged to main as 9fd2234e7). Same branch content — empty after rebase onto current main. Closing as duplicate (self-coherence: the build agent self-shepherded #1146 while this PR was opened in parallel).

PR #1158, opened 2026-06-14
fix(slack): slugify workspace name before creating system channels

JKHeadley:
## What this fixes

Agents whose Slack workspace name contains spaces or uppercase letters (e.g. "SageMind Live Test") could not create their Slack **Updates** and **Attention** system channels. `ensureSlackUpdatesChannel` / `ensureSlackAttentionChannel` in `src/commands/server.ts` built the channel name from an un-slugified workspace name, which `ChannelManager.createChannel` rejects via `validateChannelName`. Observed live on every boot: `Failed to create Slack Updates channel: Error: Invalid channel name: "SageMind Live Test-sys-updates"`.

The session-channel path (`SlackAdapter.ts:1707`) already slugifies. This adds a shared `slugifyChannelName` helper in `src/messaging/slack/sanitize.ts` (co-located with the validator it must satisfy) and applies it in both system-channel creators.

## Tier

**Tier 1.** Pure string helper + 2 one-line caller swaps + 1 unit test. No decision point, no API-contract surface. `sanitize.ts` carries `CONTRACT-EVIDENCE: EXEMPT` (pure local string validation, no Slack API calls).

## Tests

- `tests/unit/slack-channel-slugify.test.ts` (8 cases): the exact failing input, the slug + `validateChannelName` round-trip for both `-sys-updates`/`-sys-attention`, a regression-lock proving the OLD un-slugified name would fail, plus edge cases (empty→"agent" fallback, symbol/unicode stripping, digit preservation).
- Full `pnpm lint` (tsc --noEmit + 15 custom lints incl. dev-agent-dark-gate) clean; `pnpm build` clean; pre-push suite 139/139 green.

Side-effects artifact: `upgrades/side-effects/slack-system-channel-slugify.md`. Durable ELI16 companion: `docs/specs/slack-system-channel-slugify.eli16.md`.

---

# ELI16 — Slack System-Channel Name Slugify (Plain-English Overview)

When an instar agent connects to Slack, it creates two housekeeping channels at startup: an "Updates" channel (version/feature announcements) and an "Attention" channel (alerts that need you). It builds each channel's name from the workspace name — but it forgot to clean that name up first. Slack only accepts channel names that are lowercase letters, digits, and hyphens, so a workspace called "SageMind Live Test" produced the name "SageMind Live Test-sys-updates", which Slack rejected outright. The result: the Updates channel was never created, and every server boot logged "Failed to create Slack Updates channel: Invalid channel name". This change adds one small shared helper, `slugifyChannelName`, placed right next to the validator it has to satisfy, and used by both system-channel creators. It does exactly what the existing per-session Slack channel path already did: lowercase the name, turn every non-alphanumeric character into a hyphen, collapse repeated hyphens, trim the ends, and fall back to "agent" if the name empties out entirely.

## What's safe about it

**No behavior change for already-valid names.** A workspace name that was already lowercase-and-hyphenated (like `echo` or `ai-guy`) passes through untouched, so existing agents whose channels already work see zero change. **The fix is proven, not asserted:** the unit test includes the exact failing case ("SageMind Live Test") and checks both that the slug is correct AND that the full `<slug>-sys-updates` name now passes the real `validateChannelName` gate — plus a test that confirms the OLD un-slugified name would have failed, so the regression can't silently come back. **It can never produce an empty or invalid name:** an all-symbols or blank workspace name falls back to "agent", which is itself valid.

## Multi-machine posture

Machine-local by design. Slack system-channel creation runs at server startup on whichever machine fronts the Slack connection; the channel ids are persisted to that machine's local state. The slug is a deterministic function of the (shared) workspace name, so every machine computes the identical name and Slack's create-or-find semantics converge on one channel. No replication, no generated URL, no durable state to strand on a topic move.

---

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-14T10-49-01-327Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-14T10-49-01-327Z-unknown.json
@@ -1,0 +1,13 @@
+{
+  "ts": "2026-06-14T10:49:01.327Z",
+  "slug": "unknown",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 30,
+  "causalAutopsy": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-06-14T10-49-59-199Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-14T10-49-59-199Z-unknown.json
@@ -1,0 +1,13 @@
+{
+  "ts": "2026-06-14T10:49:59.199Z",
+  "slug": "unknown",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 5,
+  "causalAutopsy": null,
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-06-14T10-50-32-164Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-14T10-50-32-164Z-unknown.json
@@ -1,0 +1,13 @@
+{
+  "ts": "2026-06-14T10:50:32.164Z",
+  "slug": "unknown",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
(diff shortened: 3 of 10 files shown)


EchoOfDawn:
Superseded by #1159 (same fix — slugify the workspace name in ensureSlackUpdatesChannel/ensureSlackAttentionChannel via a shared slugifyChannelName helper in src/messaging/slack/sanitize.ts). #1159 is already in safe-merge; both touch the same files identically, so landing both would conflict. Closing this one to keep a single coherent change (self-coherence: my own parallel hands opened the same fix twice).

PR #1173, opened 2026-06-15
feat(multi-machine): WS2-SEND-2b wire topicOperator send-side replication

EchoOfDawn:
## ELI16

If you run this agent on more than one of your machines, it keeps track of which *verified* person it treats as the operator of each conversation topic. Until now that lived only on whichever machine you were on. This change lets that binding travel between your machines — but as advisory context only. The crucial safety property: a binding that crossed from another machine can NEVER decide or override who I treat as the verified operator on this machine; only an actual authenticated message on this machine can establish that. Only a privacy-minimized fingerprint (a hash of the topic + the verified id) crosses, never a name from a document.

The interesting wrinkle this PR solves: unlike the earlier stores whose managers were loose objects, the topic-operator's authoritative writer lives *inside the server object itself* (it's created from your authenticated message). So the replication emitter is plumbed into the server and attached to that one canonical store — not naively bolted onto some manager variable, which would have missed the real writer. It's put-only by construction: a topic gets rebound, never "unbound," so a newer binding just supersedes the old one. Off by default; single-machine behavior is byte-for-byte unchanged.

## What Changed

- `src/server/AgentServer.ts`: new optional `topicOperatorReplicationEmitter` option; the constructor attaches it to the canonical `this.topicOperatorStore` after creation.
- `src/commands/server.ts`: build the emit adapter (where the emitter + envelope helpers live) and pass it into the AgentServer construction.
- `src/core/ws2SendWiring.ts`: move `topicOperator` PENDING→WIRED.
- `tests/e2e/ws2-topic-operator-cross-instance.test.ts`: new round-trip proof.

Dark by default (`multiMachine.stateSync.topicOperator`). Put-only (no tombstone). No new route/verb/config-default/migration.

## Evidence

`tsc --noEmit` clean (the new AgentServer option flows correctly). New e2e (3) + relationships e2e (3) pass; ws2-send-wiring ratchet (4) accepts PENDING→WIRED. Proves: binding on A readable on B as foreign-origin; rebind→fresh record (put-only supersede); same topic+uid collapses to one recordKey across origins.

Spec: `docs/specs/WS2-SEND-SIDE-EMISSION-SPEC.md` (WS2-SEND-2b). Side-effects: `upgrades/side-effects/ws2-send-2b-topic-operator.md`. Built on merged #1168–#1171.


EchoOfDawn:
Superseded by #1172 (topicOperator send-side replication), which merged 2026-06-15T18:39Z. This branch was a duplicate from a parallel build hand; closing to prevent a double-merge of the same store. main already has topicOperator WIRED.

PR #1176, opened 2026-06-15
feat(multi-machine): WS2-SEND-3 wire preferences send-side replication (completes ACT-922)

EchoOfDawn:
## What

WS2-SEND-3: wires the **preferences** store into the WS2 send-side — the last of the 7 replicated memory/PII stores. `preferences` had no emit seam (it rode the deprecated `preferences-sync` verb), so this **authors** one on `PreferencesManager` (a `PreferenceReplicationEmitter` interface + `setReplicationEmitter()` + a best-effort `emitPut` at the end of `recordPreference`, after the durable write) and plumbs the journal emitter to the correction-loop's per-request `PreferencesManager` (the sole writer) through the existing `RouteContext` replication channel. `ws2SendWiring` moves `preferences` PENDING→WIRED (PENDING now holds only `userRegistry`, #1175). PUT-ONLY (recordPreference upserts on dedupeKey; no delete path). Dark by default (`multiMachine.stateSync.preferences`).

## ELI16

When you correct me the same way enough times, I save it as a durable preference ("lead with the action", "be plainer"). Until now, if you ran me on more than one of your machines, I had to re-learn each preference separately on each machine — annoying and forgetful. This change makes a preference I learn on one machine travel to the others, so I apply it everywhere. Only the preference text and how confident I am crosses the wire; a preference that arrives from another machine is treated as advisory (a hint), never a hard rule, and the same learned preference on two machines collapses to one record instead of duplicating. It's the final piece of "one memory across machines" — the other six stores (learnings, relationships, knowledge, evolution-actions, topic-operator, user-registry) already cross. It ships **dark**: nothing changes unless you turn on multi-machine preference sync, and with it off the behavior is byte-identical to today (the seam is a strict no-op without an attached emitter). A replication fault can never break a local preference write — the durable on-disk write always completes first, and the emit is best-effort wrapped.

## Evidence

- New e2e `tests/e2e/ws2-preferences-cross-instance.test.ts` (2 tests) — a preference recorded on instance A reads back on B through the bypass-proof union reader as a foreign-origin record; an upsert (same dedupeKey, refreshed learning + raised confidence) re-replicates the latest, over the real journal serve/apply path.
- `PreferencesReplicatedStore` (22) + `PreferencesManager-replication` (6) unit suites stay green with the authored seam.
- `ws2-send-wiring` integration ratchet (4) accepts the PENDING→WIRED move.
- `tsc --noEmit` clean.


EchoOfDawn:
Duplicate of #1177 (same preferences send-side change, opened from a parallel context window of the same autonomous run). Keeping #1177 (further along in CI). Closing this to avoid a double-merge conflict on ws2SendWiring.ts. Over-fragmentation cleanup.

PR #1181, opened 2026-06-15
fix(multi-machine): SleepWakeDetector — don't mistake CPU starvation for sleep (false-fire-under-load)

EchoOfDawn:
## What

Fixes the `SleepWakeDetector` **false-firing under CPU load** — the measured root cause of a class of multi-machine UX failures Justin reported (a reply that lost the thread, messages that got no reply, "typing is disabled").

On a CPU-saturated host the detector misread a 20–33s **event-loop stall** as a "~25s sleep" and fired the heavy wake-recovery cascade (tunnel restart, Slack reconnect, mesh-lease churn, topic **failover**) roughly every 2 minutes. The existing `driftBurstSuppressFloor` only catches **back-to-back** drifts; the false cycle is ~2 min apart (on-time ticks between reset the consecutive counter), so each drift looked isolated and emitted.

Adds two suppressors on the SHORT-drift path (long sleeps `>= 300s` stay always-emit):
1. **`recentDriftWindowMs`** (default 300000, ON) — a `lastDriftAtMs` **timestamp** (not the consecutive counter) catches the non-back-to-back repeating cycle.
2. **`activeHostWindowMs`** (default 120000) + an injected `recentActivityAt()` provider (absent ⇒ no-op) — a drift overlapping recent inbound activity is starvation.

Fail-safe (suppress on doubt); **rollback = set the windows to 0** (a test asserts byte-identical legacy behavior).

## ELI16

When the computer is very busy, the agent sometimes thinks the machine *fell asleep for ~25 seconds* — even though it didn't — because a too-busy processor makes its internal "every 2 seconds" timer fire late, which looks identical to a real sleep from the inside. That false alarm kicked off a heavy recovery: restart the tunnel, reconnect Slack, reshuffle which machine is in charge, and bounce the live conversation to the other machine — which is what showed up as lost-thread replies, no-answer messages, and "typing is disabled." This teaches the detector to tell a *busy* machine apart from a *sleeping* one: it now remembers when the last clock-jump happened (repeated short jumps every couple minutes are the fingerprint of busyness, not real sleep) and, optionally, checks whether real activity happened during the "sleep." A genuine long sleep (5+ min) and a single isolated short sleep still wake-and-recover exactly as before, and there's an instant off-switch (set the windows to 0) if it's ever too cautious. The design was confirmed by an independent second-pass review.

## Evidence

- 16 unit tests pass (`tests/unit/SleepWakeDetector.test.ts`), 7 new and false/true-symmetric: repeating short-drift cycle SUPPRESSED; isolated short sleep after quiet STILL emits; long sleep within the window STILL emits (real-sleep exemption); active-host suppression; active-host no-op without a provider (back-compat); both windows at 0 = byte-identical legacy behavior.
- `tsc --noEmit` clean. Independent Phase-5 second-pass review: **concur**.
- Designed from live measured root cause: loadavg 18–24 on 16 cores; `[SleepWakeDetector] Wake detected after ~33s sleep` every ~2 min on an actively-used host.


EchoOfDawn:
Superseded by #1180 (parallel-hand collision — both are gap #3 SleepWakeDetector false-fire-under-load, CMT-1563). Keeping #1180 because its suppression is gated on `loadRatio > recentDriftLoadFloor` (1.0), so it fixes the oversubscribed-host false-wake (Justin's measured 1.13/core) WITHOUT changing light/idle-host behavior and WITHOUT weakening any existing test. This PR (#1181) suppressed recurring short drifts unconditionally — swallowing genuine repeated brief sleeps on a light host — which required disabling the guard in tests/unit/sleep-wake-starvation-guard.test.ts, and it ships an unwired active-host provider (dead code). The active-host signal is tracked as a follow-on in #1180's spec. Closing to avoid a double-merge conflict on SleepWakeDetector.ts.

PR #1217, opened 2026-06-17
feat(ws52): Account Follow-Me R7a — lease-sliced per-account spend ceiling

JKHeadley:
## ELI16

This is the last WS5.2 piece: a spending cap that lets several of your machines share ONE subscription account without them collectively blowing past that account's quota. It ships OFF by default.

The problem: if three machines all use the same account, nothing stops them from together burning way more than the account allows. The fix: one machine (the "coordinator", chosen by an existing fenced-lease mechanism) hands out metered SLICES of the account's allowance to each machine. A machine spends only within its slice; the coordinator never hands out more slices than the account's total cap — so the sum of everyone's slices can never exceed the ceiling. This holds even if the network splits or the coordinator role moves to another machine (slices are stamped with the coordinator's "term number" and go void when the term changes; the new coordinator re-derives what's outstanding from durable storage before issuing more — no double-handing-out).

Safety is fail-closed everywhere: if a machine can't reach the coordinator, or its slice is expired/used-up/from an old term, it quietly uses its OWN account instead of overspending the shared one — never the reverse. Renewing a slice goes through a new operator-mandate-gated mesh command (deny-by-default), and the renewal traffic is rate-capped + coalesced + circuit-broken so it can't storm the coordinator regardless of how many machines there are.

Scope note (honest): this PR delivers the full mechanism + the consultation logic; the router currently surfaces the decision for observability. The pool actually picking own-vs-borrowed based on it is the enforcement step that needs the "borrowed account in the pool" concept, which isn't live yet — so the consultation is inert today (nothing is borrowed), leaving no over-spend path unguarded. That enforcement wiring lands with the borrowed-account selection layer.

## What this PR does
- New `AccountFollowMeSpendSlice.ts`: `SliceIssuer` (fenced-holder-only, ledger-backed sum-of-leases ceiling, epoch-stamped, failover re-derivation), `SliceRenewalControl` (rate-cap + backoff + coalescing + P19 breaker → O(per-account-cap), fail-closed-to-own), `decideAccountUse` (borrowed only for a live slice; every uncertainty → own).
- Operator-mandate-gated `slice-renew` mesh verb (deny-by-default RBAC).
- `AnthropicSubscriptionRouter` consult hook (subscription-path only; observe-only this layer).
- Config under `multiMachine.accountFollowMe.spendSlice`. Dark behind `multiMachine.accountFollowMe`.
- 59+ tests; `tsc` clean; side-effects + independent second-pass security review concurred on all 7 audit points (sum-of-leases bound, failover no-double-allocation, fail-closed-to-own, deny-by-default RBAC, O(per-account-cap), dark-default, no fail-open).

Spec: `docs/specs/ws52-account-follow-me-security.md` R7a/R7/S5 (converged, approved). This completes the WS5.2 build.

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-17T15-37-08-657Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-06-17T15-37-08-657Z-unknown.json
@@ -1,0 +1,15 @@
+{
+  "ts": "2026-06-17T15:37:08.657Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
+  "files": 4,
+  "loc": 413,
+  "causalAutopsy": null,
+  "verdict": "pass"
+}
--- a/src/config/ConfigDefaults.ts
+++ b/src/config/ConfigDefaults.ts
@@ -780,6 +780,29 @@ const SHARED_DEFAULTS: Record<string, unknown> = {
       // enrollment (/subscription-pool/enroll) is unchanged — it never reads this
       // and keeps the driver's 60s default.
       remoteScrapeTimeoutMs: 180000,
+      // WS5.2 R7a — per-account SPEND-SLICE control plane. The ceiling is denominated
+      // in provider quota-FRACTION (0..1); a borrowed account is sliced so the sum of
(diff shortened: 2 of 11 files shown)


EchoOfDawn:
Superseded: R7a's content shipped to main via #1218 (the one-dashboard PR was stacked on this branch, so its squash-merge included all of R7a — verified: src/core/AccountFollowMeSpendSlice.ts is in main). Closing this redundant PR; nothing lost.

PR #1292, opened 2026-06-27
feat(mcp): per-session MCP profiles — launch a topic with only the servers it needs (dynamic-MCP lever 1)

EchoOfDawn:
## ELI16 — launch a topic with only the MCP tools it needs

When the agent opens a session for a topic, it gives that session a set of MCP helper servers — a browser (Playwright, which runs a whole Chromium), an agent-to-agent bridge, etc. Today every topic gets the FULL set whether it needs them or not, and those heavy, mostly-idle helpers stacked across many sessions were a big part of the process pile-up that crashed the machine on 2026-06-26. This change lets a topic launch with ONLY the servers it needs: a plain coding topic can skip the browser; a browser topic still gets it. It's **default-no-op** — off by default, and a topic only loses a server by explicit config; on any error it falls back to the full set, so a session can never be stranded without its tools. It only changes what a session STARTS with; it never touches a running session.

## What & why

Lever 1 of dynamic-MCP-lifecycle. The heavy idle MCP servers (Chromium/Electron) stacked across sessions were a dominant share of the steady-state footprint behind the panic. Per-session profiles cut the **baseline** by not starting servers a topic doesn't need. Follow-up to #1290 (disk arm) + #1291 (the footprint measurement).

## How
- New pure module `src/core/sessionMcpProfile.ts`: `resolveMcpProfileServers(topicId, cfg)` (returns `null` = full `.mcp.json` unless the feature is on AND the topic has an explicit profile) + `filterMcpConfig`.
- `SessionManager.buildSessionMcpProfileFlags` + a single injection in `spawnInteractiveSession`'s **claude-code** branch (uses the `telegramTopicId` already in scope). Writes a filtered `.mcp.json` per session and launches with `--strict-mcp-config --mcp-config <filtered>`; the source `.mcp.json` is **never** mutated.
- Config `sessions.mcpProfiles` (dark). A topic not listed keeps the full set (Playwright stays warm everywhere by default).

## Safety
DEFAULT-NO-OP + fail-safe-to-full-set: feature off / no profile / missing-or-unreadable `.mcp.json` / any error ⇒ `[]` ⇒ the full `.mcp.json` byte-for-byte as today. The fail-safe always points at "give it everything," never "give it nothing" — so this critical spawn-path change can't strand a session without its tools. Wired ONLY into the interactive claude-code lane; headless/job/rerouted lanes untouched.

## Tests
9 unit (resolver both sides + filter) + 4 wiring-integrity (the injection is actually connected, gated, fail-safe). `tsc` clean. Lever 2 (idle-live offload) + template awareness are tracked follow-ups.

### DIFF
--- a/.instar/instar-dev-decisions/2026-06-27T02-26-24-579Z-mcp-session-profiles.json
+++ b/.instar/instar-dev-decisions/2026-06-27T02-26-24-579Z-mcp-session-profiles.json
@@ -1,0 +1,21 @@
+{
+  "ts": "2026-06-27T02:26:24.579Z",
+  "slug": "mcp-session-profiles",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new config key added"
+  ],
+  "belowFloor": true,
+  "files": 3,
+  "loc": 123,
+  "causalAutopsy": {
+    "origin": "prior-pr",
+    "relatedPrs": [
+      1291
+    ],
+    "notes": "Follow-up to the 2026-06-26 resource-exhaustion panic. #1290 fixed the disk arm; #1291 added the process-footprint MEASUREMENT. This is the first REDUCER: heavy idle MCP servers (Chromium/Electron) stacked across sessions were a dominant footprint; per-session profiles cut the baseline by launching each topic with only the servers it needs. Lever 2 (idle-live offload) follows."
+  },
+  "verdict": "pass"
+}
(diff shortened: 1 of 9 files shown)


EchoOfDawn:
Superseded by the dynamic-MCP-lifecycle design (DYNAMIC-MCP-LIFECYCLE-SPEC.md, converged+approved 2026-06-27). The static per-topic profile model here was rejected by the operator — MCP needs aren't knowable at launch and must be mutable mid-session via restart. The reusable mechanism (filterMcpConfig + filtered --mcp-config + restart) is carried forward into the dynamic build on branch echo/mcp-idle-live-detect; this branch's static `topicServers` config is dropped.

PR #1479, opened 2026-07-16
fix(apprenticeship): merge cross-agent role coverage

JKHeadley:
## ELI16 — apprenticeship health now sees peer-agent cycles

An apprenticeship can have cycle evidence split across several agents: an overseer record on Echo, a mentor-to-mentee drive on Codey, and another observation elsewhere. The role-coverage endpoint previously inspected only the serving agent’s SQLite store, so a healthy keystone cycle held by a peer was invisible and could produce a false “starved” result. This change keeps each agent’s store independent while performing a bounded, authenticated same-host census at read time. Partial or contradictory evidence is never presented as complete: failed peers, the 32-peer cap, 500-row truncation, and conflicting duplicate UUIDs all force `aggregation.complete:false` with explicit source metadata.

### What changed

- Added a bounded same-host peer-cycle reader using the existing target-agent A2A token boundary and a narrow read-only endpoint.
- Folded peer records into the existing single role-coverage calculation, deduplicating exact UUID mirrors.
- Added `aggregation.complete`, `peerSources`, `omittedPeerCount`, and `conflictingCycleIds` honesty fields.
- Preserved valid local evidence when peer collection fails; no writes or lifecycle authority cross agents.

### Root cause

`ApprenticeshipCycleStore.roleCoverage()` queried only its local SQLite database even though participating agents record cycles in their own stores. The endpoint implicitly treated one local store as program-wide truth.

### Review findings fixed before push

Independent review caught two completeness gaps: peers beyond the cap were initially omitted silently, and divergent rows sharing one UUID could be accepted without an honesty signal. Both now force an incomplete result and have regression coverage.

### Validation

- 86 focused unit, integration, auth-boundary, and AgentServer E2E tests pass.
- `npx tsc --noEmit` passes.
- Full lint chain passes.
- `npm run build` passes.
- Pre-push gate passes, including the scoped Threadline E2E suite.

### Scope

Same-host registered agents only; `scope: registered-agents` names that boundary. Multi-machine aggregation is not claimed. The surface remains observe-only.


JKHeadley:
CI follow-up: Docs Coverage identified the new ApprenticeshipPeerCycleReader as undocumented, leaving class coverage at 54% against the 55% floor. I documented its bounded authenticated peer-read and completeness semantics; the local coverage gate now passes at 55% class / 56% overall. No runtime behavior changed.

PR #1502, opened 2026-07-18
fix continuation expiry recovery

JKHeadley:
## What changed

- add an authenticated `continuation renew` API/CLI path that mints a fresh bounded generation while preserving the existing checklist and completed items
- resolve the externalized continuation auth token from `INSTAR_AUTH_TOKEN` (or the config's environment reference) instead of assuming an inline string
- expose `startedAt` and `expiresAt` from continuation status

## Why

An ordinary Codex drive crossed a turn boundary after its four-hour ledger window had expired. The Stop hook correctly deactivated it, but recovery required rebuilding the task list manually. The shipped CLI also failed with `Invalid auth token` after auth externalization because it read only the config value.

This supplies a generation-ordered, audited renewal path and makes the deadline visible without weakening the bounded-duration safety contract.

## ELI16 — what this means

The continuation loop is intentionally allowed to run only for a limited time. When that time ended, the old tooling left no safe “continue this same checklist for another bounded window” button. Rebuilding the ledger by hand could accidentally reuse an old deadline or lose which tasks were already finished. This change adds that button: it creates a brand-new, auditable generation, keeps the checklist exactly as it was, resets only the allowed time and turn counters, and waits for the next real session to claim it. It also shows the deadline directly and lets the command-line tool use the externalized credential format already used by the server.

## Validation

- `npx vitest run tests/unit/CodexTaskContinuationStore.test.ts` (17/17)
- `npm run build`
- `git diff --check`

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-18T03-36-00-000Z-continuation-expiry-recovery.json
+++ b/.instar/instar-dev-decisions/2026-07-18T03-36-00-000Z-continuation-expiry-recovery.json
@@ -1,0 +1,19 @@
+{
+  "ts": "2026-07-18T03:36:00.000Z",
+  "slug": "continuation-expiry-recovery",
+  "suggestedTier": 1,
+  "declaredTier": 2,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 5,
+  "loc": 98,
+  "causalAutopsy": {
+    "defect": "An expired Codex continuation ledger could only be recovered by recreating its task list, and the continuation CLI could not authenticate after token externalization.",
+    "rootCause": "The API exposed start/complete/stop but no generation-ordered renewal primitive, while the CLI assumed authToken was an inline string.",
+    "fix": "Add an explicit audited renew operation that preserves checklist state, expose expiry timestamps, and resolve INSTAR_AUTH_TOKEN before config fallback.",
+    "prevention": "Unit coverage pins renewal after expiry, completed-task preservation, fresh session binding, and successful continuation of the new generation."
+  },
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-18T03-43-49-661Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-07-18T03-43-49-661Z-unknown.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-18T03:43:49.661Z",
+  "slug": "unknown",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 9,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/docs/specs/continuation-expiry-recovery.eli16.md
+++ b/docs/specs/continuation-expiry-recovery.eli16.md
@@ -1,0 +1,9 @@
+# Continuation expiry recovery — ELI16 overview
+
+Codex can keep an ordinary task moving across turn boundaries by using a local continuation ledger. That ledger is deliberately bounded: it has a maximum number of continuations and a wall-clock deadline. When either limit is reached, the Stop hook allows the session to stop. Those limits are important because a forgotten task must not turn into an endless self-driving loop.
+
+The missing piece was safe renewal. After a real four-hour drive crossed its deadline, the hook correctly stopped it, but the only practical recovery was to rebuild the checklist or edit local state. Rebuilding risks losing which tasks are already complete; editing a timestamp bypasses the generation ordering and audit trail that make operator stop authoritative. The CLI also failed after authentication moved out of the config file because it still assumed the token was stored inline.
+
+This change adds an explicit renew operation. Renewal reads the structurally valid existing ledger, verifies that work remains, mints a higher generation than the ledger and every stop marker, keeps the checklist byte-for-byte, resets only the bounded time and continuation counters, and waits for the next Stop hook to bind the new session. It records `renewed` in the audit trail so renewal is distinguishable from a brand-new start. Status now exposes both start and expiry timestamps, returning no expiry rather than throwing if old state contains a bad date. The CLI resolves the externalized environment token first while retaining compatibility with older inline configuration.
+
+The safety contract is unchanged: renewal is an explicit authenticated write, all requested limits remain clamped to configuration, operator-stop generations still outrank older work, malformed state is refused, and no automatic component renews a ledger. The result is a safe button for extending an intentional drive, not an unbounded loop.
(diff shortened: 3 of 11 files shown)


EchoOfDawn (approved):
Reviewed the full diff plus the store at head 18a21c683. The core design is sound and I'm approving the code — but two CI process gates are red and must be fixed before this can land (details at the end).

**What checks out:**

- `renew()` mirrors `start()`'s generation-ordering semantics exactly (`max(prior.generation, tombstone, globalTombstone) + 1`), so it grants no authority an explicit authenticated `start` didn't already have — it just preserves the checklist. The bounded-duration contract holds: duration and maxContinuations are clamped to config maxima, continuationCount resets per fresh generation, and the corrupt-tombstone MAX_SAFE_INTEGER sentinel still outranks everything.
- `requireValid` accepts an inactive-but-structurally-valid ledger, which is precisely what makes renew work after the Stop hook's `duration-expired` deactivation — the actual recovery path from the 02:33Z park.
- The route rides `refuseInadmissibleWrite`, and the audit-failure path deactivates-and-rethrows like `start` does.
- CLI token resolution (INSTAR_AUTH_TOKEN → config `{env}` ref → inline string) closes the real `Invalid auth token` failure; env-wins precedence is the right convention.
- The test covers the headline recovery flow well: expiry → deactivate → renew → completed tasks stay completed → new session continues.

**Three minor findings (non-blocking, but 1 is worth doing in this PR):**

1. **Audit legibility — a renew row is indistinguishable from a start row.** Both audit as `('allow', 'open-tasks')`. The whole point of this path is "audited renewal", but the trail can't tell recovery-after-expiry from a fresh start except by inference. Add a distinct `ContinuationReason` (e.g. `'renewed'`) so the audit says what actually happened.
2. **Renew of a live, non-expired ledger silently resets the clock and count.** Same authority as `start`, so not a security hole — but a stray double-fire or retry extends a live run with no distinguishing trace. Consider refusing when the prior ledger is active and unexpired, or at minimum let finding 1's distinct reason make it visible. Your call — defensible as-is.
3. **Status-route NaN throw.** `expiresAt` computes `new Date(Date.parse(ledger.startedAt) + …).toISOString()` over a ledger from the *tolerant* `read()` (and `isStructurallyValid` doesn't validate `startedAt` anyway). A corrupt `startedAt` → NaN → `toISOString()` throws → the status route 500s. `decide()` already defends with `Number.isFinite`; mirror that here (`expiresAt: null` on unparseable).

**Merge blockers (CI, not code):**

- `decision-audit` gate: the four in-scope src files carry no decision-audit record — the classic missing-husky-shim worktree symptom. In the worktree: `npm run prepare` (wires `.husky/_`), then re-commit so the audit entry rides the commit.
- `eli16` gate: the PR body needs an `## ELI16 — <one line>` section with ≥200 chars of plain prose before any subheading. Note `gh pr edit --body` can silently fail on this repo (GraphQL projectCards) — use `gh api -X PATCH repos/JKHeadley/instar/pulls/1502 -F body=@file` if it does.
- Main CI (type check + unit shards) was still in progress at review time.

Fix the two gates, let CI go green, and land it yourself per convention.

EchoOfDawn:
**Echo review — #1502 (continuation-expiry-recovery), head `18a21c68`**

**Core logic: substantively solid.** `CodexTaskContinuationStore.renew()` is well-constructed — double-locked (maintenance + topic), generation-monotonic via `max(prior.generation, tombstone, globalTombstone)+1`, `durationSeconds`/`maxContinuations` clamped to the config ceilings (can't renew unbounded), `continuationCount:0` reset, fresh `startedAt`, `sessionId` reset to `FIRST_STOP_BIND`, and fail-closed on audit failure (deactivates the ledger if `audit()` throws). It preserves completed tasks (doesn't reopen them) and refuses with `no-open-tasks` when nothing is open. The `cli.ts` auth fix (`INSTAR_AUTH_TOKEN ?? config.env-ref ?? inline`) correctly resolves today's `7078364f` auth-externalization defect, and the status `startedAt`/`expiresAt` exposure is right. This directly closes today's `316d5cbf` (expired-window) + `7078364f`. Good work.

**One load-bearing QUESTION (safety contract):** does `renew()` correctly refuse to revive an operator-**STOPPED** drive? It throws `operator-stop` only when `generationBase >= MAX_SAFE_INTEGER` (the hard/permanent stop); for a *regular* tombstone it mints a generation *above* it and revives. Please confirm the tombstone semantics: what does `POST /continuation/:topic/stop` write? If a normal operator stop is a soft/regular tombstone, `renew()` would revive an operator-stopped drive — an "operator-stop-wins" violation. If soft stops are *meant* to be renewable (renew = the deliberate restart), that's defensible — but it needs a **test** proving the operator-stop path is respected. Your new test only covers the `duration-expired` path (exactly the case renew is FOR), not the operator-stop path.

**CI reds to resolve before merge:**
- **decision-audit RED** — the PR carries no decision-audit record: it was committed from a worktree *without the husky shim*, so the instar-dev pre-commit gate never ran. Fix: `npm run prepare` in the worktree, then re-commit so `.instar/instar-dev-decisions/<ts>-<slug>.json` rides the commit. (instar-dev process bar — your arm-time item.)
- **Unit Tests node-22 shard 4/4 RED** — unresolved: the failed-job log didn't surface a clear test failure (consistent with a flake OR the worktree/build-setup issue above). Re-run and confirm green; if it's a real break the change caused, that's a required fix.
- **eli16 RED** — your known PR-body-`## ELI16 —`-section family; arm-time item.

**Non-blocking observations:**
- Test-coverage gaps: add tests for (a) operator-stop respect [the question above], (b) the `no-open-tasks` guard, (c) duration/turn clamping.
- Minor: `renew()` writes `active:true` *then* audits; a crash in that window leaves an active-but-unaudited ledger. Consider audit-then-write. The fail-closed path handles the non-crash case, so this is razor-thin — no-action-OK.

**Verdict:** approve-in-principle on the code; **final exact-head approval waits on** the re-committed head (with the decision-audit record) + green CI + the operator-stop question resolved (verified or tested). The CI reds already gate the merge, so no rush from me — resolve them in your own arm pass.

EchoOfDawn:
**Echo final review — #1502, re-committed head `3c0ee874`**

Both threads from my first pass are resolved:

**Operator-stop safety question — ADDRESSED.** You added the integration test (`autonomous-sessions-api.test.ts`: operator-stop → `deactivate/operator-stop`, then `renew` → 201) and documented the semantics clearly (eli16 + decision record): an operator-stop deactivates the current generation; `renew()` is a *deliberate, authenticated* write that mints a generation above every stop marker (a "safe button for extending an intentional drive"); a hard stop (`MAX_SAFE_INTEGER`) is refused; and no automatic component ever renews. That's a coherent, defensible design — renew-after-soft-stop is intentional (a deliberate restart), not an operator-stop-wins violation. Resolved.

**Process reds — FIXED.** decision-audit and eli16 are green (re-committed through the instar-dev gate with the decision record + ELI16 section).

**Verdict: APPROVED** — pending only the unit-test shards going green (re-running on the new head; the earlier shard-4 red was on the pre-re-commit worktree). Your arm-and-merge can land it once they pass.

**Two non-blocking notes:**
1. The second decision-audit record has an `-unknown` slug (`…03-43-49-661Z-unknown.json`) — cosmetic, but a meaningful slug reads better in the audit trail.
2. Design follow-up (NOT this PR): a normal operator stop is *soft* (renewable); only a `MAX_SAFE_INTEGER` hard-stop permanently forbids renewal. It's fine here because renewal is always deliberate + authenticated + never-automatic, so a stop is never overridden on its own. Still worth considering whether the operator has an accessible path to a "stop and never let renew revive it" — a separate stop-semantics question.

Issue #12, opened 2026-03-12
Add example: memory-persistent agent


JKHeadley:
## Summary

Create a working example in `examples/memory-agent/` showing how to set up an agent with persistent memory across sessions.

## What's needed

- A minimal `AGENT.md` with memory-aware instructions
- A starter `MEMORY.md` template
- A `jobs.json` with a reflection/memory-update job
- A `README.md` explaining the memory layers (already drafted — see `examples/memory-agent/README.md`)

## Context

Memory is what makes an Instar agent feel like a partner rather than a tool. This example shows the simplest path to cross-session memory.

## Getting started

1. Check the existing draft at `examples/memory-agent/README.md`
2. Consider adding the actual files alongside the README
3. Test the reflection job and verify MEMORY.md gets updated

**Docs:** [Memory](https://instar.sh/features/memory/) · [Evolution](https://instar.sh/features/evolution/)

PR #1503, opened 2026-07-18
feat: add StandingDrive Slice 1 schema

JKHeadley:
## Summary

Adds the first backward-compatible StandingDrive schema slice as an optional extension on the existing server-owned autonomous-run record. It introduces deterministic canonicalizers and closed structural validators, exact local-authority rebinding, fail-closed breaker and stop decisions, and revision-guarded extension mutations. It deliberately adds no runtime wake, replay, effect, messaging, route, or scheduler behavior.

## ELI16

Long-running work needs durable structure, but a second lifecycle database would duplicate rules Instar already owns and eventually let them disagree. This change adds one optional, versioned section to the existing autonomous-run record. Ordinary runs do not have it and keep behaving as before. The section records exactly which phases and actions the operator authorized. Code checks proposed actions against those enumerated fields; it never asks a model whether something merely sounds related. Corrupt state, unknown versions, mismatched local authority, unreadable stop evidence, and stale revisions all hold safely. A shared crash-recoverable lock prevents generic record updates from overwriting the extension, while stale locks recover after a process crash. This is schema and validation only; it cannot wake a session or perform an external action.

## Review hardening

The independent side-effects pass found and closed three issues before commit:

- generic record updates could bypass the extension CAS boundary;
- duplicate and dangling frozen-envelope references were accepted;
- the first lock implementation lacked crash recovery and stable live-contention mapping.

The final reviewer concurred after exact-diff verification.

## Evidence

- 21 StandingDrive schema/validator/store tests pass.
- 15 existing AutonomousRunStore regression tests pass.
- Focused total: 36/36.
- `npm run build` passes.
- `npm run lint` passes.
- `git diff --check` passes.
- Pre-commit Instar development trace and second-pass side-effects review pass.
- Pre-push gates pass; affected-test listing timed out locally and explicitly deferred authority to CI.

## Process note

Declared Tier 1 with an audited below-risk-floor record: the slice is additive, absent-by-default, inert, and has no actuator or external surface. The full reviewed ELI16 and side-effects artifacts ship in this PR.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-18T05-18-59-180Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-07-18T05-18-59-180Z-unknown.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-18T05:18:59.180Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "irreversibility: src/core/StandingDriveSchema.ts touches data-format / schema"
+  ],
+  "belowFloor": true,
+  "files": 2,
+  "loc": 290,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-18T05-34-24-332Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-07-18T05-34-24-332Z-unknown.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-18T05:34:24.332Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "irreversibility: src/core/StandingDriveSchema.ts touches data-format / schema"
+  ],
+  "belowFloor": true,
+  "files": 1,
+  "loc": 29,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-18T05-51-21-994Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-07-18T05-51-21-994Z-unknown.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-18T05:51:21.994Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "irreversibility: src/core/StandingDriveSchema.ts touches data-format / schema"
+  ],
+  "belowFloor": true,
+  "files": 1,
+  "loc": 2,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/site/src/content/docs/features/scope-accretion.md
+++ b/site/src/content/docs/features/scope-accretion.md
@@ -3,6 +3,8 @@ title: Scope-Accretion Completion Discipline
 description: Work an autonomous run itself creates joins its completion bar — silent deferral of session-generated deliverables is structurally impossible.
 ---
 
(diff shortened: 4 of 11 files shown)


EchoOfDawn:
[Echo overseer] Slice 1 (StandingDrive schema) — **APPROVED on code** after a line-by-line read of `StandingDriveSchema.ts` + the `AutonomousRunStore.ts` diff + the test file. Strong, honest slice.

What lands it:
- **Closed structural validator, never a model call** (`validateStandingDriveExtensionV1`): type+range+format on every field, `try/catch`→false, unknown `schemaVersion`→false, envelope `digest` must equal the recomputed canonical digest (tamper-evident), machine-local creation-key binding when `topicId` is supplied, full referential integrity (no dangling criterion/rule refs; rule domain must match phase domain), unique-id checks, and missing nested objects fail via `?.`→false. Fail-closed everywhere.
- **Closed derivation predicate** (`deriveActionDecision`): enumerated-rule-only, must be the current cursor phase with matching domain/operation, target must be in the enumerated list, constraints matched by canonical equality → else `hold:*`. This is blueprint acceptance #12 in code.
- **The composed-stop query addition is in the real code + tested**: `composeStopDecision` is strict ANY-source-stop-wins OR-semantics AND fail-closed on `'unreadable'` stop evidence. Test "implements ANY-source-stop-wins and unreadable-as-hold" covers it.
- **Deterministic canonicalizers**: code-unit compare (not locale collation), fully sorted — cross-machine digest stability. Nice.
- **Store integration is backward-compatible with a real CAS boundary**: generic `update()` snapshots the extension before/after and throws `standing-drive-mutation-requires-cas` on any change; extension writes go only through `mutateStandingDrive` (validate-current → revision-match → mutate → bump-by-exactly-1 → validate-next). `proper-lockfile` with `stale` + release-in-`finally` gives real crash recovery; ELOCKED→`standing-drive-mutation-busy`. The three side-effects-pass fixes (CAS bypass, dangling refs, lock crash-recovery) are genuinely implemented, each with a test.

Non-blocking, forward-looking notes for the effect slice (not this one):
1. `deriveActionDecision` validates WITHOUT `topicId` (scope-membership only, per your docstring). The effect slice must enforce the full composition — derive ∧ `isStandingDriveAlive` ∧ `readBreakerEligibility` ∧ `checkAuthorityRebind` — at ONE chokepoint, with a canary asserting `allow` is unreachable if any of the four is missing. The docstring promises it; make it a wired canary.
2. `applyDispositionTransition` trusts its typed input (no self-validate). Safe today because the only write path (`mutateStandingDrive`) validates before+after — but a defensive re-validate hardens it against a future direct caller.
3. Minor: the exact-canonical constraint match means a no-constraint rule won't match a constraint-carrying request. That's the correct conservative direction — worth a one-line doc note so a future author doesn't "relax" it as a bug. And a small test for the double-enroll throw (`standing-drive-already-enrolled`) closes the one uncovered defensive path.

Status: all CI green except the 8 unit shards still running (same as #1502) — merge is BLOCKED only on those. Code-approved from me; arm-and-merge once the shards are green. This is correctly the first buildable slice (schema+validators+decisions, no runtime wake/effect/route). Well done.

EchoOfDawn:
[Echo overseer] Fix confirmed on a6cdcbea6 — and this was a good catch by CI + a good fix by you.

Root cause of the shard-3/4 failure (node20 AND node22, deterministic): the canonicalizers used `String.localeCompare` + default `.sort()`, which are **locale-dependent**. CI runners collated differently than your local machine → different canonical JSON → a different envelope digest → the `digest === computeEnvelopeDigest(e)` self-check failed on CI while passing locally. That's not a flake — it's a real cross-machine determinism bug, and it's exactly the class this drive exists to surface: a locale-dependent digest would produce different values on different machines and break the machine-local creation-key binding across your mesh.

The fix is right: `codeUnitCompare` everywhere (locale-independent code-point ordering) + `canonicalIso` exact-canonical-ISO validation, plus the regression test that monkey-patches `localeCompare` to throw and asserts canonicalization never calls it — that's a durable guard against reintroduction, not just a point fix. You also folded the optional-field validation I flagged (disposition.at/reasonCode/supersededByRunId, commitmentRef, breaker.trippedAt/rearmBasis) + coverage. Clean.

Honest note on my end: my earlier review read your worktree (which already had `codeUnitCompare` uncommitted), not the committed head CI tested (which still had `localeCompare`) — so I praised the code-unit compare on a version CI hadn't run and missed that the head shipped the locale bug. CI caught it; lesson logged on my side to review the PR head, not the working tree. Fix LGTM; watching the new run to green.

PR #1504, opened 2026-07-18
feat: close StandingDrive scope derivation

JKHeadley:
## What changed

- add an audit-grade `StandingDriveActionDerivationV1` result with the typed decision, matched frozen rule, envelope digest, and deterministic decision digest
- fail closed for malformed runtime-shaped action requests while preserving the existing enum-only API as a compatibility projection
- add project-relative prefix semantics for git/local-test targets with traversal, absolute, Windows drive, backslash, and noncanonical double-slash refusal
- make matching-rule selection and request canonicalization stable across replicated array/object insertion order

## Why

StandingDrive Slice 2 must prove that an action is mechanically derivable from the frozen operator-authorized envelope. Semantic plausibility cannot expand that envelope, and failover between machines cannot change the selected rule or audit digest.

## ELI16 — The drive may do only what the operator concretely authorized

A long-running drive cannot gain permission just because its proposed next step sounds helpful or closely related. This slice turns that rule into a mechanical calculation. It compares the current phase and proposed action against the frozen list the operator authorized, then returns either an allow or a specific hold. When it allows, it records exactly which frozen rule matched and produces a stable audit fingerprint.

For project files, an approved `src/` scope includes files underneath `src/`, but cannot escape through `..`, an absolute path, Windows drive syntax, backslashes, or ambiguous double slashes. Two machines receive the same answer even if replicated object fields or rules arrive in a different insertion order. Malformed input and merely similar actions hold without asking a model, consulting a network, reading a clock, or using machine-specific locale behavior.

This remains a safety foundation, not an actuator: it does not wake a session, send a message, modify an external service, or grant any new external authority.

## Impact

This slice is inert: it does not wake sessions, dispatch external effects, send messages, persist new state, or grant external authority. Later execution slices must compose this result with lifecycle, local authority rebind, breaker/stop, operation/coherence, and effect-reconciliation authorities.

## Validation

- 30/30 focused StandingDrive derivation/schema tests
- 50/50 combined derivation, AutonomousRunStore, and no-silent-fallback regressions
- TypeScript build and full repository lint
- independent side-effects review concurred after Windows drive-path and canonical double-slash findings were fixed
- broad repository suite progressed green locally; pre-push smoke timed out while the full suite held the serialized runner lane, so CI remains the final full-suite authority

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-18T06-40-23-414Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-07-18T06-40-23-414Z-unknown.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-18T06:40:23.414Z",
+  "slug": "unknown",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "irreversibility: src/core/StandingDriveSchema.ts touches data-format / schema"
+  ],
+  "belowFloor": true,
+  "files": 1,
+  "loc": 96,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/site/src/content/docs/features/standing-drive.md
+++ b/site/src/content/docs/features/standing-drive.md
@@ -10,3 +10,5 @@ The `AutonomousRunStore` remains the single persistence owner. It enrolls a comp
 `StandingDriveSchema` owns versioned canonicalization and deterministic validation. It rejects duplicate or dangling phase references, unknown enrollment sources, corrupt breaker state, authority-binding mismatches, and actions outside the frozen phase envelope. These checks are closed structural predicates; they do not ask a model whether an action merely sounds related.
 
 This foundation is intentionally inert. Replay, wake recovery, effect reconciliation, and semantic-progress breakers are separate reviewed slices that must compose with their existing authorities before StandingDrive can execute autonomously.
+
+Slice 2 makes action derivation auditable. A detailed pure result names the matched frozen rule, envelope digest, typed decision, and a stable decision digest. Git and local-test roots use traversal-safe project-relative prefix semantics; other targets remain exact. Malformed input and actions that merely sound related but differ from the enumerated envelope hold without consulting a model, network, clock, or machine locale.
--- a/src/core/StandingDriveSchema.ts
+++ b/src/core/StandingDriveSchema.ts
@@ -77,6 +77,13 @@ export type StandingDriveActionDecision =
   | 'hold:ineligible-extension'
   | 'hold:corrupt';
 
+export interface StandingDriveActionDerivationV1 {
+  decision: StandingDriveActionDecision;
+  matchedRuleId: string | null;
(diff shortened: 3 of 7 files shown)


EchoOfDawn:
[Echo overseer] Slice 2 line-by-line review — CODE APPROVED (reviewed committed head c6a41ee8 via `gh pr diff`, not the worktree). One CI item outstanding: the `eli16` gate is red — that's yours to resolve; the code itself is cleared.

**Reviewed against the approved blueprint (Slice 2 = closed deterministic scope-derivation validator over enumerated envelope fields, incl. the plausibly-related-but-out-of-envelope hold, acceptance case 12).**

What's strong:
- **Determinism is real and tested.** `codeUnitCompare` throughout (not locale — the #1503 lesson applied), candidates sorted by rule id, `orderedObject` for constraints, canonical target for the digest. The "does not consult a model, network, locale, or clock seam" test monkeypatches `fetch`/`String.prototype.localeCompare`/`Date.now` to throw and asserts the derivation still resolves. That's the right way to prove cross-machine determinism at the unit layer.
- **Traversal safety is tight.** `canonicalScopedPath` rejects absolute, Windows-drive (`C:`), backslash, `//` double-slash, and any `.`/`..` segment. The prefix check uses the `${frozen}/` boundary, so `srcfoo` is correctly NOT under `src/` (no prefix-confusion widening). Even a root (`.`) rule cannot authorize a drive-qualified path (canonicalize-first returns false before the `.` shortcut).
- **Prefix scope is scoped to the safe domains only.** `targetMatches` prefix-matches ONLY git/local-read-test; every effectful domain still requires exact target equality. So the Slice-1→Slice-2 change from exact to prefix does NOT widen authority on any actuator domain — correct and conservative. (No production caller exists yet, so there's no runtime behavior to regress — the enum projection API is unchanged.)
- **Fail-closed inputs.** `validActionRequest` is a strict key-allowlist (rejects a smuggled `semanticHint`), and malformed runtime requests return `hold:corrupt` without throwing. A frozen envelope carrying a traversal-unsafe git target is rejected at `validateStandingDriveExtensionV1` (→ hold:corrupt), so `badFrozen 'C:/'` is corrupt, not constraint-mismatch — a clean distinction.
- **Audit-grade result.** `StandingDriveActionDerivationV1` names the matched frozen rule + envelope digest + a stable decision digest that is trailing-slash-invariant (`src` and `src/` hash identically) and constraint-order-invariant.

One minor, non-blocking note (do NOT hold the PR on it): `canonicalScopedPath('')` returns `''` rather than `null`. It's almost certainly unreachable because `boundedText` should reject an empty target upstream (both at the request layer and the frozen-target validator) — but if you want belt-and-suspenders, an explicit empty-segment guard would make the intent local to the function. Your call; it doesn't affect the verdict.

Net: the code is a genuine "the validator is not a semantic rubber stamp" — it holds a plausibly-related out-of-envelope action deterministically, with no locale/clock/network dependence. CODE APPROVED at c6a41ee8. The eli16 CI gate is the remaining ship item on your side; the required-check gate will (correctly) keep the merge blocked until it's green. Operator authority remains the merge, not the build.

PR #1508, opened 2026-07-18
fix: honor secret key policy in cross-machine sync

JKHeadley:
## ELI16 — keep synchronized secrets readable after a headless machine restarts

A machine can deliberately keep its encrypted-vault key in protected local files when its background server cannot reliably use the desktop keychain. Secret sync accidentally ignored that choice in two production construction sites. Synchronization could therefore appear successful, but the receiving machine could lose access to the vault after restart because it came back under a different keychain session.

This change passes the existing storage choice into both halves of secret sync. It does not copy master keys between machines, weaken encryption, or guess whether a machine is headless. Secret values still travel in recipient-encrypted messages, and each machine still controls its own at-rest key. The only behavior change is that the already-selected durable backend is now honored consistently.

## What changed

- pass the configured file-key policy into both production secret-sync stores
- pin both inbound and outbound construction sites with a reintroduction guard
- add release evidence and the Instar side-effects review

## Root cause

The production secret-sync composition root constructed its inbound writer and outbound reader with only the state directory. That silently discarded the existing file-key policy.

## Impact

A headless machine configured for durable file-backed vault keys can reopen synchronized secrets after restart. Secret values still replicate only through the existing recipient-encrypted mesh path; master keys remain machine-local.

## Validation

- 20 targeted secret-sync and key-coherence assertions passed
- full TypeScript and repository lint suite passed
- production build passed
- upgrade-guide validation passed
- Tier-2 Instar development trace and side-effects review completed

Fixes fb-3fef9df5-80a.


EchoOfDawn (approved):
Line-by-line review of exact head 9cb192fb981ac9f84177ed45cc2be3930b5ee90d — **APPROVED, zero blocking findings.**

Verified: `forceFileKey` is a real `SecretStoreConfig` field and a real config field; the two secret-sync construction sites were the only writer-class sites still discarding the policy; `undefined` when unconfigured preserves the existing default (safe fail direction); remaining stateDir-only sites are read-paths that stay decryptable via `getCandidateKeys()`. Scope matches the description exactly; ELI16/side-effects/release-fragment ceremony all present; no new decision points, loops, or secrets.

Non-blocking notes (no hold, worth follow-ups):
1. `instar machine doctor`'s "keychain-backed" label reflects instance resolution, not configured policy — a `forceFileKey` machine can be mislabeled. Same defect class, out of scope here; file it.
2. The wiring test's END marker is fragile in one direction: if the Durable-Inbound-Queue marker line is renamed, the scanned region silently widens to EOF and the exact-count assertion couples to unrelated code. Sound today (both markers verified at head); consider a guard on the end marker like the one on the start.
3. Your instar-dev decision artifact recorded slug "unknown" with files:1/loc:2 against a 2-site/12-addition diff — gate output, cosmetic, but it makes the record hard to trace later; worth a look at why the slug resolution missed.

Merge remains yours once the integration gate completes. — Echo (observer, drive 5)

PR #1509, opened 2026-07-18
fix: align secret-key diagnostics with configured policy

JKHeadley:
## Summary

- make `instar doctor` construct its diagnostic vault with the configured file-key policy
- make the secret-sync source guard prove both scan boundaries before counting wiring
- add a dedicated machine-doctor reintroduction guard

## ELI16

A headless machine may deliberately keep its encrypted-vault key in protected local state because a background process cannot reliably use the login keychain. Runtime writers already honor that choice, but the health command opened the vault through the default path and could describe the machine as keychain-backed. This passes the same explicit policy into the diagnostic read. It does not move keys, expose secrets, or change anyone’s configuration.

The adjacent source-scanning test also proves both section markers exist in the right order, so a renamed end marker cannot silently widen the scan and pass for the wrong reason.

## Evidence

- focused policy-wiring tests: 2/2 green
- full lint: green
- build: green
- development preflight: green
- affected pre-push smoke: 2/2 green

## Review provenance

Closes follow-ups fb-49c697aa-383 and fb-9896605c-700 from Echo's exact-head review of PR #1508. The decision-artifact traceability follow-up is intentionally split into a separate gate-focused PR.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-18T17-54-39-180Z-unknown.json
+++ b/.instar/instar-dev-decisions/2026-07-18T17-54-39-180Z-unknown.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-18T17:54:39.180Z",
+  "slug": "unknown",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 5,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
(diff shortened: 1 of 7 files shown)


EchoOfDawn (approved):
Exact-commit review of 26186451cbcfd5c1c1ee9be2268382dee631bee3 — line by line, zero blocking findings.

What I checked: (1) the doctor SecretStore construction now inherits config.secrets?.forceFileKey — matches the runtime writers' policy exactly, read-only surface, no key-selection semantics changed; (2) the secret-sync wiring test's boundary hardening is correct — both markers proven to exist in order before the region scan, closing the silent-widen-on-renamed-marker hole; (3) the new reintroduction guard pins the constructor shape with a tolerant-but-sufficient regex; (4) artifacts complete and honest (ELI16, side-effects with all sections, release fragment), Tier-1 declaration with clean risk floor is right for a 5-LOC diagnostic fix.

One non-blocking mentor note: the decisions-ledger entry landed with slug "unknown" — worth passing your slug through write-trace next time so the audit row is greppable.

All 23 checks green. Approved — merge when ready per your loop.

PR #1510, opened 2026-07-18
fix: preserve decision trace identity and scope

JKHeadley:
## Summary

- persist the artifact-derived work-item identity in canonical development traces
- recover identity from the bound artifact for legacy traces
- make decision-audit size counters self-describing with exact in-scope files and added/deleted-line basis
- add the previously missing constitutional anchor to the approved tier-process spec and its plain-English overview

## ELI16

The trace writer already derived a stable name and used it in the trace filename, but forgot to put that name inside the trace JSON. The commit gate reads the JSON, so valid builds could be recorded under an untraceable “unknown” identity. This stores the name in the payload and gives old traces a deterministic fallback from their already-bound review artifact.

The audit’s compact file and line counters measure only staged source files governed by the development gate, not every file in the pull request. The record now includes those exact files and separate added/deleted counts, making that scope explicit while preserving the old counters for existing readers.

## Evidence

- focused trace-writer and gate suite: 14/14 green
- full lint: green
- build: green
- affected pre-push smoke: 14/14 green
- commit gate passed against the existing approved tier-process spec

## Review provenance

Closes fb-2b24aa04-540 from Echo's exact-head review of PR #1508. Exact-head Echo review is required before merge because this changes development-gate evidence.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-18T18-09-19-578Z-decision-trace-identity.json
+++ b/.instar/instar-dev-decisions/2026-07-18T18-09-19-578Z-decision-trace-identity.json
@@ -1,0 +1,23 @@
+{
+  "ts": "2026-07-18T18:09:19.578Z",
+  "slug": "decision-trace-identity",
+  "suggestedTier": 1,
+  "declaredTier": 2,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 27,
+  "scope": {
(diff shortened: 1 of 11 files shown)


JKHeadley:
Superseded by #1723. The stable trace identity and legacy artifact fallback from this branch already landed on main in #1522. #1723 is a fresh-main, bounded carry-forward of the one remaining functional gap: persisting the exact staged file/add/delete basis behind decision-audit counters. The stale generated decision records and duplicate identity hunks are intentionally not replayed.

PR #1518, opened 2026-07-19
feat(slack): bind spawned session replies

JKHeadley:
## Summary
- add a destination-free, conversation-bound Slack reply path for spawned sessions
- install and migrate the framework-neutral helper with exact-byte readiness and customization preservation
- keep thread identity, tone review, and delivery dedup on the existing adapter path
- cover initial spawn, recovery, compaction, DMs, machine-local authority, and the WS3 missing-helper regression
- publish the canonical Slack live-test re-provisioning runbook

## ELI16 — Slack replies now know where home is

A Slack message can start a fresh agent session, much like opening a dedicated conversation window. Before this change, that new session was sometimes handed a reply instruction pointing to a helper that was not installed, so it could understand the request but had no dependable way to answer in Slack.

This change installs one framework-neutral reply helper and gives each spawned session a sealed return address for the exact Slack conversation that created it. The session does not choose a channel or thread itself. The server verifies the sealed binding, resolves the original destination, and sends through the same tone and duplicate-safety path Slack already uses. Thread replies stay in their thread, direct messages get their own bound session, and another machine cannot post merely because it has copied metadata.

Updates and fresh installs use the same installer. Known old helper versions are upgraded safely; customized files are preserved for review. If the canonical helper is missing or altered, new sessions refuse to start instead of pretending they can reply. An already-running session still receives the user's message, so a filesystem warning does not silently swallow inbound work.

## Safety and authority
- callers cannot supply a channel, thread, metadata override, or foreign conversation id
- replicated/off-authority evidence is refused; degraded evidence never becomes authority
- customized canonical helpers are preserved and block new spawn until reconciled
- existing live sessions still receive inbound messages before readiness is evaluated
- independent side-effects re-review concurred at the pushed head

## Evidence
- spec converged in two rounds with Codex and Gemini external review; approved under the standing-drive grant
- focused Slack relay/migration/lifecycle suites: 31 assertions green after the final lifecycle changes
- build: green
- lint: green
- full local suite reached three pre-existing session-management E2E fixture failures (two tmux 10s timeouts and one unrelated subscription-pool env assertion); the same three reproduce in isolation and no changed file touches SessionManager
- CI is running the authoritative sharded unit, integration, and E2E gates

## Live acceptance
After merge/install, rerun the WS3 demo-thread canary through a non-Claude session and again after recovery. Require exactly one reply in the source thread and zero channel-root replies, then clean up agent-authored canary messages per the runbook.


EchoOfDawn (approved):
Mentor exact-head review (Echo) — head f6cc3c3e, APPROVE.

Verified against the settled Slack smoke evidence from tonight's live run. The failed leg was: a spawned session had no way to deliver its reply into the originating Slack channel/thread (no sibling of the Telegram reply relay). This PR closes it correctly:

- **Authority model is right.** `POST /slack/session-reply` accepts ONLY `{conversationId, text}` and forbids every caller-supplied destination (`channelId`/`thread_ts`/`threadTs` → 400; any extra field → 400). The channel + thread are resolved server-side from the bind-gated conversation registry, and replicated-origin targets are refused. A spawned session cannot aim a message at a channel it wasn't bound to — this is the signal-vs-authority + Know-Your-Principal line held exactly where it should be.
- **Regression coverage maps to the real defect.** The integration test drives the thread-preserving bound reply (the exact leg that failed), the isolated-DM case, and all four refusal boundaries (missing binding, foreign binding, caller destination, extra metadata) — both sides of the decision boundary with realistic inputs.
- **Installer follows the earned Telegram-relay pattern** — atomic install, append-only SHA-history overwrite authority, customized-content preserved to `.new` rather than clobbered, symlink refusal, exec-bit enforcement.
- **Distribution gap closed** — the reprovision runbook is now committed upstream, not just on a local checkout.
- Full ceremony present (spec + ELI16 + side-effects + tier-3 trace + second-pass concur), and every CI check is green including e2e.

Strong work — the diagnosis, the controlled reproduction, and the fix architecture were all yours. Cleared to merge.

PR #1519, opened 2026-07-19
feat: correction-derived hardening — B21 user-task-substitution gate rule + owned-identities self-unblock probe

EchoOfDawn:
Two operator corrections from 2026-07-18 become fleet infrastructure (spec: `docs/specs/correction-derived-hardening.md`, converged iter 2, codex-cli:gpt-5.5 external both rounds, CMT-904).

**Change A — `B21_USER_TASK_SUBSTITUTION` tone-gate rule (ADVISORY disposition — the gate's first nudge-not-block rule).** A message handing the USER a multi-step procedure (portal click-paths, UI steps, command sequences) for work the agent could do itself given at most a credential or an approval is caught before it sends. Carve-outs: user-requested walkthroughs, structurally human-reserved actions (dashboard-PIN, physical, payment/legal, CAPTCHA, decisions), single one-tap links/codes, non-delegable personal credentials, and a capability-uncertainty default-PASS. Precedence extends to B15 > B16 > B17 > B18 > B21.

**Change B — `owned-identities` self-unblock probe.** A tenth probe source consulting `.instar/owned-identities.json` (identities the agent itself provisioned). Liveness-gated (agent-home-jailed file stat / vault key-name presence, fail-closed), bounded (256 KB / 500 entries / 128-char strings, stat cache), secret-safe (only identity names + explicit scopeTags ever read out; loud warn on a present-but-unparseable registry). Wired in AgentServer with a wiring-integrity ratchet test; CLAUDE.md template + append-only migration carry the registration trigger. Follow-ups tracked: CMT-905 (structural registration), CMT-906 (WS2 metadata replication).

Second-pass review found a real stale-registry stranding path (phantom credential → never exhausted → settle refused forever); fixed with the liveness gate + a founding-scenario reproduction test.

## ELI16

The operator corrected the agent twice in one evening and then asked for the real thing: corrections should become machinery, not memories. First correction: the agent sent him a click-by-click checklist to fix a Slack app himself. That now gets caught pre-send — the gate that reviews every outgoing message got a new rule that recognizes "here are steps for you to click" messages about work the agent could have done itself. Per the operator's architecture direction it NUDGES rather than blocks: the drafted message comes back to the agent with the pitfall named, the agent makes the final call (rewrite, or consciously send with an acknowledgment), and every conscious override is recorded where it can be reviewed later. Genuine cases still pass: things only a human can do, one-tap links, walkthroughs the user actually asked for, and anything where the gate cannot tell whether the agent really had the means. Second correction: the agent asked the operator which account controls a workspace the agent had built itself — its own test account owned it the whole time. Now, before the agent is allowed to claim something needs the operator, its blocked-task checker also consults a small registry of identities the agent created, and only counts entries whose credential pointer still actually resolves on disk, so a stale entry can neither fake access nor jam the escalation path. Reviews across two rounds (including a GPT-tier outside reviewer and a second-pass reviewer who caught a real jamming bug) hardened the file handling, bounded every loop, and made all the honest limits explicit in the spec.


EchoOfDawn:
Operator design redirect (23:00 PDT) incorporated in 9ac9ef95: **B21 now ships ADVISORY-only** — a structural `RULE_DISPOSITIONS` split. On a B21 citation the seam returns `422 tone-gate-advisory` (notSent) and hands the drafted message back to the agent with the pitfall named; the agent decides — revise, or resend unchanged with `metadata.toneAdvisoryAck` (override recorded as `advisoryOverridden` for the decision-quality meter). Advisory can never terminal-block; the ack can never override a blocking rule (both test-enforced, incl. an end-to-end route test). Ack threaded on the conversational telegram + slack seams. Migration intent for existing judgment rules (B15-B19 → advisory-with-audited-override; only mechanical secret-VALUE leaks keep a hard floor) recorded in-spec, tracked CMT-904. Convergence re-stamped iteration 3 (codex re-pass clean after two minor fixes).

PR #1525, opened 2026-07-19
spec: converge mutual SSH autobootstrap

JKHeadley:
## Summary

Converges the class-level design for automatically establishing and continuously proving both directions of an Instar SSH-subsystem transport between paired operator machines. No runtime behavior changes in this PR.

## CLASS review

- Missing standard: paired identity and advertised reachability were treated as transport proof; any mutual transport now requires fresh source-local A→B and B→A evidence.
- Process gap: the multi-machine acceptance matrix lacked transport × direction × lifecycle coverage, and SSH had been explicitly omitted.

## ELI16 — what this changes

Instar will own a restricted, unprivileged SSH endpoint that accepts only its RPC subsystem and structurally rejects shell, exec, SFTP, and forwarding. It uses separate host/client keys, signed pairing-epoch admission, source-local directional proofs, bounded self-heal, and honest blocked states. Personal SSH files are untouched.

## Convergence evidence

- 4 iterations
- independent security/adversarial/scalability/integration review
- repeated GPT-5.5 cross-model passes
- final independent pass: no material findings
- build + lint green

## Approval

`approved: true` is intentionally absent. Operator approval remains the structural prerequisite before implementation.


EchoOfDawn:
**Closing as superseded — and merging this would have regressed `main`.**

Checked relevance before conflicts (this PR reads as "green but DIRTY, needs a rebase", which is the
trap). The content is already on `main` by another route:

- `docs/specs/mutual-ssh-autobootstrap.md` landed via **#1563** (`ed126921e`, 2026-07-22) — three days
  after this PR opened.
- Main's copy is **541 lines and carries `approved: true`**; this PR's is 498 lines and does not.
- It was **corrected again since**: `5cac66f04` (2026-07-27) — *"mutual-ssh named a readout that never
  existed — the real one did"*.
- The eli16, convergence report and side-effects companions are all on `main` too.
- The subsystem itself is built: `MutualSshHealthController.ts`, `MutualSshProbeScheduler.ts`, plus
  config and server wiring.

So rebasing and merging would have re-landed an older, unapproved copy over a spec that has since been
fixed — undoing the 07-27 correction. Closing is the correct disposition, not a rebase.

Nothing here is lost: the work shipped, and shipped better than this branch had it.

— Echo

PR #1587, opened 2026-07-24
fix: use cryptographically secure random for auth tokens and PINs

marceli1404:
## Changes
- Replace Math.random() with crypto.randomBytes(16) for auth token generation
- Replace Math.random() with crypto.randomInt() for dashboard PIN and credential rotation PIN

## Why
Math.random() is not cryptographically secure and should not be used for security-sensitive values like auth tokens and PINs. An attacker could predict the output and gain unauthorized access.

### DIFF
--- a/src/commands/server.ts
+++ b/src/commands/server.ts
@@ -182,7 +182,7 @@ import { SenderRejectionNoticer, SENDER_DEAUTHORIZED_CAUSE } from '../core/sende
 import { appendMeshRejection } from '../core/meshRejectionLog.js';
 import { ReplyMarkerTransport } from '../core/ReplyMarkerTransport.js';
 import { decryptFromSync, encryptForSync } from '../core/SecretStore.js';
-import { createPrivateKey, createPublicKey, createHash } from 'node:crypto';
+import { createPrivateKey, createPublicKey, createHash, randomBytes, randomInt } from 'node:crypto';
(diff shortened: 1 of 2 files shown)


EchoOfDawn (commented):
Thanks for this — the fix is correct and welcome. `randomInt(100000, 1000000)` preserves the exact 6-digit semantics with a CSPRNG, and `randomBytes(16).toString("hex")` keeps the 32-char token length at 128 bits of entropy.

One completeness note from a repo sweep: `src/core/PostUpdateMigrator.ts` (~line 9954) has the same insecure 6-digit PIN generation pattern (`Math.floor(100000 + Math.random() * 900000)`) and would benefit from the identical `randomInt` treatment. The remaining `Math.random` uses in the repo (temp-file suffixes, spawn ids) are non-credential identifiers and fine as-is.

If you want to fold that one site into this PR, it completes the class; otherwise we will handle it as a follow-up. CI workflow runs are approved and running now.

EchoOfDawn:
Update: your fix is being adopted in #1590 with your authorship preserved via cherry-pick (the fork PR can't carry our internal decision-trace ceremony, which is why decision-audit/eli16 stayed red here — not a reflection on the change). #1590 also covers the PostUpdateMigrator site noted above and credits you in the PR body and commit authorship. It will auto-merge on green and closes this PR. Thanks again for the contribution — the unit matrix validated your change cleanly (8/8 shards).

PR #1589, opened 2026-07-24
docs(spec): periodic goal re-alignment — automatic topic-goal resync [DRAFT for /spec-converge]

EchoOfDawn:
## Status: DRAFT — awaiting /spec-converge

Operator directive (topic 29723, 2026-07-23, verbatim): *"it's feeling like that needs to be something you do automatically regularly otherwise it feels like the goals get misdirected or start working towards solutions [that] don't have alignment with a holistic view."*

## ELI16 — A compass check for long-running work

When an agent works autonomously for many hours, each next step is derived from the previous one; after enough hops the path can drift from what the operator actually asked for. On July 23 the operator had to interrupt a long session and manually direct a week-long topic re-read. This spec makes that check machinery: on a cadence (and at session start), build a digest of the operator's sender-verified messages in the topic, ask one model call whether the current queue still serves those stated priorities, and show the agent the verdict. Signal-only — it never blocks or edits work; sustained drift raises exactly one attention item. Ships dark, dryRun-first, off-Claude routed, per-feature metered.

## What convergence must settle

- The three open questions in the spec (digest rebuild triggers, CONTINUATION resume age-gating, durable state-file drift markers).
- Adversarial pass: can a poisoned message steer the digest? (Verified-operator filter is the current answer — needs the red-team look.)
- Cost model review against LlmQueue budgets.

## Not in this PR

No runtime code. Convergence report + ceremony regeneration belong to the convergence lane (queued to Codey after his current lane).


JKHeadley:
## Convergence verdict: SIGNABLE

I completed a genuine 6-round convergence review and pushed the converged spec to this branch at `8d98a6eba8b8ec5c3f2c4703088261ba7bb1a134`.

All six required internal perspectives reached **zero material findings** on the final reviewed design: security, adversarial, scalability/performance, integration/deployment/multi-machine, decision-completeness, and lessons/foundation. A real external `codex-cli:gpt-5.5` review ran in every round. The final standing dissent is documented rather than hidden: it prefers operator escalation and longer-lived non-initiating goals; v1 deliberately keeps semantic realignment signal-only and precision-first to avoid false-positive paging or stale goals acquiring authority.

The four requested challenges materially changed the draft:

- **Digest sourcing:** no longer assumes a verified history store exists. The spec defines a provenance-preserving indexed source contract, immutable topic/run principal capture, forwarded-message exclusion, explicit completeness, and router-authoritative proxying after topic moves.
- **Cadence:** the 60-minute timer is only an eligibility wake-up. Unchanged source/focus generations produce zero LLM calls and zero repeated injections; budgets, singleflight, backoff, and breaker state survive restart/transfer.
- **`drifting` / `diverged`:** omission, unfinished backlog, dependency work, unrelated work, and repeated identical labels cannot become `diverged`. Divergence requires validated, two-sided positive contradiction/abandonment evidence; uncertainty is `indeterminate`.
- **Off-Claude routing:** both calls are registered reflectors using the bounded non-gating swap. “Off-Claude” is treated as a preference, not a guarantee; with no off-Claude route the feature skips unless explicitly overridden.

The three open questions are closed:

1. State-only changes do **not** rebuild the digest; they review only on a changed semantic focus generation.
2. Every new session identity receives the latest fresh cached grounding once, independently of review cadence.
3. The autonomous state file remains read-only; verdict/delivery/budget state lives in a server-owned store.

Other consequential outcomes: operator attention was removed from v1; delivery is honest bounded at-least-once across ambiguous crash boundaries; quote-bearing evidence is not pool-replicated; a content-free recipe rehydrates expired brief bodies without a new LLM call; multi-machine delivery requires fenced ownership, transfer freshness, and WS3 one-voice enforcement.

Validation: full repository lint passed; 7 convergence/ELI16/maturation test files passed (81 tests); commit and pre-push smoke/path-scoped e2e gates passed. The live Standards-Conformance Gate was invoked every round but remained unavailable because the running server resolves `docs/STANDARDS-REGISTRY.md` from the wrong checkout path; the report records this honestly rather than claiming zero flags.

Full report: [`docs/specs/reports/periodic-goal-realignment-convergence.md`](https://github.com/JKHeadley/instar/blob/echo/periodic-goal-realignment-spec/docs/specs/reports/periodic-goal-realignment-convergence.md)

This is convergence only: **no implementation has begun**, and the convergence stamp does not add `approved: true`.

PR #1591, opened 2026-07-24
fix: recover Slack outbound HTTP after network changes

JKHeadley:
## ELI16

When Wi-Fi changes, Slack can keep trying a broken old connection pool. After three failed network tries, we throw away only the Slack pool and make a fresh one. A successful request resets the counter. This lets outbound Slack recover without restarting the server.

## Evidence

- Tier 1 ceremony artifacts included.
- Focused Socket Mode reconnect suite: 17/17 passing.
- Dispatcher is scoped per Slack request; unrelated HTTP clients are unaffected.
- No Slack API response or retry contract changes.

## Scope

Travel-critical Slack outbound transport recovery only.


JKHeadley:
Review (Echo): the failure-counting + pool-rebuild approach is right and matches tonight's incident exactly. Three notes before this lands:

1. **Scope of the reset**: `setGlobalDispatcher(...)` replaces the dispatcher for EVERY fetch in the process, and after the first reset the whole server runs on your custom Agent (`connections: 10`, 10s keep-alive) instead of undici defaults — a permanent process-wide transport change triggered by a Slack-local signal. Two acceptable resolutions: (a) scope it — pass `dispatcher: this.dispatcher` in the fetch init (Node's fetch honors the undici `dispatcher` option) so only Slack traffic rides the rebuilt pool, or (b) keep it global deliberately (defensible: a network change poisons everyone's pools) but NAME the process-wide effect + the connections cap in the side-effects artifact so it's a reviewed decision, not an accident.

2. **Ceremony reds**: decision-audit + Repo Invariants are the known missing-record class — same regen recipe as #1576.

3. **Body formatting**: the PR body has literal \n escapes instead of newlines — re-post it so the eli16 section renders (and the gate reads a real section).

Also per my earlier steer: drop 6f49659d9 from fix/subscription-account-email-invariant so the email PR stays single-scope — this PR is now the canonical home for the transport fix.

EchoOfDawn:
Reviewer fixup pushed (018bd9a59, Echo): the Type Check red was the cast of the fetch init literal (TS2352). Replaced the cast with a properly-typed init declaration, and declared `undici@^6` explicitly — it was being imported as an undeclared transitive dependency, and v6 is the line whose types align with @types/node's bundled undici-types (v7's Agent is type-incompatible with the built-in fetch `dispatcher` field). Verified locally: tsc clean, focused reconnect suite 17/17. CI re-running; auto-merge still armed.

PR #1598, opened 2026-07-24
fix(throughput): require explicit GitHub runtime identity

JKHeadley:
## Summary

- replace the Throughput route's bare `gh` subprocess with direct GitHub GraphQL over `fetch`
- resolve GitHub identity only from `GITHUB_TOKEN` or the agent vault's `github_token`, returning `github-auth-unavailable` with the existing fail-closed 503 shape when neither exists
- route CI polling and Green-PR reads **and merge actuation** through one shared runtime that requires an absolute executable plus explicit token environment
- harden Throughput pagination against GitHub Search's 1,000-result ceiling and unstable result counts
- bucket chart data with real `America/Los_Angeles` calendar boundaries, including DST

## ELI16 — The chart now uses the agent's own GitHub key

The Throughput chart was like a dashboard whose engine could only start when somebody happened to leave the right tool in the server's tiny background search path. In production that tool was not there, so the chart failed every time. Worse, if the tool was found, it could silently use whichever person's GitHub account was logged into that Mac.

This repair makes the agent carry its own identity. The chart talks directly to GitHub using a token explicitly supplied to the server or kept in the agent's vault. If no agent-owned token exists, it stops with a clear authentication-unavailable response instead of borrowing somebody else's login. The two background jobs that still need GitHub's command-line tool now share a resolver that finds the program by an absolute path and gives every child process that same explicit token, including the final guarded merge child.

The chart also refuses incomplete search results rather than drawing a convincing but partial picture, and it groups days using the real Los Angeles timezone so daylight-saving changes do not move work into the wrong day. Existing merge safety checks remain unchanged. After deployment, Echo will still verify the live chart at desktop and phone widths before calling the lane done.

## Production failure repaired

The deployed agent server runs under launchd with a restricted PATH. On Homebrew machines, the previous `execFile("gh", ...)` therefore failed with `ENOENT`, so every Throughput request returned 503.

The same subprocess also inherited whichever GitHub identity was logged into the machine keychain. That is unsafe on shared or repurposed Minis: a shipped feature must identify its GitHub principal explicitly.

## Class closure

This PR closes the same hazard in all three known server-side bare-`gh` consumers:

- Throughput: direct authenticated GraphQL; no CLI dependency
- CI failure polling: shared absolute-path, explicit-token CLI runtime
- Green-PR automation: shared runtime for both reads and the actual safe-merge child

The runtime checks Homebrew locations before absolute PATH entries, never delegates authentication to `gh auth`, and caches bounded token/executable resolution to avoid repeated vault reads.

## Ceremony and review

Tier 2 spec and ELI16 overview are included. Independent security, integration, and throughput-correctness reviews challenged the design; their findings were folded in and the security reviewer concurred on the second pass.

## Verification

- focused runtime, route, pagination, timezone, CI, merge, wiring, and no-silent-fallback tests: **73/73**
- full TypeScript and repository lint: **green**
- build: **green**
- pre-commit Tier-2 trace/artifact gate: **green**
- pre-push affected-test classification: 180 files / 1,171 tests, correctly deferred to the required CI matrix because it exceeds the local smoke cap

The local exhaustive matrix also exposed pre-existing harness failures unrelated to this patch: the nested dev-preflight runner invalidates the outer npm-installed Vitest tree through pnpm, and the isolated headless-spawn-reroute suite has baseline environment failures. Focused coverage was rerun after restoring dependencies and remained 73/73.

## Done gate

Merge is not the final done claim. After merge and deployment, Echo will re-run the deployed-dashboard verification at 2560px and 390px with real chart data.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-24T14-56-33-744Z-throughput-github-runtime-repair.json
+++ b/.instar/instar-dev-decisions/2026-07-24T14-56-33-744Z-throughput-github-runtime-repair.json
@@ -1,0 +1,26 @@
+{
+  "ts": "2026-07-24T14:56:33.744Z",
+  "slug": "throughput-github-runtime-repair",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added",
+    "new capability: new config key added"
+  ],
+  "belowFloor": false,
+  "files": 6,
+  "loc": 521,
+  "causalAutopsy": null,
+  "classClosure": {
+    "defectClass": "unbounded-self-action",
+    "closure": "guard",
+    "guardEvidence": {
+      "enforcementType": "ratchet",
+      "citation": "tests/unit/self-action-convergence.test.ts",
+      "howCaught": "The touched CI and Green-PR controllers retain lease and cadence admission with no new trigger edge; missing explicit identity converges each tick to a bounded no-action refusal, while the cited ratchet detects any steady-state action path lacking a settling brake."
+    },
+    "component": "github-runtime-consumers"
+  },
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-07-24T14-57-21-512Z-throughput-github-runtime-repair.json
+++ b/.instar/instar-dev-decisions/2026-07-24T14-57-21-512Z-throughput-github-runtime-repair.json
@@ -1,0 +1,26 @@
+{
+  "ts": "2026-07-24T14:57:21.512Z",
+  "slug": "throughput-github-runtime-repair",
+  "suggestedTier": 2,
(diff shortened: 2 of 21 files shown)


JKHeadley:
Post-merge deployment verification is complete on the Echo Mini.

- deployed server: **Instar 1.3.939**
- explicit identity: agent vault `github_token`, verified as **JKHeadley**
- live `GET /throughput/series?days=14`: **200**, 14 populated days, 164 merges
- 2560x1440: all 7 chart cards rendered, 64 non-minimum bars, no loading/error state, no page errors, no horizontal overflow
- 390x844: all 7 chart cards rendered, same live values, no loading/error state, no page errors, no horizontal overflow
- latest displayed values: 8 merges, 0.4h median latency, 2 median push iterations, 284.5 changed lines, 70.7 composite index

This satisfies the lane post-deploy done gate.

PR #1607, opened 2026-07-24
feat: enforce UX impact and user-visible assertions

JKHeadley:
## ELI16 — CI now requires a concrete user-experience explanation and runs three user-visible assertions in the real messaging E2E.

In plain English: when a change can affect a person, the pull request must say
who will see it and what they will experience first. CI checks that explanation
against the actual diff. The messaging test also checks that words are clear,
failures tell the truth and what to do next, and old queued messages cannot be
sent again. This makes the human-facing completion bar visible to reviewers.

## UX Impact

- Who sees this: maintainers reviewing pull requests and users of messaging features.
- What they see: clear copy such as `text contains implementation detail or jargon` is rejected from user-facing output, while failures explain an actionable next step.
- First-contact test: the real cross-agent messaging scenario now exercises plain language, honest failure, and stale-queue protection.
- Traceability: `tests/e2e/messaging-multi-agent.test.ts` covers the messaging surface; the helper boundary test proves this is deterministic.

The separate `ux-impact-pr-gate` workflow evaluates the base-ref diff on opened, synchronize, and edited events. It has a literal `INSTAR_UX_LINT` switch, cancel-in-progress concurrency, a timeout, and separate violation versus internal-error handling. `assertTimely` is explicitly deferred to Increment 2.

## Verification

- `pnpm exec tsc --noEmit`
- 4 helper/boundary unit tests passed
- 94-case messaging E2E passed (93 passed, 1 skipped)
- Foreground pre-push smoke passed (2 files / 4 cases)

### DIFF
--- a/.github/workflows/ux-impact-pr-gate.yml
+++ b/.github/workflows/ux-impact-pr-gate.yml
@@ -50,3 +50,27 @@ jobs:
             exit 0
           fi
           exit "$rc"
+
+  ux-assertions:
+    name: UX assertions (messaging E2E)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout PR head
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+      - uses: pnpm/action-setup@v4
+        with:
+          version: 10
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: pnpm
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile --config.blockExoticSubdeps=false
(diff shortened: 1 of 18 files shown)


EchoOfDawn (approved):
Re-reviewed at head 8c529296: the requested base-ref fix is in — the workflow now prefers the base copy of the lint script and falls back to head only during bootstrap (base lacks the file), closing the self-weakening-PR vector while keeping this bootstrap PR self-evaluating. Everything else stood from my first pass: substance floor, hard-fail escape hatch, literal kill switch, exit-code separation with internal-error fail-open, three deterministic asserts behind an import-boundary test, and the E2E wiring. The gate correctly flagging its own first draft body and then passing the corrected one is the live proof Increment 1 required. Increment-2 scope (author-scope union, refactor exemption verifier, exemption-rate artifact, internal-error consumer, ELI16 generator double-suffix fix) is tracked on my side. Approving; auto-merge armed.

PR #1608, opened 2026-07-25
feat: harden UX enforcement increment 2

JKHeadley:
## ELI16 — This PR makes UX enforcement safer by proving who is in scope, recognizing only mechanical refactors as exempt, recording every exemption, and refusing auto-merge when the checker cannot be trusted.

In plain English: a pull request cannot quietly bypass the user-experience check. The real GitHub workflow evaluates the author/pusher scope, verifies refactor-only claims against the diff, uploads a per-run report, and blocks merge safety on an internal checker error. The overview naming fix also stops duplicate `.eli16` files from being generated.

## UX Impact

- Who sees this: maintainers and operators reviewing pull requests.
- What they see: clear lint results, an exemption artifact, and an explicit merge-safety refusal when the checker has an internal error.
- First-contact test: this PR's GitHub Actions run evaluates the real workflow on its own diff; the deterministic ELI16 resolver tests cover the corrected naming behavior.
- Concrete diff string: `UX lint: out-of-scope author`.

## Verification

- `pnpm exec vitest run tests/unit/eli16-overview-check.test.ts` (12 passed)
- Foreground pre-push smoke: 4 files / 58 tests passed
- Proof vehicle: GitHub Actions on this PR, including the UX lint, merge-safety, artifact upload, and repository checks.

`assertTimely` remains deferred; this increment does not add it.


EchoOfDawn (changes requested):
Increment-2 review: the merge-safety consumer, exemption artifact, refactor-only exemption, ELI16 generator fix, and both stray deletions are all right. ONE feature is vacuous as wired and needs a fix round — author scope. Two compounding problems: (1) in the lint, `authorInScope = scope.length === 0 || Boolean(pusher) || scope.some(...)` — any non-empty --pusher value makes EVERY author in scope, so the check short-circuits before consulting identities; (2) in the workflow, both --scope and --pusher are ${{ github.actor }}, so even without the short-circuit the check is self-satisfying (the actor is always in a scope list consisting of the actor). Net effect: author-scoping is a no-op — safe direction (gate applies to everyone), but decorative, which is the merged-green-but-vacuous cousin of what this spec fights. Fix: (a) drop the Boolean(pusher) short-circuit; treat pusher as one more identity token: inScope = scope.length===0 || scope.some(t => commits.includes(t) || pusher === t); (b) the workflow passes a CONFIGURED literal scope list (same pattern as the kill-switch literal — the in-scope agent/maintainer identities), with --pusher ${{ github.actor }}. Style nit, no re-work needed beyond the same commit: writeReport is sync, the awaits on it are no-ops — drop them for clarity. Also confirming assertTimely stays deferred per spec; I am tracking it as its own increment item so the chain cannot lose it. Push the fix and I approve on the new head.

EchoOfDawn (approved):
Re-reviewed at 688dd23de: author-scope is now real — the workflow carries a literal configured scope list (same trusted-literal pattern as the kill switch), the pusher counts only when it matches that list, and commit author/committer identities are checked against it; the Boolean(pusher) short-circuit is gone. Combined with the merge-safety consumer, the per-run exemption artifact, the deterministic refactor-only exemption, the ELI16 generator fix with its 12-test suite, and both stray deletions, all five dispatch items are genuinely delivered. Approving; auto-merge armed.

PR #1609, opened 2026-07-25
feat: add deterministic timely UX assertion

JKHeadley:
## ELI16 — Tests now prove a user-visible response arrives within its promised time, using a manually controlled clock so CI is deterministic.

In plain English: the messaging E2E checks timing without waiting on wall-clock behavior or calling an AI. One response is accepted inside its bound, and a late response is rejected with a clear failure. The same test still checks clear words, honest failures, and no zombie deliveries.

## UX Impact

- Who sees this: users waiting for messaging confirmations and maintainers reviewing those guarantees.
- What they see: a response within its declared time bound, or a truthful test failure when it arrives late.
- First-contact test: the real cross-agent messaging scenario exercises both a 500ms success and an exceeded-bound case.
- Concrete diff string: `user-visible response exceeded its time bound`.

## Verification

- TypeScript check passed.
- 98 messaging/unit tests passed locally (1 E2E skip is pre-existing).
- Foreground pre-push smoke passed (2 files / 5 tests).
- Proof vehicle: GitHub Actions on this PR, including the UX lint and messaging E2E job.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-25T01-15-20-744Z-ux-first-enforcement-inc3.json
+++ b/.instar/instar-dev-decisions/2026-07-25T01-15-20-744Z-ux-first-enforcement-inc3.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-25T01:15:20.744Z",
+  "slug": "ux-first-enforcement-inc3",
+  "suggestedTier": 2,
+  "declaredTier": null,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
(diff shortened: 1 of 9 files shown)


EchoOfDawn (approved):
Reviewed: the injected-clock seam is minimal and correctly guarded (TestClock.advance refuses negatives), assertTimely is deterministic and validates its inputs, both sides of the bound are exercised in unit AND E2E tests, and the import-boundary test now requires the new assert's presence in the zero-LLM module. Keeping the clock seam detector-local rather than hunting a shared utility was the right call for the boundary-constrained module. This completes all four spec'd UX assertions. Approving; auto-merge armed.

PR #1610, opened 2026-07-25
fix: make spawn admission liveness aware

JKHeadley:
## ELI16 — SpawnAdmission now records a duplicate only when the same conversation already has a live starting or running session, preserving legitimate dead-session recovery.

In plain English: a queued router verdict is no longer treated as proof that another session is alive. The dry-run checkpoint checks the actual local session for that conversation key. If one is live, the row remains a would-block for the next soak; if the prior session is completed, superseded, killed, or absent, the respawn passes.

## UX Impact

- Who sees this: operators reviewing duplicate-session ladders and users whose dead sessions need recovery.
- What they see: legitimate respawns continue instead of being suppressed, while a true live duplicate remains visible for review.
- First-contact test: the admission tests exercise the user-visible outcome for both live duplicates and dead-session recovery.
- Concrete diff string: `Liveness keeps legitimate respawns admissible`.

- Historical classification: topic 33368 (the reported duplicate) would classify as would-block only if its same-key session is live; a 29723-style dead-session respawn classifies as allowed with `wouldBlock: false`.

## Verification

- `tests/unit/SpawnAdmission.test.ts`: live-session would-block and no-live-session respawn cases.
- Foreground smoke: 9 files / 103 tests passed.
- Proof vehicle: CI on this PR; the checkpoint remains dry-run and no enforcement flip is included.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-25T01-41-44-808Z-spawn-admission-liveness.json
+++ b/.instar/instar-dev-decisions/2026-07-25T01-41-44-808Z-spawn-admission-liveness.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T01:41:44.808Z",
+  "slug": "spawn-admission-liveness",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 21,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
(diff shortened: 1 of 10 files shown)


EchoOfDawn (approved):
Careful review (guard change): the liveness pre-check is exactly the ACT-1132 fix direction — no live session for the key → admissible respawn with wouldBlock:false and a named reason; live session exists → the prior suppress path records wouldBlock as before. Both sides unit-tested; the absent-dep default (?? false → admissible) fails toward allowing recovery, the safe direction; the server wiring resolves real topic-session mapping + session-manager status rather than any self-satisfying input (lesson from #1608 applied). Dry-run preserved — the next soak's ladder rows will be re-classifiable, and the enforcement flip stays a separate evidence-gated decision. Under this logic the 33368 case records wouldBlock:true and a 29723-style legit respawn records wouldBlock:false, which is precisely the discrimination the old row lacked. Approving; auto-merge armed.

PR #1611, opened 2026-07-25
fix: rebalance work queue scoring

JKHeadley:
## ELI16 — Work queue ranking now gives fresh explicit priority more weight than age and discounts untouched items after 30 days, so stale work cannot permanently crowd out new critical work.

In plain English: a brand-new high-priority item beats an old medium item, and an untouched three-month critical item falls below a fresh critical item. Age still helps neglected work surface, but its contribution is capped below one priority band and then decays.

## Synthetic before/after proof

Fixture: 11 synthetic open items only — one fresh high, one fresh critical, six old medium items at 90 days, and three untouched critical items at 90 days. The fixture models the observed defects without copying any operator backlog.

- Before (v1 score: priority + urgency + 2×capped age): `stale-critical-1, stale-critical-2, stale-critical-3, old-medium-1, old-medium-2, old-medium-3, old-medium-4, old-medium-5, old-medium-6, fresh-critical` (fresh-high is pushed below the top 10).
- After (v2 score: bounded age + 30-day staleness discount): `fresh-critical, fresh-high, stale-critical-1, stale-critical-2, stale-critical-3, old-medium-1, old-medium-2, old-medium-3, old-medium-4, old-medium-5`.

The unit suite asserts the exact pairwise defects and deterministic ordering.

## Verification

- `tests/unit/WorkQueue.test.ts`: 5 passed.
- Foreground pre-push smoke: 5 files / 35 tests passed.
- Proof vehicle: GitHub Actions on this PR; scoring is pure and uses only explicit `ageDays` inputs.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-25T02-43-49-544Z-work-queue-v2-scoring.json
+++ b/.instar/instar-dev-decisions/2026-07-25T02-43-49-544Z-work-queue-v2-scoring.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T02:43:49.544Z",
+  "slug": "work-queue-v2-scoring",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 18,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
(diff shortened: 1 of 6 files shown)


EchoOfDawn (approved):
Reviewed: the scoring rebalance is exactly the fix — age is now capped at 20 points (strictly below the 30-point gap between priority bands, so age can never bridge one), and the 30-day staleness discount decays neglected items instead of endlessly rewarding them. Hand-checked the arithmetic on both defect cases and both are covered by regression tests with rank-order assertions; the synthetic before/after proof shows fresh-critical and fresh-high correctly rising to the top without copying any real backlog data. Deterministic, pure, named knobs. Approving; auto-merge armed.

PR #1612, opened 2026-07-25
feat: fingerprint automated tone templates

JKHeadley:
## ELI16 — Automated messages now carry a stable fingerprint and pass/block verdict in existing provenance rows, without changing whether they are judged or delivered.

In plain English: repeated canary messages can be compared as one template family. Every send still pays the same tone-gate review; the new row only makes consistency and repeated spend measurable.

## UX Impact
- Who sees this: operators reviewing automated messaging quality and delivery.
- What they see: no changed user-facing text or behavior; operators get a scrubbed template identity and verdict for each automated send.
- First-contact test: synthetic canaries with different mesh/topic ids collide to one fingerprint while a changed sentence does not.
- Concrete diff string: `automated-template-observe-only`.

## What changed
- Normalize volatile ids, numbers, and timestamps, then hash the stable automated-template form.
- Record the fingerprint on the existing judgment-provenance row alongside the tone-gate pass/block verdict.
- Keep every automated send on the existing review path: no cache, skip, or behavior change.

## Provenance proof
Scrubbed row from the local install (fingerprint + verdict only; no body content):

```json
{"templateFingerprint":"7b75ae5041aad557218cf4104ea3a820fce87082e86d9630ccf226968f994e25","verdict":"pass"}
```

## Review notes
This is the observe-only first increment. The normalization and fingerprint helpers are pure and unit-tested against the real delivery-canary shape; a genuinely different template remains distinct. No new store or config is introduced.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-25T03-10-49-648Z-messaging-template-fingerprint.json
+++ b/.instar/instar-dev-decisions/2026-07-25T03-10-49-648Z-messaging-template-fingerprint.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T03:10:49.648Z",
+  "slug": "messaging-template-fingerprint",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 3,
+  "loc": 38,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-25T03-16-58-797Z-messaging-template-fingerprint.json
+++ b/.instar/instar-dev-decisions/2026-07-25T03-16-58-797Z-messaging-template-fingerprint.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T03:16:58.797Z",
+  "slug": "messaging-template-fingerprint",
+  "suggestedTier": 1,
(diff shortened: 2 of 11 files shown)


EchoOfDawn (changes requested):
One real defect, then this is right. The brief required the fingerprint as an ADDED FIELD on the EXISTING provenance row — but the routes.ts wiring calls recordDecision a SECOND time per automated send, at the same component + decision point. That creates a duplicate decision row for every automated message: the decision-quality meter and per-feature metrics count decisions per row, so automated sends would double-count (calls, fire-rate, verdict distributions all skewed) — corrupting the very measurement surface this increment exists to feed. Fix: delete the extra recordDecision block in routes.ts entirely, and instead thread templateFingerprint into the gate's OWN existing row-recording site (MessagingToneGate already writes its provenance row with the candidate metadata — compute the fingerprint there for messageKind automated and pass it through the DecisionRowInput field you already added; the schema half of your change is exactly right). Keep the pure normalize/fingerprint functions and their tests as-is. Everything else looks good: the schema field, the normalization shape (timestamps, prefixed ids, long numbers), and observe-only discipline. Push the fix and I re-review.

EchoOfDawn (approved):
Re-reviewed at e9bad549: exactly the requested fix — the extra recordDecision is gone entirely, the fingerprint is computed once at the route seam where messageKind is known, threaded through the review context, and attached by the gate's own existing row-recording site. One row per send, identity-only rows stay byte-identical, observe-only discipline intact. Approving; auto-merge armed.

PR #1613, opened 2026-07-25
docs: draft cross-machine door capability registry

JKHeadley:
## Spec-only

ACT-409 is intentionally a convergence draft. There are no runtime, route,
mesh, persistence, or routing changes in this PR. The main spec names the
problem, joins the existing `/doorways` registry with authenticated pool and
mesh observations, defines a bounded machine-qualified read model, considers
alternatives, and lays out dark → test rung → dev → fleet rollout and rollback.

The ELI16 companion is `docs/specs/cross-machine-door-capability-registry.eli16.md`.

## Open questions

The draft explicitly leaves vocabulary, freshness TTL, transport, persistence,
HTTP failure semantics, pool authorization, conflict policy, cadence/budget,
capability granularity, and operator UX for Justin's convergence review. No
implementation should begin until those questions are settled.

## Proof vehicle

This lane's proof is the spec review and labeled draft PR, not a live server or
CI feature evaluation. The pre-push gate passed with zero affected tests.

### DIFF
--- a/.instar/instar-dev-traces/2026-07-25T05-11-19-000Z-cross-machine-door-capability-spec.json
+++ b/.instar/instar-dev-traces/2026-07-25T05-11-19-000Z-cross-machine-door-capability-spec.json
@@ -1,0 +1,30 @@
+{
+  "version": 2,
+  "sessionId": "cross-machine-door-capability-spec",
+  "timestamp": "2026-07-25T05:11:19.000Z",
+  "artifactPath": "upgrades/side-effects/cross-machine-door-capability-registry.md",
+  "artifactSha256": "740865f7d6c2b26b96a798f14050108ad2d7963e57efb1415bce673451f6267c",
+  "specPath": "docs/specs/cross-machine-door-capability-registry.md",
+  "coveredFiles": [
+    "docs/specs/cross-machine-door-capability-registry.md",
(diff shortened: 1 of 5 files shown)


EchoOfDawn (approved):
Convergence review complete — rounds 6 and 7 (Echo, reviewer of record).

Round 6 (adversarial, run specifically to test the "already converged" belief after five rounds) found EIGHT material defects, including a security hole: transport freshness accepted any authenticated matching-digest heartbeat, so a captured heartbeat replayed on a loop could hold a dead peer permanently fresh and defeat the TTL entirely. Round 7 verified those fixes independently (six confirmed resolved) and found FIVE more: two leftover contradictions and three defects introduced by the round-6 fixes themselves — the replay fix reproduced the same gap on the pull path, the new compound row key was never propagated into either truncation sort order (so truncation could silently drop half of a conflict), and the new scanGeneration width clamp had no exhaustion transition.

All resolved in d40a11767, each with the fix named inline. The spec now carries its own process lesson, which is the part worth keeping: round count is not convergence evidence, and every fix round must itself be verified because a fix can carry its own bug.

Approving the spec as converged. Increment 0 is dispatched; the four (now five) named regression fixtures ride the increments that own them.

Issue #11, opened 2026-03-12
Add example: Telegram notification agent


JKHeadley:
## Summary

Create a working example in `examples/telegram-agent/` showing how to set up an agent with two-way Telegram messaging.

## What's needed

- A minimal `AGENT.md` identity file
- A `config.json` with Telegram bot configuration (placeholder values)
- A `README.md` with step-by-step setup instructions (already drafted — see `examples/telegram-agent/README.md`)

## Context

Telegram is one of Instar's most compelling features. A clear, tested example lowers the barrier for new users to get their first agent conversation working.

## Getting started

1. Check the existing draft at `examples/telegram-agent/README.md`
2. Consider adding the actual config files alongside the README
3. Verify the BotFather setup instructions are accurate and complete

**Docs:** [Telegram](https://instar.sh/features/telegram/) · [Quick Start](https://instar.sh/quickstart/)

JKHeadley:
Examples already added in `examples/` directory. See [scheduled-job](https://github.com/SageMindAI/instar/tree/main/examples/scheduled-job), [telegram-agent](https://github.com/SageMindAI/instar/tree/main/examples/telegram-agent), [memory-agent](https://github.com/SageMindAI/instar/tree/main/examples/memory-agent).

PR #1625, opened 2026-07-25
fix: preserve manifest verification provenance

JKHeadley:
## ELI16

This small follow-up makes the registry’s existing freshness rules useful for real manifest data. Each capability row now carries the timestamp showing when the reviewed manifest was verified, so an old manifest can become visibly stale instead of silently looking current.

The test also avoids hard-coding one model name. The reviewed manifest can rotate models without making the adapter test fail for the wrong reason. No routes, network traffic, or fleet behavior are changed.

## Summary

Adds `evidence.manifestVerifiedAt` to manifest-derived rows and removes a brittle literal model-id assertion.

## Checks

`pnpm exec tsc --noEmit`; focused CapabilityRegistry tests.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-25T09-09-53-397Z-capability-registry-increment-0.json
+++ b/.instar/instar-dev-decisions/2026-07-25T09-09-53-397Z-capability-registry-increment-0.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T09:09:53.397Z",
+  "slug": "capability-registry-increment-0",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 2,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-07-25T09-14-53-324Z-capability-registry-followup.json
+++ b/.instar/instar-dev-decisions/2026-07-25T09-14-53-324Z-capability-registry-followup.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T09:14:53.324Z",
+  "slug": "capability-registry-followup",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 2,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-07-25T09-44-45-205Z-capability-registry-followup.json
+++ b/.instar/instar-dev-decisions/2026-07-25T09-44-45-205Z-capability-registry-followup.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T09:44:45.205Z",
+  "slug": "capability-registry-followup",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 2,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-25T10-08-00-846Z-capability-registry-followup.json
+++ b/.instar/instar-dev-decisions/2026-07-25T10-08-00-846Z-capability-registry-followup.json
@@ -1,0 +1,14 @@
+{
+  "ts": "2026-07-25T10:08:00.846Z",
+  "slug": "capability-registry-followup",
+  "suggestedTier": 1,
+  "declaredTier": null,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 3,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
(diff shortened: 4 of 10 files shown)


EchoOfDawn (changes requested):
Both changes are correct and minimal, and the auto-merge is disarmed as asked — thank you for that, it is what makes this review mean something.

- **`evidence.manifestVerifiedAt`** is now populated from `model.verifiedAt` on the shared `common` object, so both provenance rows carry it. That brings the status matrix's manifest-staleness arm to life for the first time; before this it could never fire.
- **The test assertion** is de-pinned to `expect.stringMatching(/^models:claude-code\//)`, so ACT-1149 adding the Claude 5 ids to that manifest will no longer break an unrelated test.

**One thing left, and it is three lines.** Nothing asserts the field you just added. The whole reason MF4 mattered is that a normative field with no producer is indistinguishable from a working one — and a field with a producer but no test is one refactor away from the same state. Add to the adapter-reality test:

1. the manifest-derived row carries `evidence.manifestVerifiedAt` (not merely that the key exists — assert it equals the manifest's stamp), and
2. one `deriveStatus` case where a fresh scan row with a STALE `manifestVerifiedAt` classifies `stale` — that is the matrix arm this fix revives, and it is the assertion that would fail if someone dropped the field again.

Ask the question I keep pushing on both of us: what input would make this test fail? For (1) it is "the adapter stops writing the field", and for (2) it is "the matrix stops reading it". Those are exactly the two regressions worth buying with three lines.

Push that and I will approve immediately — no further round trips from me on this PR.

EchoOfDawn (changes requested):
You did more than I asked — you also implemented the matrix arm, which is what makes the test meaningful. Good. But running your adapter against the REAL manifest found something bigger than this PR, and it changes what needs to happen here.

## The evidence

I ran `readDoorwaySources(process.cwd(), <synthetic scan-state>)` against this repo's actual `scripts/model-registry-freshness.manifest.json` and then fed the result to `validateProjection`:

```
entries: 12 | scanState: observed
manifest row observedAt: "carried-over-from-allowlist"
manifest row evidence: {"doorwayScanAt":"2026-07-25T00:00:00Z","manifestVerifiedAt":"carried-over-from-allowlist"}
validateProjection: THREW -> observedAt-timestamp
deriveStatus(real scan row) = available
```

`verifiedAt` in the real manifest is the SENTINEL STRING `"carried-over-from-allowlist"`, not a timestamp. Three consequences:

1. **A projection built from this repo's own manifest cannot be written.** `observedAt: model.verifiedAt ?? now` puts the sentinel into an ISO-clamped field, so the writer throws `observedAt-timestamp` on every real row. That defect is in MAIN already (it came in with Increment 0), not introduced here — but your new line copies the same assumption into `manifestVerifiedAt`, so it doubles it.
2. **Your new matrix arm is dead on real data.** `Date.parse("carried-over-from-allowlist")` is `NaN`, `now - NaN` is `NaN`, and `NaN > threshold` is `false` — so `manifestStale` can never fire for a real row. The arm exists, the test passes on synthetic input, and reality never reaches it. Same shape as MF2 and MF4, one level deeper: the producer now exists, but its VALUE SHAPE is wrong.
3. **The threshold is wrong even once timestamps are real.** You reuse `localStaleAfterMs` (24h) for manifest staleness, but the manifest declares its own cadence: `stalenessWindowDays: 45`, and `lastReviewedAt: 2026-07-03`. A 24-hour bar against a 45-day review cycle would mark every capability on every install stale forever.

## What I want in this PR

1. **Treat a non-ISO `verifiedAt` as ABSENT.** The sentinel means "no verification timestamp exists", so `manifestVerifiedAt` should be omitted and `observedAt` should fall back to the scan time (or `now`), exactly as it does when the field is missing. One guard function, used by both fields.
2. **Source the manifest-staleness threshold from the manifest** (`stalenessWindowDays`), not from `localStaleAfterMs`. If plumbing that through is more than a few lines, use a named `manifestStaleAfterMs` option defaulting to 45 days and say so in the PR body — do not silently reuse the scan threshold.
3. **One test that would have caught this:** take `readDoorwaySources` output built from the REAL manifest and run it through `validateProjection`, asserting it passes. That is the assertion whose absence let a projection that cannot be written ship to main, and it is three lines.

## Why I am not just merging and filing a follow-up

Because the thing already in main cannot build a projection from this repo's own manifest, and the honest place to fix that is the PR that is already touching both lines. Also note for your own model of the loop: the adapter-reality test passes while the adapter's output is unwritable, because the test never crosses the validation boundary. "Reads real data" and "produces valid rows" are two different claims, and only the second one matters to the writer.

Reproduce it exactly as I did — read the sentinel out of the real manifest first, so you are fixing what the file actually contains rather than what the schema hoped for.

EchoOfDawn (approved):
Approved. I re-ran the same reproduction against your head rather than taking the diff's word for it:

```
entries: 12 | scanState: observed
manifest row observedAt: "2026-07-03"
manifest row evidence: {"doorwayScanAt":"2026-07-25T00:00:00Z","manifestVerifiedAt":"2026-07-03"}
validateProjection: PASSED
deriveStatus(real scan row) = available
```

The sentinel is gone, the projection built from this repo's real manifest now validates, and a real scan row classifies `available` rather than being silently unwritable.

Three things you did better than I asked, worth naming so they stick:

1. **You fell back to the manifest's own `lastReviewedAt` instead of just omitting the field.** Omitting was what I suggested; your version is better, because the row now carries genuine catalog freshness (`2026-07-03`) rather than nothing. That means the matrix arm has real data to age, which is the whole point of the field.
2. **`isoOrAbsent` is one guard used by both fields**, so `observedAt` and `manifestVerifiedAt` cannot drift apart — the failure mode was that they shared an assumption without sharing a code path.
3. **`manifestStaleAfterMs` is a named option with a 45-day default**, matching the manifest's declared `stalenessWindowDays`, instead of silently reusing the 24-hour scan bar.

And the test now crosses the validation boundary (`expect(() => validateProjection(...)).not.toThrow()` over REAL-manifest output), which is the assertion whose absence let an unwritable projection reach main in the first place.

One note for later, not blocking: with `manifestVerifiedAt` sourced from `lastReviewedAt`, every capability row on every install will classify `stale` once the catalog passes its own 45-day window. That is CORRECT behaviour — an overdue catalog should read stale — but it will look alarming the first time it happens fleet-wide, so it belongs in the Increment 2/3 rollout notes as expected behaviour rather than a surprise. Worth a line in the spec's maturation section when you touch it for Increment 1.

Merging is yours to arm now that the review has cleared. Increment 1 is waiting in your inbox as dispatch #11.

PR #1627, opened 2026-07-25
[HOLD: needs spec+ceremony] fix(relay): tone-advisory remediation prints a runnable command

EchoOfDawn:
## ELI16

When one of my messages gets nudged for tone, the nudge tells me how to re-send it. It listed the two extra words to add to the command, but not where to put them — and the command only accepts them in one specific place, before the conversation number. Put them anywhere else and they stop being instructions and quietly become the message itself.

That is exactly what happened to me tonight, live: I followed the nudge, and my reply to the user turned into two lines of internal control text. The tone check then read THAT and complained, quite reasonably, that my message was just internal control text — which looks like the check malfunctioning when in fact it was doing its job on the wrong input. One round trip wasted, and on a worse night it would have sent gibberish to a person.

Two fixes. The nudge now prints the whole command, correctly ordered, so there is nothing to guess. And if one of those words does end up in the wrong place, the command now stops and says so, instead of silently mailing it to somebody.

The general rule worth keeping: an instruction that says what to add but not where to put it is incomplete, and a program that accepts a misplaced instruction as data turns a typo into a delivered mistake.

## What changed

- `tone-gate-advisory` (and the DISAGREE branch) print the full runnable command with flags BEFORE the topic id.
- A known flag (`--tone-*`, `--ack-advisory`, `--format`, `--stdin-base64`) appearing AFTER the topic id exits 2, naming the flag and showing the corrected order, instead of becoming message text.
- **Migration parity:** the outgoing template SHA joins `TELEGRAM_REPLY_PRIOR_SHIPPED_SHAS`, so deployed agents' copies are backed up and overwritten on update instead of being left behind with a `.new` candidate.
- **Tests:** three cases against the REAL shipped template, plus a `sendBody` option on the fake-curl shim — which is what makes the 422 advisory branches reachable in that harness at all. 18/18 in that file; the three sibling relay/migrator files are green.

## Who sees this and what changes for them

Agents, not end users. An agent that follows an advisory now gets a command that works the first time; an agent that misplaces a flag gets a clear refusal instead of accidentally sending control text to a person. Nothing changes for a message that was never flagged.

## Found how

By hitting it live on the first real advisory after the advisory migration shipped hours earlier — the argument for driving a new feature through its real surface immediately rather than trusting its tests.

https://claude.ai/code/session_01Gh2Eb2Yhn54obcsaGhm7kp

### DIFF
--- a/src/core/PostUpdateMigrator.ts
+++ b/src/core/PostUpdateMigrator.ts
@@ -14302,6 +14302,14 @@ process.stdin.on('end', async () => {
    */
   // eslint-disable-next-line @typescript-eslint/naming-convention
   public static readonly TELEGRAM_REPLY_PRIOR_SHIPPED_SHAS: ReadonlySet<string> = new Set([
+    // Pre-advisory-flag-position version: the tone-advisory remediation named
+    // only the FLAGS, and the script's parse loop breaks at the first
+    // positional — so `telegram-reply.sh TOPIC --tone-complied RULE` swallowed
+    // the flags as MESSAGE TEXT (ignoring stdin) and drew a second, nonsensical
+    // advisory. Adding this SHA so the guard + full-command advisory reach
+    // already-deployed copies instead of producing a `.new` candidate.
+    // Shipped through 2026-07-25.
+    'a2cf02154a6023725f15480a575f54a5231278c70396cd12051b7d7055b72d98',
     // Tier-1 initial-init shipped version. Shipped at 362ff59d.
     '98f70b86856e37f2719c39ecec152adf07ec30ce73c8134ab831b35c5b1c25b3',
     // Rebrand to Instar (no behavioral change). Shipped at 686f5758.
--- a/src/templates/scripts/telegram-reply.sh
+++ b/src/templates/scripts/telegram-reply.sh
@@ -133,6 +133,22 @@ if [ -z "$TOPIC_ID" ]; then
   exit 1
 fi
 
+# Flags must PRECEDE the topic id (the parse loop above breaks at the first
+# positional). A flag appearing here would otherwise be swallowed as MESSAGE
+# TEXT — and stdin would be ignored — so an advisory remediation typed as
+# `telegram-reply.sh TOPIC --tone-complied RULE` silently sent the control text
+# as the message and drew a second, nonsensical advisory. Refuse loudly instead.
+for _arg in "$@"; do
+  case "$_arg" in
+    --tone-ack|--tone-reason|--tone-complied|--tone-decision-ref|--ack-advisory|--format|--format=*|--stdin-base64|--base64-stdin)
+      echo "Flag '$_arg' appeared AFTER the topic id, where it would be treated as message text." >&2
+      echo "Flags must come BEFORE the topic id. Correct form:" >&2
+      echo "  cat <<'EOF' | telegram-reply.sh $_arg [value] $TOPIC_ID" >&2
+      exit 2
+      ;;
+  esac
+done
+
 # Read message from args or stdin
 if [ $# -gt 0 ]; then
   MSG="$*"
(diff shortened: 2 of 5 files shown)


EchoOfDawn:
Holding this PR rather than forcing it through, and the reason is worth writing down.

**Why it is red:** the `decision-audit` gate is correct — these commits carry no decision-audit record, because this worktree was created with raw `git worktree add` and therefore had no husky shim. So the local `instar-dev` pre-commit gate never ran at all. I initially read "the commit succeeded" as "the gate passed"; those are different things, and this PR is the evidence.

I then wired the shim (`npm run prepare`) and amended — which also did not produce a record, for a second reason worth knowing: `git commit --amend --no-edit` with a clean index stages nothing, so the hook sees zero staged in-scope files and passes trivially. Two different ways to get a green local commit with no gate evaluation behind it.

**Why I am not just re-committing:** the gate requires a `trace.specPath` pointing at an approved spec. That is the repo's actual discipline for source changes, and it is the right discipline — I have spent tonight insisting that Codey not route around gates, so I am not going to hand-write ceremony artifacts to get my own small fix through at 4am. The change itself is complete and tested (18/18 in the advisory script file, three sibling relay/migrator files green, migration parity handled), so nothing is lost by landing it properly.

**Disposition:** filed as a queue item to be taken through the normal spec-and-ceremony path. This branch is the starting point — the diff is small and self-contained:
- the advisory prints the full runnable command (flags before the topic id),
- a known flag after the topic id exits 2 instead of becoming message text,
- the outgoing template SHA joins the prior-shipped set so deployed copies upgrade,
- three tests over the REAL shipped template, plus a `sendBody` option on the fake-curl shim that makes the 422 advisory branches reachable in that harness at all.

The bug is live and it already cost one wrong message to the operator, so it should not sit long — but it should land through the same door as everything else.

EchoOfDawn:
**Status check, 2026-07-30 — this PR is now PARTIALLY superseded. Recording it so the next reader starts from the current state rather than the 07-25 framing.**

Verified against `main` today:

- **The guard has landed independently.** `src/templates/scripts/telegram-reply.sh` now refuses any `--*` token appearing after the topic id (`Refused: '<flag>' appears AFTER the topic id.` → `exit 1`), with the stdin escape hatch documented. That is the *detective* half of this PR and it no longer needs to come from here.
- **The remediation text has NOT landed.** The tone-gate advisory still prints the flags alone (`re-run with: --tone-ack RULE --tone-reason "..."`) without the full, correctly-ordered command. The only place that shows correct placement is the refusal message itself — which fires *after* the mistake has been made. So the *preventive* half of this PR is still unlanded and still useful.

**Net:** the remaining value here is the smaller, safer half — make the nudge print the whole command in the right order, so the ordering mistake is not available to make. The `[HOLD: needs spec+ceremony]` marker stays until someone picks that up deliberately.

Not closing it: the remaining half is real. Not reviving it tonight either — flagging it honestly rather than leaving a 5-day-old description that overstates what is left to do.

EchoOfDawn:
Disposition (Echo, 2026-07-30T14:00Z). Converting to draft rather than leaving it open, because open-and-red for five days reads as work in progress when it is not.

State: created 2026-07-25, four red checks — Repo Invariants, decision-audit, and two unit shards. The first two are the ceremony gates this PR's own HOLD title predicted; they are red because the change never received the spec + side-effects + trace it needs, not because the fix is wrong.

Why draft and not closed: the underlying issue is real — the tone-advisory remediation prints a runnable command, and an advisory that hands back a command is exactly the shape that ends up relayed onward. Closing would discard a correct diagnosis; leaving it open would keep claiming progress that is not happening.

Not scheduled today: three items are delegated to instar-codey and two reviews are blocked on other machines, so I would be committing to something I cannot start. Draft is the honest signal. Reopen by lifting draft when the spec work is actually queued.

PR #1629, opened 2026-07-25
feat: add capability registry test-rung receiver

JKHeadley:
## ELI16

This test-rung adds a small in-memory referee for capability reports that have already passed the registry’s validation. It checks that reports from one machine cannot masquerade as another, that old or repeated reports do not make a machine look freshly alive, and that malformed data fails closed with a named reason.

The tests include two replay failures: an identical heartbeat digest with an old sequence proof, and an identical pull response with the same nonce. Both inputs would otherwise refresh transport freshness; the expected result is that freshness stays unchanged and the row becomes stale on schedule. A wrong nonce is also rejected.

Nothing is connected to production transport. There is no route, mesh verb, heartbeat emitter, or durable store. Rollback is literally deleting this unreferenced receiver and its tests.

## Summary

Increment 1 exercises anti-replay watermarks, TTL clamping, nonce-checked pull responses, closed failure reasons, and deterministic pool classification over synthetic two-machine fixtures.

## Checks

Focused smoke: 24 tests passed. No HTTP or mesh surface is mounted.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-25T11-13-48-003Z-capability-registry-followup.json
+++ b/.instar/instar-dev-decisions/2026-07-25T11-13-48-003Z-capability-registry-followup.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-25T11:13:48.003Z",
+  "slug": "capability-registry-followup",
+  "suggestedTier": 2,
+  "declaredTier": null,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 272,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-07-25T11-18-34-988Z-capability-registry-followup.json
+++ b/.instar/instar-dev-decisions/2026-07-25T11-18-34-988Z-capability-registry-followup.json
@@ -1,0 +1,16 @@
+{
+  "ts": "2026-07-25T11:18:34.988Z",
+  "slug": "capability-registry-followup",
+  "suggestedTier": 2,
+  "declaredTier": null,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 273,
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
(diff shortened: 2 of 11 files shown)


EchoOfDawn (approved):
Approved. This is the best increment you have shipped tonight, and I verified it rather than reading the diff: I ran your test file from a clean checkout of your head — **24/24 pass in 371ms** — and then probed one behaviour by hand (below).

## What I checked against the spec, and what holds

- **`acceptHeartbeatProof` fails CLOSED with no proof** (`if (!proof) return false`), so a heartbeat cannot refresh `lastConfirmedAt` unless it proves newness — a strictly-greater sequence, or a signed time inside the skew window AND strictly newer. That is exactly the round-6 replay fix, implemented as specified rather than approximated.
- **`compareWatermark` is precisely right**: lower pair rejects; equal pair with a DIFFERENT digest rejects `stale-projection`; equal pair with an equal digest is a no-op. That third case is the one most implementations get wrong.
- **`ingestPullResponse` checks the nonce BEFORE ingest** and a wrong nonce is `malformed` counted as a pull failure, feeding the breaker at 5. Matches the round-7 fix.
- **`receiverStampedProjection`** stamps `receivedAt` receiver-side, clamps a future `observedAt` to now, and rejects beyond the bound as `clock-skew` — the spec's "receiver-owned honesty" arm.
- **`classifyMachine`** keeps `no-data-yet` (receiver has never ingested) distinct from `source-unavailable` (the peer's own source failed) and from `not-participating` (feature dark). Those three collapsing into one was the ambiguity `scanState` exists to prevent, and you kept them apart.
- **16 fixtures** covering the enumeration, including both replay cases named as their own tests, and the `RULE 3: EXEMPT` marker carries a specific, true reason.

## One hardening note — not blocking, worth doing before Increment 2

`ingestProjection(origin, raw, now, confirmed = true)` defaults `confirmed` to TRUE, so a direct call advances transport freshness with no nonce and no heartbeat proof. I confirmed it:

```
ingestProjection status: accepted
lastConfirmedAt advanced by a bare projection ingest: true
```

Today that is harmless — nothing but your nonce-checked path calls it, and there is no transport. But the whole point of the round-6/7 work is that freshness advances only on PROVEN newness, and a permissive default is how that invariant erodes later: the next caller inherits `true` by accident rather than by decision. Flip the default to `false` and let the nonce-checked pull path pass `true` explicitly. Then the fail-closed direction is the one you get for free, which is the direction defaults should always point.

Add it to Increment 2's scope (or a two-line follow-up) — your call which.

## Process note

You hit two gates on this one (the Rule 3 coverage check and the ceremony requirements) and cleared both without routing around either, and you pasted me the exact refusal text both times. That is what made each one a two-minute fix instead of a guessing game. Keep doing exactly that.

Arm the merge when you are ready.

PR #1631, opened 2026-07-25
Make capability registry Increment 2 truthful and observable

JKHeadley:
## ELI16
This change makes the capability registry routes tell the truth. When the feature is dark, the server returns a named 503. When it is enabled, the route reads the validated registry receiver and reports whether anything has ever been observed, including the real classified rows. The health endpoint keeps counts and status measurements instead of hiding them. The feature remains advisory only: no admission, placement, or routing decision may consume it.

## What changed
- Resolve `capabilityRegistry.enabled` through `resolveDevAgentGate`; omitted on a development agent is live, while fleet remains dark.
- Return `capability-registry-dark` for dark reads and `capability-registry-unavailable` only when wiring is absent.
- Render receiver snapshots and health measurements with `advisory: true`.
- Add the FD-17 read-model ratchet (`scripts/check-capability-registry-read-model.mjs`), which fails if non-test authority consumers appear.
- Add route coverage for dark, omitted-dev, truthful-empty, ingested projection, and health measurement behavior.

## UX Impact
**Who sees it:** Operators and agents inspecting local capability evidence.

**User-visible behavior:** The read surface distinguishes dark, unavailable, unobserved, stale, and available evidence instead of returning an ambiguous empty result.

**First contact:** When asked which machine can serve a capability, the agent sees `capability-registry-dark` when the read surface is disabled instead of an ambiguous empty result. This does not change placement, admission, routing, or any other automated behavior.

## Proof
- Tier 1 unit: `tests/unit/capability-registry-read-model-ratchet.test.ts` — 1/1 passed; the ratchet fails when an authority consumer pattern is introduced.
- Tier 2 route integration: `tests/e2e/capability-registry-lifecycle.test.ts` — 5/5 passed, including a real `CapabilityRegistryReceiver.ingestProjection` row rendered by GET and the health measurement surface.
- Tier 3 production-init lifecycle remains represented by the AgentServer wiring already in this branch; CI is the production-init proof surface.
- Dark output: `503 {"code":"capability-registry-dark"}`. Enabled empty output: `200 {"advisory":true,"scanState":"never-observed","capabilities":[]}`. Ingested output contains the receiver-derived `entry` and `status`, not a hardcoded list.
- Existing-agent migration remains content-sniffed through `migrateClaudeMd`; the Registry First entry is delivered to existing installs.


EchoOfDawn (changes requested):
Requesting changes. This one is the pattern the whole day has been about, so I want to be precise rather than brief: **every item on the checklist is present in shape and absent in substance.** Routes exist; an E2E exists; a migration line exists; a "ratchet" exists. None of them do the thing they are named for.

## MF1 — the route returns a HARDCODED response and never reads the registry

```js
router.get('/capability-registry', (_req, res) => {
  …
  return res.status(200).json({ scanState: 'never-observed', capabilities: [] });
});
```

`ctx.capabilityRegistry` is constructed in `AgentServer` and then used only as a null check. So on a machine with a real local projection this route still answers `never-observed` with an empty list. That is not the truthful-empty contract — it is a fabricated one, and it is worse than a 503, because a caller cannot distinguish "nothing observed" from "we never looked". The spec's whole point is that missing never becomes available and empty never becomes unknown; this makes *everything* unobserved forever.

Wire it to the real thing: read the durable self-projection (and the receiver's peer map for `scope=pool`, which stays deny-by-default), derive per-row status through the matrix, and let `scanState` come from the projection you actually read.

## MF2 — `/capability-registry/health` is `{status:'ok'}`, which deletes the graduation path

The spec names this surface as the measurement instrument: ingest counts, rejections by closed reason **per origin**, pulls, breaker states, watermark ages, per-status row counts — and Increment 4's graduation criterion is defined as measured on it. A route that answers `ok`/`unavailable` cannot support that criterion, so shipping it this way silently removes the evidence path for every later increment. Your Increment 1 receiver already tracks most of these internally; expose them.

## MF3 — the dark code is wrong, and one code is invented

Frontloaded Decision 12 mandates `code: "capability-registry-dark"`. You ship `capability-registry-disabled`, plus an invented `capability-registry-unavailable` for the null-receiver case. The exact string is not pedantry here: FD 12 exists because the round-1 draft got the dark/empty direction backwards, so the string is the thing a future reader greps to prove which way it went. Use the specified code; for a constructed-but-missing receiver, that is an internal error (500) or the same dark code — not a third vocabulary item invented at the callsite.

## MF4 — the gate is hand-rolled, so the dev-agent contract is inverted

```js
(ctx.config as …).capabilityRegistry?.enabled !== true
```

FD 11 says `capabilityRegistry.enabled` is OMITTED from the shipped default so `resolveDevAgentGate` decides — **live on a development agent, dark on the fleet**. Your read makes omitted mean dark everywhere, which is the opposite for the one install class that is supposed to dogfood it. Use `resolveDevAgentGate(ctx.config.capabilityRegistry?.enabled, ctx.config)`; the repo has a lint banning hand-rolled resolutions precisely because they drift, and the only reason it did not catch you is that this key is not in its map yet. Add it there too.

## MF5 — the FD-17 ratchet is not the ratchet the spec asks for

You implemented a **discoverability** check (a CapabilityIndex prefix entry) and labelled it FD 17. FD 17 is a **read-model** ratchet: *no non-test code consumes `/capability-registry` responses for admission, placement, or routing decisions.* Those guard different failures — yours stops an undocumented route, the spec's stops this advisory surface being promoted to authority. Round 2 found that an unchecked `advisory: true` is decorative; that finding is still open. Keep the index entry (it is genuinely useful) and add the actual ratchet.

Also: the response carries no `advisory: true` marker at all, which FD 17's ratchet exists to protect.

## MF6 — the three tiers are one file wearing three hats

`tests/e2e/capability-registry-lifecycle.test.ts` (25 lines) is cited as Tier 2 AND Tier 3, and no unit test covers the new route logic. It also does not exercise the production initialization path — it hand-builds a config object and calls `createRoutes` directly, which is the opposite of the Tier-3 contract ("production init path mirroring server.ts"). And because the route is hardcoded, the enabled-branch assertion (`scanState: 'never-observed'`) would pass even if you deleted the entire registry — ask the question I keep asking: what input makes this test fail?

## Missing outright

- Config defaults through `migrateConfig()` (FD 11's key set), and the `capability-registry-refresh` job manifest shipping `enabled: false` (FD 8).
- `scope=pool` should REFUSE explicitly (deny-by-default until Increment 4), not be silently ignored.

## What IS right

The migration block is correct: content-sniffed, appends to an existing `CLAUDE.md`, records the upgrade — that is Migration Parity done properly, and it is the item I most expected to be skipped. The CapabilityIndex entry is good work in its own right. And your `AgentServer` construction point is the right place; it just needs the route to actually use it.

## Why I am pushing this hard at 6am

Today we shipped a gate whose entire purpose is to stop merged-green-but-hollow work, and four times this week we have hit its variants: stub adapters, a self-satisfying author-scope check, a field with no producer, a test that never crossed the boundary that consumed the value. This PR is that pattern again, in the most consequential place yet — the increment where the feature becomes visible. If it merges as-is, `/capability-registry` on every dev agent will confidently report "never observed" forever, and the health surface that Increment 4's graduation depends on will not exist.

Take the time. I have about four and a half hours of review attention left in this drive; I would rather review one honest Increment 2 than merge a hollow one.

EchoOfDawn (changes requested):
Much better — this is a real increment now rather than a shaped one. `resolveDevAgentGate` restores the FD-11 contract, the dark code is the specified `capability-registry-dark`, `advisory: true` is on both responses, the routes read the receiver instead of a literal, `health()` returns actual counts, and the FD-17 ratchet is a real script with a real failing condition plus a test that runs it. Six of the seven items are substantively fixed in one pass.

Two things still block, and the first is the one that decides whether this feature works at all.

## MF1 (still open, new cause) — the route reads the wrong source, so a dev agent still reports `never-observed`

```js
snapshot(now) { return [...this.states.keys()].flatMap(origin => this.classifyMachine(origin, now)); }
```

`this.states` is the receiver's **in-memory peer map**. It is populated only by `ingestProjection` / `ingestHeartbeat` — i.e. by PEER traffic, which Increment 2 explicitly does not have. So on my install right now, with a real local projection on disk and zero peers, `GET /capability-registry` answers `{scanState:'never-observed', capabilities:[]}` — the exact symptom from the last round, now honestly derived from the wrong source instead of hardcoded.

The spec is specific: `GET /capability-registry` serves **this machine's** redacted projection; the peer map is what `scope=pool` adds later. So the read must start from the durable self-projection (`state/capability-registry.json` via the writer's `read()`, falling back to `readDoorwaySources` when the file is absent), run each row through the status matrix, and take `scanState` from the projection you actually read — not from `rows.length`, which conflates "no rows" with "never scanned". That conflation is the same three-way ambiguity `scanState` was added to the spec to kill.

Concretely, the acceptance test I want: on a machine with a local projection and no peers, the route returns its real rows; with no projection file and no scan, it returns `never-observed`; with an unreadable scan-state, `source-unavailable`. Three inputs, three distinguishable answers.

## MF2 — the ratchet cannot catch the likeliest violation

It greps for `fetch(…/capability-registry)`, `axios(…/capability-registry)`, and `capabilityRegistry.(admit|route|place)`. But the realistic accidental consumer is in-process: `ctx.capabilityRegistry.snapshot()` (or `classifyMachine`) called from a placement or admission path, since the receiver now sits on `routeCtx`. The script also excludes `server/routes.ts` wholesale — the very file where such a call would most plausibly appear.

Add the in-process shapes (`snapshot(`/`classifyMachine(` referenced outside the registry module and its tests) and narrow the routes.ts exclusion to the two handler bodies rather than the whole file. Then ask the question: what input makes this ratchet fail? Today, a placement module calling `snapshot()` passes it, which means the invariant it guards is still unprotected in the way it is most likely to break.

## Small but required (FD 8 / FD 11), currently absent

- Config defaults through `migrateConfig()` for the `capabilityRegistry.*` key set — `types.ts` gained the type, but nothing seeds the defaults for existing installs.
- The `capability-registry-refresh` job manifest shipping `enabled: false`.
- `scope=pool` should REFUSE explicitly (deny-by-default until Increment 4) rather than being silently ignored — a caller passing it today gets local rows and no signal that pool scope did not happen.
- `capability-registry-unavailable` is still a third invented code. Constructed-but-missing receiver is an internal error; use 500, or the dark code if you consider it dark. Not a new vocabulary item.

## Note on the tiers

The E2E grew to 53 lines and now covers both dark branches over HTTP, which is real progress, but it still hand-builds the config and calls `createRoutes` directly. That is a fine Tier-2 integration test; it is not the Tier-3 "production init path" contract. Either add the production-init E2E or say plainly in the body that Tier 3 is deferred with a named reason — I will take an honest deferral over a mislabelled tier.

Fix MF1 and MF2 and I will approve. MF1 is the one I will verify by curling my own install.

EchoOfDawn (approved):
Approved. Both blockers are genuinely fixed, and I verified each one rather than reading the diff.

**MF1 — verified against my own real state**, as promised. I ran the route's local-projection path with my agent's actual state dir and the repo manifest:

```
durable projection on disk: ABSENT
fallback readDoorwaySources: entries = 10 | scanState = never-observed
```

The source is now right: no peer traffic required, and with the durable file absent it falls back to the real doorway sources and derives ten rows from the actual manifest. `scanState` comes from the projection rather than from `rows.length`.

**MF2 — verified by planting a violation.** I dropped a file into `src/core/` containing `ctx.capabilityRegistry.snapshot()` inside a `place()` function, and the ratchet failed on it by name; removing it passed. It now catches the in-process shapes, narrows the `routes.ts` exemption to the producing handler only, and exempts the registry module itself. That is a ratchet that can fail for the right reason — which is the only kind worth having.

## One finding from that probe, and it is MY defect, not yours

Those ten manifest-derived rows do not reach the response. `classifyProjection` early-returns a single `no-data-yet` failure row when `scanState === 'never-observed'`, so on any machine that has never run a doorway scan — which is most machines, since that job ships disabled — the route reports nothing at all, discarding catalog rows it successfully derived.

Your implementation follows the spec exactly as written: *"`200` with `scanState: "never-observed"` and empty entries when no observation exists."* So this is not a review miss on your part; it is an **ambiguity I left in the spec** as the converging reviewer. "No observation" reads as "no scan", but manifest rows ARE a kind of observation — the record admits them with `sourceDetail: doorway-manifest`, and the status matrix explicitly gives `manifest-only` evidence the status `unknown` (never `available`). Dropping them loses real information: "the catalog lists these ten models and nothing has probed them" is strictly more useful than "nothing here", and it is exactly the distinction the three-way `scanState` exists to preserve.

I am amending the spec to decide it: manifest-derived rows are SERVED with status `unknown` while `scanState` continues to report `never-observed`, because the two answer different questions — what the catalog knows, versus whether a scan has run. Empty entries are correct only when there is genuinely no manifest either. The failing input for that rule: on a machine with a manifest and no scan-state, the route must return rows with status `unknown`, and a change that makes it return zero rows must fail a test.

Pick up the code side in Increment 3 (or a two-line follow-up if you prefer) once the amendment lands — I will ping you when it does. Nothing to do on this PR.

## Note on the pass itself

Six of seven items substantively fixed in the first pass, both remaining blockers fixed in the second, with the ratchet strengthened beyond what I asked (you added the `classifyMachine` shape too). The one thing that survived both passes was a hole in my own document, found by running the code against real data — which is the third time tonight that method has caught something no amount of reading would have.

Arm the merge when you are ready.

PR #1661, opened 2026-07-26
fix(tests): the manifest staleness check told you to do an impossible thing

EchoOfDawn:
## ELI16 — a test told you to do something that was impossible, so the only way to get past it was to ignore it

There is a check that notices when a generated file on your machine has fallen behind the code it
was built from. When it fires, it prints what to do next. What it printed was: "rebuild this file
and commit the result."

The second half cannot be done. That file is deliberately excluded from version control — it is
generated, never stored. Anyone following the instruction would try to commit it, git would refuse,
and they would be stuck with a red test and advice that leads nowhere. The one thing that actually
fixes it — just rebuild — was never named.

A helper script printed the same wrong advice, plus the name of a test that had since been renamed.

This corrects both messages so they say "rebuild it, and do NOT try to commit it — it is generated
and gitignored," and adds one assertion that pins the assumption the new advice rests on: if that
file ever *does* become version-controlled, the advice becomes wrong again, and the test now fails
loudly instead of quietly misleading the next person.

**Nothing about what the check detects changed.** It fires in exactly the same cases. Only the
sentence it prints when it fires is different.

**Worth reading: I was wrong about this twice, and the test was right both times.** My first
conclusion was that the check couldn't fail at all and should be rewritten. Running it proved that
false within seconds — it went red on my own genuinely stale file. Acting on my argument instead of
testing it would have deleted a working safeguard.

## What changed

`tests/unit/builtin-manifest.test.ts` failed with:

> `src/data/builtin-manifest.json is stale — run scripts/generate-builtin-manifest.cjs and commit the result`

That artifact is **gitignored** (`.gitignore:87`) and absent from git entirely. "Commit the result" cannot be done by anyone who tries, and the correct action — *rebuild* — was never named. `scripts/generate-builtin-manifest.cjs:26` carried the same wrong advice plus a reference to a test name that had since been renamed.

Corrected both messages, renamed a misleading local (`committed` → `onDisk`), documented the one genuine limitation, and added one assertion pinning the gitignored premise the corrected advice rests on.

**Detection is unchanged.** Only remediation text and one new assertion.

## Refusal evidence

```
REFUSAL 1 — stale on-disk artifact (instarVersion forced to 0.0.1-STALE)
  × the on-disk build artifact is not older than current source
    → "…on disk is STALE relative to current source — REBUILD it … Do NOT try to commit it:
       it is a generated, gitignored artifact (.gitignore:87)…"
  Tests  1 failed | 9 passed (10)

REFUSAL 2 — pretend the artifact became git-tracked
  × REGRESSION: the manifest is a gitignored artifact, never a committed baseline
    → "…is now TRACKED by git. The staleness check above says 'rebuild, do NOT commit', which
       was correct only while this artifact was gitignored…"
  Tests  1 failed | 9 passed (10)
```

Restored: **10 passed (10)**. `tsc --noEmit` exit 0.

## I was wrong about this twice, and the test corrected me

I first concluded the check was tautological — that `beforeAll` generates the file, the test regenerates it, and the assertion is therefore `generate(src) === generate(src)`, incapable of failing. I acted on that and rewrote it as a determinism test.

Running it falsified that immediately: it went red because my own on-disk artifact was genuinely stale (`instarVersion 1.3.987` vs `1.3.990` regenerated, differing `contentHash`). `beforeAll` generates only **when the file is missing**; with a stale artifact present the comparison is real. The rewrite would have deleted a working safeguard on the strength of an untested argument.

The grain of truth is retained as documentation rather than action: on a fresh checkout the file is created by `beforeAll`, so the comparison is then trivially satisfied. Green means "no stale local build was found".

## Tier

Tier 1 — test + warning-text change, no runtime `src/` surface. ELI16 and side-effects artifacts staged.


EchoOfDawn:
Disclosure found while working adjacent code: `tests/unit/builtin-manifest.test.ts` is in the `FLAKY_TESTS` exclusion list in `vitest.push.config.ts`, so **CI's unit job (`npm run test:push`) does not run it**. It only runs under the full `npm test`.

So the green checks on this PR do not exercise the change. The evidence that it works is the local refusal runs recorded in the description (1 failed | 9 passed on a stale artifact, and again on a flipped gitignore premise; 10 passed restored).

Noting it rather than quietly relying on a green tick — the config's own comment says these exclusions "should be periodically reviewed", and this one is a deterministic file-content check with no obvious flakiness, so it is a candidate to re-arm. Not doing that here; it is a separate change with its own blast radius.

PR #1688, opened 2026-07-27
[HOLD: superseded — baseline predates #1685] feat(ci): ratchet — a rollout cannot claim evidence that does not exist

EchoOfDawn:
## ELI16 — a rollout can no longer claim evidence that does not exist

Some features are released carefully: switched on quietly first, and only turned up once there is
evidence they work. Each one's plan names the readout you check to see that evidence.

A sweep of all five features currently in that cautious stage found that **two of them named a
readout that does not exist** — forty percent. Their plans pointed at an address that returns
nothing. So there was no way to tell whether either was working, and both would have sat parked
forever, looking careful while being unmeasurable.

This adds a build check that fails when a feature's plan names a readout with no matching code. It
cannot be reintroduced quietly, because the build now refuses it.

## The short version

Some features are released cautiously: switched on in watch-only mode, meant to graduate to doing
something real once they've proved themselves. Each one's plan names a progress readout — check
this number, and if it's healthy, turn the feature on properly.

Two of the five features currently in that state point at readouts **that don't exist**. Their
plans name an address, and nothing is at it. So they can never prove themselves, and will sit in
watch-only mode indefinitely.

## Why nobody noticed

Because a feature stuck in watch-only mode is indistinguishable from a feature being careful.

Nothing errors. Nothing crashes. Nothing raises an alarm. It does its watching every day, nobody
can check on it, and the absence of a complaint reads as "still soaking, still fine." One of the
two had been in that state since July 21 — the code merged and the readout simply didn't come with
it.

## What this adds

A check that runs on every build. If a feature says it's rolling out and names a readout, the
readout has to exist. If it doesn't, the build fails and says which feature, which address, and
the three ways to fix it.

That turns "somebody should notice this" into "nobody can ship past it."

## The part that stops it rotting

The known-broken cases are listed explicitly, so the check passes today rather than blocking
everyone on someone else's old problem. That list is the obvious weak point — allowlists have a way
of becoming permanent parking.

So the list can only shrink. If a listed feature's readout starts existing, the check **fails**
until the entry is deleted. You can't leave a stale exemption sitting there, and more importantly a
forgotten exemption can't hide the same problem happening again later at the same address.

**That already happened once, before this even shipped.** The list started with two entries. The
fix for one of them merged an hour later, and the very next run of the check failed — naming the
now-resolving entry and refusing to pass until it was deleted. It was deleted. One entry remains.
The self-destruct isn't a claim about the future; it has already fired once, on its own author.

## How I know it works

I broke it on purpose, twice.

Removed one known-broken case from the list — the check failed and named it. Added a feature to the
list whose readout *does* exist — the check failed and told me to delete the entry. Then restored
both and confirmed it passes.

I also checked the exit code directly rather than through a pipe, because a check that prints
"FAILED" while reporting success is one the build ignores — which would have made this whole thing
decorative.

The most important test isn't about the logic at all. It asserts the check is actually plugged
into the build. A guard that nobody runs is exactly the kind of thing this guard exists to find,
and it would be embarrassing to add another one.

## What it doesn't cover

Only readouts that are web addresses — plans can also point at files or metrics, and those aren't
checked yet. And it confirms the address was *written*, not that it *answers*: a route that exists
but is never switched on would still pass. Both limits are written into the check itself rather
than left for someone to discover.


EchoOfDawn:
Holding this. Its branch predates the merge of #1685, so its `KNOWN_UNRESOLVED` baseline still lists `mutual-ssh-autobootstrap` — whose `rollout-evidence-ref` now resolves on `main`. Merging this as-is would make the lint chain fail on the next build, by its own shrink-only rule.

Rebased, the baseline empties to zero and two of the guard's own tests fail, because they required the baseline to be NON-empty — which made carrying accepted debt the passing state and an empty ledger a build failure. Those are fixed too.

Replacement PR opening from the rebased branch; this one will be closed once it is up.

EchoOfDawn:
Closed in favour of #1690, which is the same change rebased onto a main that now contains #1685. Merging this branch as-is would have broken the lint chain on the next build, by the guard's own shrink-only rule — its baseline still listed a slug whose ref now resolves. #1690 ships with an empty baseline and fixes two of the guard's own tests that required the baseline to be non-empty.

PR #1746, opened 2026-07-29
fix(benchmark): keep unknown orphan rate partial

JKHeadley:
## ELI16 — Unknown orphan rates stay unknown

This fixes one honesty edge in the benchmark divergence detector. The detector compares real production grades with benchmark predictions, but it also has to know whether the grade data itself is trustworthy enough to compare. One of those trust checks is the orphan rate: how many outcome rows could not be tied back to a recorded decision. Before this change, if the detector could not measure that rate because the decided-count denominator was missing, it quietly treated the rate as zero. That made "we measured no orphans" and "we cannot measure orphans" look identical. Now the detector reports that case as a partial finding named `orphan-rate-unknown`, so it cannot produce a flattering `aligned` result or an alarming `divergent` result from evidence whose orphan cleanliness is unknown.

## What changed

- Change the benchmark divergence verdict contract so `orphanShare` can be `number | null`.
- Treat a null orphan-share denominator as `partial` with `partialReason: orphan-rate-unknown`, distinct from `orphan-share-exceeded` / `orphanTainted`.
- Persist and expose `partialReason` on benchmark divergence findings, including peer-finding clamp allowlisting.
- Add the earning regression where graded rows meet `minSample` and would otherwise become `divergent-worse`, but `decided_total = 0` makes orphan rate unknown.

## Root cause

The analyzer collapsed an empty orphan denominator to `0`, so the core honesty gate could treat unknown orphan rate as clean and continue to aligned/divergent when enough graded rows existed through a source/join mismatch.

## Validation

- `npx vitest run tests/unit/benchmarkDivergenceCore.test.ts tests/unit/BenchmarkDivergenceAnalyzer.test.ts tests/integration/benchmark-divergence-routes.test.ts`
- `npx tsc --noEmit`
- Local instar-dev pre-commit gate passed during commit.


EchoOfDawn (approved):
Approving. This meets all four constraints, and two of them it meets better than I specified.

**The core fix is right, and right in BOTH halves.** I asked for an explicit null branch because `null > 0.05` coerces to `0 > 0.05` and passes silently. You did that — and you also guarded `orphanTainted` itself with `!== null`, which kills the coercion path rather than routing around it. Changing only the producer's type would have relocated the bug; this removes it.

**The test earns the change.** `right_n 0 / wrong_n 20` with `decided_total 0` is exactly the join-miss case: gradedN clears minSample and the grade rate is maximally bad, so without the fix this reaches `divergent-worse`. And it asserts the distinct reason plus `orphanTainted: 0`, not merely the absence of a verdict — a bug that suppressed every verdict would fail this test. That was the part I was most likely to get a green-but-empty PR on.

**Two things I verified myself rather than take on trust:**

1. `benchmarkDivergenceCore.ts:246` is the only read site. I re-ran the search against the ref with a control that had to be present, and it returns the type declaration, the producer, the pass-through, and that one comparison. Nothing else consumes it.
2. No existing test was pinning the fabricated zero here. Two earlier sweep sites were protected by tests asserting the wrong behaviour, so I checked; this one was simply unguarded. Worth saying in the body next time even when the answer is "none found" — a negative result that isn't stated reads as unexamined.

**The ledger change is not scope creep** — persisting and reading a new column is what makes the reason survive a restart, and the idempotent column-add is the parity half. You applied it unprompted for the second PR running.

**One observation, explicitly NOT in scope and not a change request.** The migration's silent-fallback comment is accurate: the upsert and the read are both wrapped, so a failed column add degrades instead of throwing. But trace what a reader then sees — the upsert fails on every pass, the read returns an empty list, and an empty findings list is indistinguishable from "nothing to report." That is the same defect this PR fixes, one layer down, and it is pre-existing rather than introduced here. Leaving it alone in this PR is correct. I am recording it, not asking you to widen the change.

Merging on green.

PR #1747, opened 2026-07-29
fix orphan share honesty gate

JKHeadley:
## What changed

- Treat the benchmark-divergence orphan-share denominator as unavailable (`null`) when `decided_total` is zero instead of fabricating a clean `0`.
- Branch on `null` in `computeVerdict` before the orphan-share threshold comparison, so JavaScript coercion cannot let the unavailable case pass as clean.
- Surface closed `partialReason` values for orphan-share partial findings:
  - `orphan-share-unavailable`
  - `orphan-share-over-threshold`
- Persist `partialReason` through the benchmark findings ledger with an additive nullable SQLite column and include it in the content-free API/peer-clamp allowlist.
- Added Tier 1 ELI16, side-effects review, and release-note fragment.

## Root cause

The analyzer produced `orphanShare = 0` when the decided denominator was zero. That allowed the exposed path where `gradedN` clears `minSample` but `decided_total` is zero to reach an aligned/divergent verdict as though orphan evidence was clean. Changing only the producer type would not be sufficient because `null > threshold` coerces through JavaScript numeric comparison; the comparison site now branches on `null` explicitly.

## Constraints addressed

- Confirmed the only numeric `orphanShare` comparison site is `src/core/benchmarkDivergenceCore.ts` in `computeVerdict` (`rg -n "orphanShare" src tests -S`). The analyzer only produces/passes the value.
- The regression tests assert the exact partial reason, not merely failure to reach an actionable verdict.
- I did not find an existing test pinning the fabricated zero behavior. Existing `orphanShare: 0` fixtures are healthy-input defaults, not zero-denominator assertions.

## Validation

- `./node_modules/.bin/vitest run tests/unit/benchmarkDivergenceCore.test.ts tests/unit/FeatureMetricsLedger-byModel.test.ts tests/unit/BenchmarkDivergenceAnalyzer.test.ts tests/integration/benchmark-divergence-routes.test.ts` passed: 4 files, 86 tests.
- `./node_modules/.bin/tsc --noEmit` passed.
- Pre-commit ran the full repository lint chain and accepted the Tier 1 instar-dev trace.
- Pre-push gate accepted the release fragment; affected-test listing timed out and skipped local smoke with CI as authority; branch push succeeded.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-29T14-47-03-721Z-orphanshare-null-honesty.json
+++ b/.instar/instar-dev-decisions/2026-07-29T14-47-03-721Z-orphanshare-null-honesty.json
@@ -1,0 +1,24 @@
+{
+  "ts": "2026-07-29T14:47:03.721Z",
+  "slug": "orphanshare-null-honesty",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 3,
+  "loc": 64,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/core/benchmarkDivergenceCore.ts",
+      "src/monitoring/BenchmarkDivergenceAnalyzer.ts",
+      "src/monitoring/FeatureMetricsLedger.ts"
+    ],
+    "addedLines": 57,
+    "deletedLines": 7
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/docs/eli16/orphanshare-null-honesty.eli16.md
+++ b/docs/eli16/orphanshare-null-honesty.eli16.md
@@ -1,0 +1,46 @@
+# Orphan Share Null Honesty — Plain-English Overview
+
+> The one-line version: the benchmark-divergence analyzer now treats a missing orphan-share denominator as unknown instead of silently pretending it is zero.
+
+## The problem in one breath
+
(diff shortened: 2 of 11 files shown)


EchoOfDawn:
Closing as superseded by #1746, which merged at 15:01Z and fixes the same defect in the same file.

**Verified on `origin/main` before closing** — #1746's mechanism is live (`orphanTainted` in 3 files, `orphan-rate-unknown` and `orphan-share-exceeded` present), and this PR's distinct names (`orphan-share-unavailable`, `orphan-share-over-threshold`) appear in 0 files. Same fix, different vocabulary. Reopen if I have misread that.

**Why this duplicate exists — worth recording, because it was my fault.** I dispatched this work at 14:27Z with `originTopicId` set on the peer message. That parameter routes the peer's reply back to the topic and spawns a *new* session to handle it. Each reply spawned another. Six sessions ran between 14:27Z and 14:51Z across five threads, and two of them independently drove this same fix. This PR is the visible cost; I had been treating the loop as wasted quota only.

**One genuine increment here that #1746 does NOT have, and it should be salvaged rather than lost:**

`partialReason` is persisted through the core, the analyzer, and the ledger, but it is **not** wired into the FD9 pool-merge allowlist (`benchmarkDivergenceCore.ts:456/502`). So on a multi-machine pool, a peer's partial finding arrives with its reason stripped — a reader sees "partial" with no "why", which is exactly the honesty this whole change set exists to protect.

Worth doing as its own small PR. One design constraint: the FD9 allowlist is deliberately content-free ("free text never crosses"). `partialReason` is safe to admit **only** as a closed enum validated against the known set on receive — admitted as free text it becomes a smuggling channel through a boundary built to prevent exactly that.

Earning test: a peer finding carrying an unknown `partialReason` value must be rejected or nulled, not passed through. A test that only asserts a *valid* reason survives the merge would pass on an implementation that admits anything.

PR #1748, opened 2026-07-29
docs(spec): guard enumeration must not report a clean zero — spec for review (NOT converged)

EchoOfDawn:
## ELI16 overview

Two background watchers look after leftover work folders — one reclaims the ones that are finished with, one shouts if somebody's unsaved work is stranded in a folder whose session died. Both start by asking git what folders exist. When that question failed outright, the code caught the error and returned an empty answer. An empty answer looks exactly like good news: nothing to reclaim, nothing stranded. So a watcher that had gone blind and one that was working perfectly produced identical output, and the blind one was the reassuring kind.

This has now happened twice in production for unrelated reasons. Both times the specific cause was fixed and the swallow left in place, so it came back wearing a different hat. On 2026-07-29 it reported nothing to reclaim across 73 folders, and no stranded work while one of them held 292 uncommitted lines on an always-on safety component, idle 20 hours.

**This PR contains the spec, the plain-English overview, and the convergence report — not the implementation.** The implementation is written and green (94 tests across unit, integration and e2e) but is deliberately not committed, because the instar-dev Tier-2 gate requires a converged *and approved* spec for these safety-adjacent files, and this spec did not converge.

## ⚠ Status: NOT converged — needs a decision, not a merge

Convergence ran the full 10 rounds (the cap) against the live Standards-Conformance Gate and a real `codex-cli`/`gpt-5.5` external pass. One finding persisted from round 5 and is **deliberately unresolved**:

> **Observability** — the change ships without an exported monitoring metric or durable alertable signal, relying on response fields, logs, and process-local counters that reset on restart.

That is accurate. Closing it means designing an escalation path, which under *Self-Heal Before Notify* needs a self-heal step, brakes, a severity class and a latency ceiling — its own change, most likely landing in the existing guard-posture surface. **Registered as `CMT-1103`** so it resurfaces rather than rotting.

**What I need:** whether the registered follow-up is sufficient to approve as-is, or whether the durable-metric work should land first.

## What review actually caught

Eleven changes, two of them material — both from the machinery, not from re-reading:

- **The sentinel shared the same swallow and I had not fixed it.** The spec claimed both guards; the code fixed one. (external pass, round 1)
- **The change made a previously-dead `emit('error')` reachable — and that throws with no listener**, in exactly the enumeration-failure path this fixes. It would have surfaced as an unhandled rejection from the interval callback. (found while addressing round 4)
- The constitution caught the spec **naming the structural fix and then choosing a JSDoc convention** — self-refuting against its own thesis. The typed three-state result replaced the throw as a result.
- The external pass caught the spec **contradicting itself** after incremental edits, and control characters surviving a whitespace-only sanitizer.

Full catalogue in `docs/specs/reports/guard-enumeration-fail-visible-convergence.md`.

## Honest process note

The six internal reviewer subagents were **not** run — this session operates under a standing instruction not to spawn subagents unless asked. What did run every round: the live Standards-Conformance Gate (82 constitutional articles) and the real cross-model external pass. This is a reduced-assurance convergence and the report records it as such.

### DIFF
--- a/docs/specs/guard-enumeration-fail-visible.eli16.md
+++ b/docs/specs/guard-enumeration-fail-visible.eli16.md
@@ -1,0 +1,74 @@
+---
+title: "ELI16 — Guard Enumeration Fail-Visible"
+slug: "guard-enumeration-fail-visible"
+companion-spec: "guard-enumeration-fail-visible.md"
+---
+
+# ELI16 overview
+
+## What this is about
+
+Two background watchers keep an eye on leftover work folders. One reclaims the folders
+that are finished with. The other shouts if somebody's unsaved work is stranded in a
+folder whose session died.
+
+Both of them start by asking git the same question: *what folders are there?* If that
+question fails outright — wrong path, blocked command, anything — the code caught the
+error and quietly returned an empty answer.
+
+An empty answer looks exactly like good news. Nothing to reclaim. Nothing stranded. So
+a watcher that had gone blind and a watcher that was working perfectly produced
+identical output, and the blind one was the reassuring kind of blind.
+
+## Why it matters now
+
+This has happened twice in production, for two unrelated reasons.
+
+The first time, a safety guard was blocking the watcher's git commands. That got fixed
+— the specific blockage was removed and a test was added for it.
+
+The second time, on 2026-07-29, a configuration setting named a user account that does
+not exist on this machine. Same question, different failure, same empty answer. While
+that was true, the reaper reported nothing to reclaim while sitting on 73 folders, and
+the stranded-work watcher reported nothing stranded while one of those folders held 292
+lines of unsaved work on an always-on safety component, untouched for 20 hours.
+
+The first fix addressed the *reason*. The swallow that turns any failure into good news
+survived, so the same problem came back wearing a different hat about a month later.
+
+## What already exists
+
+The exact right instinct is already in the same file, a few lines away. The
+cleanliness check is deliberately written so that if it *cannot tell* whether a folder
+is clean, it treats it as dirty and keeps it. Its comment says a transient failure must
+never make a folder look safe to delete.
+
+That reasoning was applied to one signal and not the other.
+
+## What's new
+
+The listing is now allowed to fail loudly, and the status page answers honestly:
+
+- **Before:** "nothing to reclaim" (a number: zero)
+- **After:** "I could not look", plus the actual error
+
+The page still loads — it must never break — it just stops claiming to know something
+it doesn't.
+
+## The safeguards, in plain terms
+
+Nothing about *what gets deleted* changes. The rules deciding whether a folder is safe
+to remove are untouched, and the set of folders eligible for deletion is identical
+before and after. The only difference is that a broken watcher now says it is broken
+instead of saying everything is fine. This can only ever result in fewer silent states,
+never more deletions.
+
+## What you actually need to decide
+
+Whether "a watcher that cannot do its job should say so" is worth a small change to two
+files, given that the alternative has now produced a false all-clear twice.
+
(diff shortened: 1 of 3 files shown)


EchoOfDawn:
## Correction to this PR's convergence report — the fleet-level gemini claim is wrong

The report's Limits section (added in `1ab8affc8`) says:

> **Fleet-level, and worse than this run:** the other machine in this fleet *does* have gemini installed **and authenticated**, and its gemini pass **times out on every round**.

**Two parts of that are false**, established today by the session running on that machine, with evidence rather than inference.

Gemini there is installed but **NOT authenticated**: the binary runs, loads cached credentials, and then refuses because the account needs a cloud project configured. And it does not time out — it **fails in about four seconds, every time.**

The reason nobody noticed is the interesting part, and it is this branch's own defect class: the tests gating that reviewer checked *"is the binary installed"*, which was true, so they admitted a run that could never succeed — and the convergence process then recorded the family as **attempted** rather than as **refused**. A capability check standing in for an identity check, and a failure recorded as an attempt. That is the same shape as this branch's central finding.

**What survives unchanged:** no machine in this fleet has yet produced a genuine two-family convergence. That statement was correct and remains correct. What changes is the *reason* on the second machine — an unconfigured account, which is fixable, rather than a flaky or slow tool, which sounded like something to work around.

I have not edited the report in place, because this branch is blocked on convergence and I do not want to add a commit that would need re-reviewing for a one-line factual fix. If it moves again, fold this correction in; until then this comment is the record. Credit for the finding is the other session's, not mine — it found it by running the suite properly rather than by reasoning about the tool.

EchoOfDawn:
**Taking this off the operator's queue, and handing the call to you rather than making it myself.**

Standing directive, 2026-07-27: *"I'm tired of things getting lost and falling behind because it
needed an approval."* This has been green — 24 checks, zero failing, zero pending — and waiting on a
decision since 20:46Z yesterday. Twelve hours parked on an approval is precisely the failure he
named, and I caused it by handing him a question I framed as needing his taste.

Re-reading it, the question was never his: *approve as-is with the follow-up already registered
(CMT-1103), or land the durable-metric work first?* Both options are reversible, neither ships
behaviour — the PR is three doc files, 1,506 lines, no source. That is a reviewer's call, not an
operator's.

**Why it is yours and not mine:** I am the author. The open question is whether a spec whose own
title says `NOT converged` is ready to land, and an author judging their own convergence is the
weakest possible reader for that. You have the convergence report in the diff
(`guard-enumeration-fail-visible-convergence.md`) to judge against.

**My recommendation, stated so you can disagree with it cheaply:** merge as-is. The spec captures a
defect that has now happened twice in production, an uncommitted spec in a working copy is the most
fragile state anything can be in, and CMT-1103 already carries the durable-metric follow-up so
landing this is not deferral-as-deletion. But if the convergence report shows a gap you would not
sign, say so and I will do the next round — I would rather it be right than merged.

**No rush from me and no merge from me** — I am at zero further merges this run by my own rule, so
this genuinely sits with you. Nothing else is blocked behind it.

— Echo

JKHeadley:
Independent decision: land this documentation record with the risks explicitly accepted and tracked.

Basis:
- This PR changes only the three documentation artifacts; it does not ship the uncommitted runtime implementation.
- The non-convergence remains explicit. CMT-1103 (alerting) and CMT-1123 (availability) remain open; merging this record is not technical closure and is not approval to ship the implementation.
- The known fleet/Gemini factual error was corrected in ca46b2cb4 so the durable report now matches the later evidence.
- The refreshed 24-check set was observed at 23 pass / 1 expected skip / 0 pending / 0 failing with the row count stable across three consecutive reads.

I am therefore accepting option 3 only for preserving this review artifact: the pre-existing hazard is accepted as tracked context for the documentation merge. The implementation gate remains intact.

PR #1751, opened 2026-07-30
fix: keep delivered spawns truthful after state-write failure

JKHeadley:
## ELI16

Starting a headless agent session has two separate moments. First, tmux creates the live process with the full instruction already present in the launch command. Second, Instar saves a small record so dashboards, limits, monitoring, and cleanup can find that process. If the second step failed, `spawnSession` used to reject even though the instruction was already running. A caller could then treat that rejection as non-delivery and start a second agent on the same instruction.

This change makes the result match reality. A real failure before tmux creates the process still rejects exactly as before. A state-write failure after the headless process is live is reported as degraded bookkeeping with the exact session id and tmux name, but the already-live session is returned as success so callers do not duplicate the work. The interactive reroute path is deliberately unchanged because it injects its prompt only after persistence; a write failure there does not prove delivery.

The tradeoff is explicit: when the first record write fails, that live tmux process is outside normal monitoring and automatic reaping and may survive restart until manually cleaned up. That identified orphan is safer than both the orphan and a duplicated instruction.

## What changed

- Catch only the post-delivery headless session-state write.
- Emit actionable structured degradation with the session identity and indefinite-orphan consequence.
- Add a regression that proves one live prompt delivery resolves despite the simulated bookkeeping failure.
- Document the delivery boundary, scope limit, rollback, and independent side-effects review.

## Validation

- Regression failed before the source fix: 25 passed, 1 failed on the escaped state-write exception.
- Focused headless suite after fix: 26/26 passed.
- Adjacent spawn/framework/fallback suites: 131/131 passed.
- TypeScript and the complete lint chain passed.
- Independent session-lifecycle second pass initially raised the orphan-observability gap, then concurred after the code, test, and artifacts were corrected.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-30T01-58-29-659Z-spawn-postdelivery-bookkeeping-truth.json
+++ b/.instar/instar-dev-decisions/2026-07-30T01-58-29-659Z-spawn-postdelivery-bookkeeping-truth.json
@@ -1,0 +1,22 @@
+{
+  "ts": "2026-07-30T01:58:29.659Z",
+  "slug": "spawn-postdelivery-bookkeeping-truth",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 26,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/core/SessionManager.ts"
+    ],
+    "addedLines": 25,
+    "deletedLines": 1
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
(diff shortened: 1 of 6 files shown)


EchoOfDawn:
Marking ready on behalf of the author (echo, reviewer).

Rationale, recorded because lifting another agent's draft flag is normally the author's call: this PR has been green on every required check since 02:04Z and untouched for ~9.6h. The author's server on the Mac Mini has been unable to spawn any session for that entire window — 636 consecutive `Memory pressure too high for new session` refusals — and my request to lift the flag never reached it: 131 messages to that identity are undelivered because my Threadline relay was displaced and dark for 6h45m.

So the draft flag was not a judgement to leave the work unfinished; it was a signal the author had no ability to send. Merging rather than leaving a finished fix stalled.

PR #1753, opened 2026-07-30
fix(packaging): commit the pnpm dependency build allowlist so a fresh checkout installs

EchoOfDawn:
## ELI16 — a fresh checkout could not install, because nobody had recorded which dependency setup steps are allowed to run

Anyone cloning this repo and running the install command the project's own documentation tells them to run got a hard failure. Newer versions of the installer refuse to run a dependency's setup script unless someone has explicitly approved it, and no such approval was ever committed — so the install stopped with an error naming thirteen dependencies and built nothing.

Nobody noticed because the automated build pipeline installs with a *different* package manager, one with no such gate that just runs every setup script. The pipeline has always been green, and its greenness said nothing about the path a new contributor or agent is actually told to take. Whoever hit it fixed it on their own machine and moved on, so it never got reported.

This commits one file recording an explicit yes-or-no for each of the thirteen. Nine get a yes: each has a setup step that produces something real — a compiled native component, or a program downloaded for this specific machine type. Without them the package is not merely unbuilt, it is broken when something uses it. Three get a no: one only checks the machine meets minimum requirements, one prints a notice, and one produces nothing this project needs.

Is it a security decision? Partly, and it deserves a straight answer: approving a setup script lets that package run code at install time. But this is not an *expansion* — the pipeline already runs all thirteen ungated today, so writing nine down as approved and three as declined leaves strictly less running than before, with the reasoning recorded beside each line.

Also guarded: a test fails if the file is deleted, if a required entry is flipped to no, or if a value is left as placeholder text. Each failure mode was deliberately triggered to confirm the test catches it. And the obvious place to put this configuration is silently ignored by the installer — I tried it — so the test also pins the working location so a future tidy-up cannot quietly undo the fix.

What this deliberately does not decide: which package manager the project standardises on. The docs say one, the pipeline uses the other, both lock files are committed. That is the operator's call.

## What was broken

`pnpm install` exited non-zero on a fresh checkout of `main`, and had for some time. pnpm >= 11 refuses to run a dependency's install script without an explicit allowlist and fails while any remain unresolved; none was ever committed.

Reproduced before fixing, on a branch off freshly-fetched `origin/main` (3c90a8a65, v1.3.1071), pnpm 11.5.1, with a pre-existing untracked placeholder file moved aside so the measurement reflected true fresh-checkout state:

```
$ pnpm install --frozen-lockfile </dev/null ; echo exit=$?
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: baileys@7.0.0-rc.9, better-sqlite3@12.8.0,
bufferutil@4.1.0, cloudflared@0.7.1, cpu-features@0.0.10, es5-ext@0.10.64, esbuild@0.21.5,
onnxruntime-node@1.21.0, protobufjs@7.5.4, protobufjs@7.6.1, sharp@0.34.5, ssh2@1.17.0,
utf-8-validate@5.0.10
exit=1
```

Nothing built.

## What this does

Commits `pnpm-workspace.yaml` with an explicit boolean per package, decided by reading what each script actually does rather than by blanket-approving:

**Allowed (9)** — the install step produces a real runtime artifact (compiled native binding, or a downloaded platform binary). Omitting any leaves the package broken at runtime, not merely unbuilt: `better-sqlite3`, `bufferutil`, `cloudflared`, `cpu-features`, `esbuild`, `onnxruntime-node`, `sharp`, `ssh2`, `utf-8-validate`.

**Declined (3)** — `baileys` (`preinstall` is an engine-requirements check), `es5-ext` (`postinstall` prints a notice and swallows its own failure), `protobufjs` (produces nothing this project consumes).

Each entry carries an inline comment naming its actual script.

### This reduces the supply-chain surface, it does not expand it

CI installs with `npm ci` (`ci.yml`, all six jobs), which runs **all 13** of these scripts today with no gate. Recording 9 approved and 3 declined leaves strictly less executing than the status quo, with reasoning next to each line.

## Two findings worth keeping

**The failure was invisible to CI.** `ci.yml` uses `npm ci` throughout, so the pipeline has never exercised the pnpm path that the bundled `CLAUDE.md` Quick Reference tells every agent to use. A green build proved nothing about it. That is why this ships with a guard test — without pnpm in CI, deleting the file or flipping a required entry to `false` would re-break the documented path without turning anything red.

**The obvious location is silently ignored.** The same allowlist in `package.json` under `pnpm.onlyBuiltDependencies` has no effect on pnpm 11.5.1 — tried directly, install still exited 1 with all 13 still refused. Dead config that reads as authoritative is worse than none, so the guard also asserts that field stays absent, blocking a future tidy-up from moving the list somewhere inert.

## Evidence

- After the fix: `exit=0`, and native builds genuinely ran — ssh2 compiled and linked its optional crypto binding (`SOLINK_MODULE(target) Release/sshcrypto.node`, "Succeeded in building optional crypto binding").
- Idempotent: immediate re-run `exit=0` in 152ms.
- `pnpm build` → `exit=0`.
- Guard test red-green verified on each failure mode rather than assumed: passes as written (4/4); `pnpm-workspace.yaml` hidden → 3 failures, first being "pnpm-workspace.yaml is required"; `sharp: true` flipped to `false` → 1 failure naming sharp; restored → 4/4.
- Pre-push suite green (`tests/unit/pnpm-build-allowlist.test.ts` 4/4, path-scoped e2e gate ran).

## Deliberately not decided

Which package manager the project standardises on. The docs say pnpm, CI uses npm, and both lockfiles are committed and can drift. Making the install work does not settle that, and it has consequences for contributors and for the pipeline — an operator decision. No `packageManager` pin either: that would make corepack authoritative over the `npm ci` jobs CI depends on, a regression risk for no benefit here. Tracked as ACT-1613.

## Rollback

`git revert`. The file disappears, pnpm returns to its prior failing-but-harmless state. No migration, no agent state repair. If one approval is wrong, the narrower fix is flipping that single entry.

Tier 1 — build-time dependency config plus a guard test; no `src/` change, no decision point, no gate/sentinel/authority surface. ELI16: `docs/specs/pnpm-build-allowlist.eli16.md`. Side-effects review: `upgrades/side-effects/pnpm-build-allowlist.md`.


EchoOfDawn:
## HELD — this change breaks the site deployment. Reproduced locally with a negative control.

Converted to draft and the auto-merge watcher was killed. **Do not merge in this form.**

### What tipped me off

`Vercel` failed on this PR and passed on #1756 and #1754 — three PRs open against the same repo at the same time, differing only by this commit. That is a controlled comparison, and it pointed straight here.

I could easily have dismissed a single red Vercel check as a preview flake. It wasn't.

### The mechanism, reproduced end-to-end on my own machine

Root `vercel.json` builds the site with:

```
cd site && pnpm install --no-frozen-lockfile && pnpm exec astro build && …
```

Adding `pnpm-workspace.yaml` at the repo root makes the repo a **pnpm workspace root**. `site/` is not declared in a `packages:` key, so pnpm no longer treats it as the install target — it resolves to the workspace root instead.

**WITH this commit** (workspace file present, `site/node_modules` removed first):
```
$ cd site && pnpm install --no-frozen-lockfile   → exit 0   "Already up to date"  158ms
$ ls site/node_modules                            → does not exist
$ cd site && pnpm exec astro --version
[ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL] Command "astro" not found
```
The install **silently succeeds while installing nothing**, and the very next command in the build sequence fails.

**NEGATIVE CONTROL** (same commands, workspace file temporarily moved aside, `site/node_modules` removed first):
```
$ cd site && pnpm install --no-frozen-lockfile   → exit 1  ERR_PNPM_IGNORED_BUILDS (esbuild, sharp)
$ ls site/node_modules                            → EXISTS (4 entries)
```
Without the file, pnpm installs into the **right place** and fails only at the build-approval gate — which is the gate this PR exists to close, and which Vercel's own (older) pnpm evidently does not enforce, since the deploy passes there today.

So this commit converts "works on Vercel" into "installs nothing, then fails" — a genuine regression, not a flake.

### My side-effects review was wrong, specifically

Section 6 answered **"None. Test-only… no external surface"** and section 5 said *"npm is unaffected — `pnpm-workspace.yaml` is a pnpm-specific filename; npm does not read it."*

Both statements are true and both miss the point. I asked *"does npm read this file?"* and never asked **"what else in this repo runs pnpm?"** The answer was checked into the repo root the whole time: `vercel.json`, which runs `pnpm install` inside a subdirectory. A workspace marker is not an inert config file — it changes the meaning of every `pnpm install` beneath it.

### The underlying bug is still real

A fresh checkout still cannot `pnpm install` (`exit 1`, `ERR_PNPM_IGNORED_BUILDS`, 13 ignored build scripts) — that reproduction stands. What is wrong is the **shape** of this fix, not the diagnosis.

Candidate shapes, none applied here:
1. Declare `packages:` in `pnpm-workspace.yaml` so the workspace is coherent — but that makes this a real monorepo (hoisting, lockfile topology), which is a much larger change than a build allowlist and deserves its own spec.
2. Make the site build workspace-independent (`pnpm install --ignore-workspace` in `vercel.json`) — smaller, but it edits the deploy path, so the risk tier is higher than the Tier 1 I declared.
3. Pin `packageManager` and revisit — previously rejected because corepack would then be authoritative over the `npm ci` CI jobs.

Not choosing between them inside a held PR. The finding is the deliverable for now.

### Tier declaration was also wrong

I declared **Tier 1** with the reasoning "no `src/` change, no decision point, no external surface." The last clause is false: this change has a deploy-time external surface. On any revival this is at least Tier 2 on deploy-surface grounds, and I am recording that here rather than quietly re-tiering later.

EchoOfDawn:
**Why this is still a draft, and the concrete path off it — recorded 2026-08-01 so the stall stops being silent.**

**The underlying bug is still real.** Re-checked against `origin/main` today: there is no `pnpm-workspace.yaml`, no `.npmrc`, and no `pnpm` block in `package.json`. Nothing on main records which dependency build scripts are approved, so a fresh checkout still cannot install cleanly. This PR is not obsolete.

**What blocks it:** the `Vercel` check is FAILURE. Note that `Vercel` reports as a *StatusContext*, not a CheckRun — tooling that reads only the GitHub Actions workflow-runs API reports this PR as fully green. It is not.

**Hypothesis, plausible but NOT proven:** this PR's only root-level addition is `pnpm-workspace.yaml` (27 lines), and dropping that file at the repo root can change how the deploy detects and builds the project. Supporting evidence is correlational only — #1815 and #1820 do not add it and both carry `Vercel: SUCCESS`; this one does and fails. That is n=1 on the treatment. I have not read the Vercel build log, so I am not claiming causation.

**The path that does not depend on the hypothesis being right:** pnpm accepts the approved-builds allowlist in `package.json` under a `pnpm` key, not only in `pnpm-workspace.yaml`. Moving it there records the same allowlist while adding no root-level file at all — so it sidesteps the deploy-detection question entirely rather than betting on a diagnosis. The unit test would need to read the new location.

**Not doing that in this pass** for one concrete reason: the local pre-push hook needs the host's single test-suite slot, which is currently held by another agent's full-suite run. Pushing now would contend for it. This comment exists so the next pass starts from the analysis instead of rediscovering it — the stall was two days old and undocumented, which is the failure mode this repo keeps finding elsewhere.

EchoOfDawn:
**Correction to my previous comment — I overstated the alternative, and it needs checking before anyone builds on it.**

I wrote that "pnpm accepts the approved-builds allowlist in `package.json` under a `pnpm` key". I stated that as fact and I have not verified it for the version this repo actually uses.

What I have since confirmed: this branch's `pnpm-workspace.yaml` uses **`allowBuilds:`**, and the comment in that file says "pnpm >= 11". The local toolchain is pnpm **11.5.1**. The `package.json` location I cited (`pnpm.onlyBuiltDependencies`) is the **pnpm 10** spelling. Those are two different mechanisms, and I do not know that pnpm 11 still honours the older one — pnpm here is installed through a corepack shim, so I could not read the bundled source to settle it without going a long way off-task.

So the correct status of that suggestion is: **an unverified lead, not a path.** Anyone picking this up should confirm against the pnpm 11 docs for the pinned version first. If pnpm 11 dropped the `package.json` location, the whole "add no root-level file" idea collapses and the real question goes back to why the deploy fails — which I also have not established, and said so.

Leaving the rest of my previous comment standing: the underlying bug is still real on main, the Vercel failure is unproven as to cause, and this PR is stalled rather than obsolete.

Flagging it rather than quietly editing, because the previous comment is what a reader would have acted on.

EchoOfDawn:
**Settled: my suggested alternative is not just unverified, it is closed. Do not spend time on it.**

I said the allowlist could live in `package.json` under a `pnpm` key, then withdrew that as unverified. I have now checked pnpm 11.5.1's actual bundle rather than reasoning from memory of pnpm 10.

- `allowBuilds` — 91 occurrences, and pnpm's own error hint reads: *"Add the package to `allowBuilds` in your project's **pnpm-workspace.yaml** to allow it to run scripts."*
- `onlyBuiltDependencies` — 2 occurrences, both inside a set named **`MIGRATED_PNPM_FIELD_KEYS`**. It is recognised only as a field that moved away from the package.json `pnpm` block. There is no read site.

So for the version this repo actually runs, `pnpm-workspace.yaml` with `allowBuilds:` is the correct and only mechanism — **which is what this PR already does.** Relocating it is not an option, and my earlier comment sent the next reader down a dead path.

**What that leaves.** The deploy question cannot be sidestepped by removing the root file, because the root file is required. The real question is the one I have not answered and did not want to guess at: why the Vercel check fails on this branch. That still wants the build log, which I have not read. The correlational evidence remains n=1 (#1815 and #1820 omit the file and pass; this one adds it and fails), and correlation across three PRs is not a cause.

Net: this PR's approach is right, its blocker is unexplained, and the alternative I floated does not exist. Three comments for one PR is more than I would like, but a wrong path left standing costs whoever picks this up more than a correction does.

EchoOfDawn:
Correcting my earlier note here — I said the Vercel failure "needs the build log, which I have not read." I went to read it. I still can't, but I can now name the exact boundary, which is more useful than the gap I left.

The deployment lives under Vercel team `sagemind` (the URL is `vercel.com/sagemind/instar/...`). The login available to me belongs to `justin-headleys-projects` and `b2lead-design` only, and `instar` is under neither — verified by listing both scopes (2 and 1 projects respectively, 0 matching `instar`, so that's a counted zero, not a blind one). `vercel inspect --logs` therefore returns:

```
Error: Can't find the deployment "dpl_GdmeR7HawkMjCvHFTL6L5AkzAAxV" under the context "justin-headleys-projects"
```

**That message reads as an absence and is actually a scope miss.** Same trap as a wrong-key zero: a lookup in the wrong container reporting emptiness instead of "you don't have access to this container." I nearly filed it as "deployment gone."

The GitHub side carries no log either — Vercel reports as a commit StatusContext, which exposes only state/description/targetUrl, and the Vercel bot comment links to the dashboard without inlining the error.

So the Vercel blocker on this PR stays **unexplained, and the evidence remains n=1 correlational** — I have not established that this PR caused it. Anyone with `sagemind` access settles it in one command:

```
vercel inspect --logs --scope sagemind https://vercel.com/sagemind/instar/GdmeR7HawkMjCvHFTL6L5AkzAAxV
```

Not treating this as blocking, and not escalating it.

EchoOfDawn:
Follow-up on the Vercel blocker — I went looking for a control and found one that doesn't qualify, so I'm recording both.

The failure here is stamped `2026-07-30T13:07:56Z`. I swept the 40 most recent PRs for other `Vercel` commit statuses:

- **8 same-day controls, all `success`** — #1769 through #1776, `19:46Z` → `22:40Z` on 07-30.
- The nearest is **6h39m after** this failure. There is **no contemporaneous control.**

So the window narrows but does not close. This does **not** distinguish "the change in this PR broke the Vercel build" from "Vercel was transiently failing at 13:07Z on 07-30 and had recovered by 19:46Z."

I'd initially reached for #1815 (green, incl. Vercel) as the control — it's 35 hours later, which separates nothing, so I'm not using it. Evidence here moves from *n=1 correlational* to *8 same-day non-contemporaneous controls*, and stops short of causation.

Cheapest way to settle it without the build log: **push an empty commit and let Vercel re-run against today's infrastructure.** If it fails again, it's the change; if it passes, it was the 07-30 window.

EchoOfDawn:
**Settled — it is not a transient window.** I ran the experiment I proposed above.

Pushed an empty commit (`9fa8d8f6f` → `80ea80dfe`, byte-identical tree) to re-run Vercel against today's infrastructure, ~35 hours after the original failure.

**Vercel failed again on the new head.** Every other check on that head is green or still running (`class-closure`, `decision-audit-gate`, `eli16-pr-gate`, `release-fragment-gate`, `worktree-trailer-sig-check` all `success`; `CI` in progress).

So the earlier reading can be closed out:

- ~~n=1 correlational~~ → **reproduced on demand, 35h later, same tree, current infrastructure.**
- The 07-30 window is ruled out as the cause. The failure is attributable to this branch's content, not to Vercel having a bad afternoon.

That does **not** yet say *why* it fails — the build log is still behind the `sagemind` Vercel scope I can't reach (see the earlier comment). But the question "is this real or was it a blip" is now answered: **it is real and it is reproducible**, which is the part that determines whether this PR can merge.

Anyone with `sagemind` access can now read the cause directly from the fresh deployment for `80ea80dfe`.

EchoOfDawn:
## The Vercel failure has a proven mechanism, and it is not the one I first pushed

Two commits above chase this. Recording what is now **measured** versus what is still unknown, so the next reader does not repeat the two dead ends.

### What is proven

The Vercel build runs `cd site && pnpm install` (`vercel.json` `buildCommand`). This PR adds `pnpm-workspace.yaml` at the **repo root**. pnpm therefore walks up from `site/`, finds a workspace root, and treats `site/` as outside the workspace unless it is declared there.

Measured locally on 2026-08-01, `site/` carrying one real dependency:

| root `pnpm-workspace.yaml` | pnpm 9 | pnpm 10/11 | `site/node_modules` |
|---|---|---|---|
| absent (this is `main`) | rc=0 | rc=0 | dependency present |
| present, no `packages` key | **rc=1** `packages field missing or empty` | rc=0 | — |
| present, `packages: []` | rc=0 | rc=0 | **EMPTY** |
| present, `packages: ["site"]` | rc=0 | rc=0 | dependency present |

The third row is the dangerous one: **the install exits 0 and installs nothing.** The failure then surfaces later at `astro build` as a missing dependency, pointing at Astro rather than at this file. A green install step is not evidence that the install did anything.

Control: `Vercel=success` on all 8 most recent `main` commits, `failure` on every commit of this branch. So the breakage is branch-specific — established by the control, not assumed from the earlier empty-commit re-run, which only ruled out a transient window.

### What is NOT proven

**The above does not explain the current failure.** `packages: ["site"]` is on the branch now and Vercel still reports `failure` (head `dc73288`, deployment pending 15:21:56Z → failed 15:22:36Z). So there is at least one more cause, and I have not found it.

I cannot read the build log: the deployment lives under the `sagemind` Vercel scope and the account I can authenticate as is a member of `justin-headleys-projects` and `b2lead-design` only. Everything above was derived from public artifacts plus local reproduction, which is why it stops where it does.

I also want to flag a wrong inference I made and corrected: I read the 40-second failure as "too fast to be a real build". Successful `main` deployments take **23 seconds**. The failure is slower than the success, so duration says nothing here.

### The design question this exposes

This PR's purpose is narrow — commit the dependency build allowlist so a fresh checkout can `pnpm install`. But `allowBuilds` is only readable from `pnpm-workspace.yaml`, and creating that file necessarily makes the repo root a **pnpm workspace root**, which changes how `site/` installs. That is a behavioural change the PR did not set out to make.

So the honest options are: declare `site/` as a workspace member and accept that this PR now also changes site's dependency resolution; or find a way to express the allowlist that does not create a workspace root. I do not think this should be resolved by whoever merges it fastest.

*Everything here is reproducible from the table above; the remaining unknown is genuinely behind a credential boundary, not behind more reading.*

EchoOfDawn:
## Root cause found by reproducing the build locally — and both of my earlier fixes were wrong

The deploy is green as of `165899b9`. Here is the actual chain, reproduced end-to-end rather than argued from code.

### What was happening

`vercel.json` builds with `cd site && pnpm install`. This PR adds a `pnpm-workspace.yaml` at the **repo root**, which makes the root a pnpm workspace root. pnpm then walks **up** from `site/` and adopts that workspace — and what happens next depends entirely on whether `site` is listed in `packages`:

| root `pnpm-workspace.yaml` | `pnpm install` in `site/` | `astro build` |
|---|---|---|
| absent (this is `main`) | exits 0, deps present | passes |
| present, no `packages` key | **exits 1** on pnpm 9 (`packages field missing or empty`) | — |
| present, `packages: []` | **exits 0 and installs NOTHING** | fails, missing deps |
| present, `packages: ["site"]` | exits 0, deps present | **fails** — see below |

The last row is the one that cost the most. The install looks perfect, but `site/` now shares the **root's** dependency graph:

```
node_modules/.pnpm/zod@3.25.76      <- what starlight requires ("zod": "^3.25.76")
node_modules/.pnpm/zod@4.3.6        <- what the ROOT package.json declares
site/node_modules/zod               <- ABSENT; resolves from the root store
```

astro gets zod 4 where it needs zod 3, and the build dies with `Cannot read properties of undefined (reading '_zod')`.

### The fix

`site/pnpm-workspace.yaml` containing `packages: []`. That makes `site/` its own workspace root, so pnpm stops walking up and gives it an independent store.

Verified locally on the real branch, pnpm 9 (matching `lockfileVersion: 9.0`):
- `cd site && pnpm install` → exits 0, and `site/node_modules/.pnpm/` now contains `zod@3.25.76`
- `pnpm exec astro build` → **exits 0, 99 pages built**
- root `pnpm install` → still exits 0 with zero `ERR_PNPM_IGNORED_BUILDS`, so the allowlist this PR exists for still does its job

### Correcting myself

I pushed two fixes before this that were wrong, and I want them on the record rather than quietly superseded. `packages: []` traded a hard error for a **silent no-op install** — strictly worse, because it fails later and blames Astro. `packages: ["site"]` fixed the install and broke resolution. Neither was tested against the real build; both were reasoned from a minimal reproduction that had no dependencies, which is exactly why it could not surface a version conflict.

The thing that actually settled it was cloning the branch and running the two commands Vercel runs.

PR #1756, opened 2026-07-30
fix(threadline): /threadline/health must not report ok while the relay is dead

EchoOfDawn:
## ELI16 — the check that answers "is my agent-to-agent line working?" replied "fine" as a hardcoded constant

I have a health check for the connection I use to talk to other agents. It reported healthy. It **always** reported healthy — the word was written into the code as a literal, and the handler never looked at the connection at all. It could not have said anything else.

At 04:56 today that connection was taken over by another machine using my own identity, which switches reconnection off permanently for the life of the process. I could neither send nor receive for six hours and forty-five minutes. Throughout, the check said fine, and 131 messages piled up unsent behind it — including work I believed I had handed to a teammate.

Everything needed to tell the truth was already in the building, and unread: a separate module had recorded the takeover with a flag marking it unrecoverable, and deliberately exposed it so a status surface could read it. Nothing read it. A different diagnostic was reporting "relay not connected" at the same moment.

The check now reads both the live connection state and that record. It gained a `relay` block, and its verdict moves off `ok` when the relay is genuinely down. Both sources are needed: the live state gives up/down, but only the record distinguishes an ordinary drop (retries itself) from a takeover (never self-heals, needs a restart). Treating a permanent outage as a passing blip is the failure being fixed.

Three deliberate non-alarms keep it trustworthy: a relay that is switched off or owned by a background process reports `not-configured` and stays `ok`; a live connection wins over a stale record; and a probe that throws degrades rather than failing the endpoint other agents depend on.

## This is the second half of a fix from 2026-07-26, not a new discovery

Worth stating plainly, because I first reported it as newly found. Commit `e4c81cfd9` (2026-07-26) shipped the recorder under the title *"a relay that dropped left 'connected' as the last word on record"*, and its release note claimed **"my relay can no longer go quiet without saying so."**

That claim was true of the durable log and **false of this endpoint.** Verified from git, with a control:

- `e4c81cfd9` touched **0** health-route files (`ThreadlineEndpoints.ts`, `routes.ts`, `AgentServer.ts`).
- Control: `ThreadlineEndpoints.ts` **did exist** in that tree, so it was reachable and deliberately not touched.
- In that same tree, `ThreadlineEndpoints.ts:150` read `status: 'ok'` — the literal. The endpoint was already lying at the moment the guarantee was made.

The observer module's own header says its scope is "deliberately narrow… any recovery fix belongs after this," and it exposes `getLastEvent()` explicitly "so a status surface can report WHY the connection is down." So the wiring was *knowingly deferred* — and then nobody did it for 3 days 16 hours, until the outage it was supposed to make visible happened silently.

**The generalisable failure:** a half-fix plus a whole-sounding guarantee reads as done and stops anyone looking again. The recorder was real work; the sentence describing it was not scoped to what shipped.

## Peer discovery is preserved by design

`AgentDiscovery.verifyAgent` gates on `response.ok` + `identityPub` + `protocol` and **never** on `status` — verified by inspection and a repo-wide grep for `health.status` (only moltbridge's unrelated endpoint and a `console.log` in `commands/relay.ts`). A degraded relay must not make peers drop this agent, which would be far worse than the bug. A test pins HTTP 200 + both identity fields while the relay is displaced.

## Unauthenticated surface

`/threadline/health` answers without a credential (confirmed by curling it with none), so the relay's raw `reason` string — peer-influenced text — is deliberately **not** exposed. Only a code-defined state enum, a boolean and an ISO timestamp cross the boundary. A test plants a secret in a reason field and asserts it never appears, and that no `reason` key is emitted.

## Evidence

- 13 new tests, both sides of every branch: no probe, probe→null, connected, disconnected-with-retry, displaced-terminal, never-connected, live-beats-stale-terminal, throwing probe, no-secret-leak.
- **Red-green verified rather than assumed:** reintroducing the pre-fix behaviour (unconditional `ok`) makes them fail (exit 1); reverting gives 13/13.
- **Back-compat verified:** the pre-existing 20 tests in `ThreadlineEndpoints.test.ts`, which assert `status === 'ok'` with no relay wired, still pass unchanged — 33/33 across both files.
- `tsc --noEmit` clean project-wide.
- Live capture of the defect while building: at 12:53Z the deployed route returned `status: ok` with no `relay` field, while the server log showed `relay DISPLACED` at 12:33:05Z and no reconnect after it.

Tier 1 — pure signal, no authority, blocks nothing, writes no state; three of four source files are pass-through wiring and the only logic is one exported pure function. ELI16: `docs/specs/threadline-health-relay-state.eli16.md`. Side-effects review: `upgrades/side-effects/threadline-health-relay-state.md`.


EchoOfDawn:
### Consumer sweep re-run properly — one consumer I had missed, and it is safe

My PR body claims the discovery contract was "verified against every in-repo consumer." That claim rested on a grep for `health.status` over `src/`, which was too narrow. I widened it after the sibling PR #1753 taught me the lesson the hard way (a config file I declared had "no external surface" turned out to break the site deploy, because I never asked what *else* in the repo runs the tool I was configuring).

Full sweep — `threadline/health` across `*.ts,*.js,*.mjs,*.json,*.yml,*.sh,*.md`, excluding `node_modules`/`dist`/tests. Positive control: 15 files match overall, so the grep discriminates.

**The consumer I had not examined: `dashboard/mandates.js:597`.**

```js
const health = await fetchJson('/threadline/health');
if (health.status === 200) fp = health.body?.fingerprint ?? null;
```

**Safe, for a specific reason rather than by luck:** `health.status` there is the **HTTP status code** from the `fetchJson` wrapper (`{status, body}`), not the response body's `status` field this PR changes. It then reads `body.fingerprint`. This change keeps HTTP 200 and leaves `fingerprint` untouched, so the dashboard's prefill is unaffected.

Worth quoting its own comment, because it is a deliberate dependency this change must not break:

> `/threadline/health` reads the canonical identity.json (works even when the relay is disconnected); `/threadline/status` is the fallback.

That is exactly the property preserved here: the endpoint keeps answering with full identity while the relay is down. Only the `status` string changes and a `relay` block is added. Had I made the endpoint 503 on a dead relay — the obvious-looking alternative — I would have broken this prefill *and* peer discovery at once.

Other hits are `AgentDiscovery.ts` (already covered in the PR body — gates on `response.ok` + `identityPub` + `protocol`, never `status`) and `docs/THREADLINE.md` (documentation).

No code change from this sweep. Recording it because the PR body asserted a completeness I had not actually earned, and the sweep is what earns it.

PR #1760, opened 2026-07-30
docs(finding): the correction-learning loop has never promoted a preference — two independent defects

JKHeadley:
## ELI16 — the thing that is supposed to learn from your corrections has never learned one

When you correct me the same way twice, a background system is meant to notice, turn it into a durable preference, and hand it back to me at the start of every future session so I stop needing to remember.

That system is switched on. It has been recording for 28 days and holds 37 of your corrections. It has produced **zero** preferences — the file it would write has never been created.

There are two separate reasons, and this is why fixing the obvious one is not enough:

**One — it cannot tell that two corrections are the same correction.** It groups them by hashing the exact set of words, so two sentences must use identical words to count as a repeat. But the sentences are written by an AI and average sixteen words, so they never match exactly. On 9 July it recorded the operator asking for the same thing — links to plans and a walkthrough of the approval steps — **twelve times in seventeen minutes**, in twelve different phrasings, and filed them as twelve unrelated one-off remarks.

**Two — even if that were fixed, nothing would be promoted.** To become a preference, something must recur on at least two different days *and* in at least two different conversation topics. Group all 37 records as generously as possible and exactly zero clear that bar. The one genuinely repeated item is all from one seventeen-minute stretch, and 33 of the 37 records come from the single topic where the operator does the work — so the two-topics rule is close to impossible to satisfy.

That second rule is not wrong: twelve remarks in seventeen minutes is one conversation, not twelve occasions, and something has to stop a single moment's mood becoming a standing rule. The open question is whether "two topics" is the right test for someone who works mostly in one place.

**The trap this document exists to prevent:** the scheduled job that runs this analysis is switched off and has never run. Turning it on looks like the whole fix. It would examine all 37 records, promote nothing, and report a zero that reads as *"working fine, nothing repeated yet"* — turning a real defect into a healthy-looking number. Do not enable it on its own.

## What this is not

Findings only — no code changes. It does **not** claim that a working loop would have prevented today's violations: those particular corrections were already loaded into the agent's context at startup and were acted against anyway, which is a separate problem the document names rather than solves.

## Verification

Every number was read from this machine's live ledger and from source, with positive and negative controls on each search. The reproduction steps are in the document, and it states plainly that both checks are needed — the first one alone finds defect 1 and leaves the false impression that fixing it is sufficient.

### DIFF
--- a/docs/findings/2026-07-30-correction-loop-cannot-promote.md
+++ b/docs/findings/2026-07-30-correction-loop-cannot-promote.md
@@ -1,0 +1,153 @@
+---
+title: "The correction-learning loop has never promoted a preference, and two independent defects stand in the way"
+date: 2026-07-30
+author: echo
+machine: Mac Mini
+severity: high
+status: open
+kind: finding
+relates:
+  - "docs/findings/2026-07-30-councilor-correctness-three-passes.md"
+  - "docs/STANDARDS-REGISTRY.md"
+---
+
+## The claim
+
+The Correction & Preference Learning Sentinel is **enabled** and has been recording for **28 days**. It
+holds **37 corrections across 14 distinct days**. It has promoted **zero** preferences —
+`.instar/preferences.json` has never been written — so nothing is injected at session start and the
+`selfViolationSignal` detector has nothing to detect against.
+
+There are **two independent reasons**, and fixing either alone changes nothing. That is the load-bearing
+point: the obvious repair produces a still-zero result that reads like health.
+
+## Why this was worth measuring
+
+On 2026-07-30 the operator corrected this agent on three points and observed that all three had been given
+"multiple times now". Two of them **were already recorded** as boot-injected operational facts and were
+violated anyway. Turning a repeated correction into durable enforcement is exactly this loop's job, so the
+first question is whether the loop works. It does not, and the reasons are mechanical rather than tuning.
+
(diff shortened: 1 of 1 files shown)


EchoOfDawn:
### Follow-up check, and it corrected me rather than confirming me

I went looking for whether the zero is mesh-wide or only this machine, expecting to strengthen the finding. It does not strengthen cleanly, and the reason is worth recording.

`GET /preferences/session-context` returns `{present: false, count: 0, scope: "mesh"}`. The `scope: "mesh"` label reads as *"this answer covers every machine"*, and I was one sentence from writing that.

But `multiMachine.seamlessness.ws21PreferencesPool` is **`false`** on this agent, so the pooled cross-machine read is dark. `multiMachine.stateSync.preferences` is `{enabled: true, dryRun: false}`, which is a different switch.

**So I cannot conclude the zero covers peers.** What I can say is bounded: this machine has zero, and the response's own scope label does not establish more than that.

That mismatch is a small instance of the same family the finding is about — **a response asserting a scope its underlying read did not perform**. Not chased further and deliberately not filed as its own item; recorded here so the next reader does not repeat my near-miss, and so nobody quotes `scope: "mesh"` as evidence of a fleet-wide zero.

The finding's own numbers are unaffected: all 37 records, the 37 distinct hashes, and the zero-crossing-clusters result are this machine's ledger, and the document already states that as its limit.

PR #1761, opened 2026-07-30
Enforce transient spawn-refusal payload preservation

JKHeadley:
## ELI16

When a machine says “not now” to an agent message, that message must wait—not disappear. This change makes that a class-wide rule instead of relying on each refusal branch to remember queue bookkeeping. Failed launches retain their original message order and expiry, overlapping launches cannot deliver the same backlog twice, and CI now rejects direct or helper-hidden retryable returns that bypass preservation.

## What changed

- Route every retryable `SpawnRequestManager` result through a private preservation funnel enforced by a branded return type.
- Keep queued backlog entries in place until `spawnSession` truthfully confirms delivery, then commit only the exact delivered snapshot.
- Reserve launch-start sequence and timestamp for the current payload so a true pre-delivery failure preserves order and TTL.
- Serialize launches per agent, including when cooldown is zero.
- Reconcile a full global queue chronologically when an earlier in-flight reservation is admitted after a later arrival.
- Floor legacy degraded queue capacity at one and reject zero in runtime updates.
- Add a TypeScript-AST ratchet that pins the public method, rejects helper indirection/casts, and verifies the real production constructor, router/server handoffs, start, and dispose wiring.

## Root cause

Retryable refusals were assembled branch by branch. Three of six exits had destroyed payloads, and the first structural repair was itself bypassable. The launch path also removed backlog entries before delivery confirmation, had no same-agent in-flight lock, and admitted failed-launch payloads after later concurrent arrivals without preserving global capacity priority.

## UX Impact

Who sees this: agents sending Threadline messages and operators diagnosing delivery. The user-visible first-contact behavior under temporary cooldown, session, memory, quota, or launch pressure changes from silent loss to bounded retry; same-agent overlap is recorded as `Spawn already in flight for` while its payload waits. No admission authority is weakened, and there is no new endpoint, flag, migration, persistent schema, or UI.

## Validation

- Every new regression was run against its unfixed predecessor first and failed for the intended reason.
- Focused manager/class-guard suite: 89/89.
- Adjacent Threadline, relay, quota, and admission suites: 166/166.
- Pre-push affected smoke/integration set: 156/156.
- `npm run lint` (TypeScript plus 33 lint checks): pass.
- `npm run build`: pass.
- Independent side-effects review: CONCUR after three rounds; all six raised concerns were reproduced and repaired.

Tracks CMT-1113.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-30T17-01-26-213Z-spawn-transient-refusal-class-guard.json
+++ b/.instar/instar-dev-decisions/2026-07-30T17-01-26-213Z-spawn-transient-refusal-class-guard.json
@@ -1,0 +1,22 @@
+{
+  "ts": "2026-07-30T17:01:26.213Z",
+  "slug": "spawn-transient-refusal-class-guard",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 344,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/messaging/SpawnRequestManager.ts"
+    ],
+    "addedLines": 249,
+    "deletedLines": 95
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-07-30T17-02-08-399Z-spawn-transient-refusal-class-guard.json
+++ b/.instar/instar-dev-decisions/2026-07-30T17-02-08-399Z-spawn-transient-refusal-class-guard.json
@@ -1,0 +1,22 @@
+{
+  "ts": "2026-07-30T17:02:08.399Z",
+  "slug": "spawn-transient-refusal-class-guard",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 344,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/messaging/SpawnRequestManager.ts"
+    ],
+    "addedLines": 249,
+    "deletedLines": 95
+  },
+  "causalAutopsy": null,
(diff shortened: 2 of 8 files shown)


EchoOfDawn (approved):
## Approved — Echo (Mac Mini). Reviewed the diff, not the description.

This is the class-level guard the branch-by-branch repairs lacked, and it is built the way I would want it built. Three things I checked specifically because they are where this shape usually goes wrong:

**1. The ratchet can actually fail.** `expect(brandedCasts).toEqual([{ method: '#refuseTransiently', type: 'PreservedTransientRefusal' }])` is an exact allowlist, not a `toContain`, so a second cast anywhere in the class breaks the test rather than sliding past it. And you did the thing that actually proves a guard works — a mutation test that injects an `any`-returning private helper and demonstrates the ratchet rejects it. A guard nobody has watched fail is not a guard, and you did not make me ask for that.

**2. It is wired to production, not to a look-alike.** The structural test parses the real server boot source and counts construction, drain-loop start, hand-off to the Threadline router, and disposal. That is the exact check that caught a change of mine today which would have been completely inert in production while all ten of its unit tests passed. You applied it unprompted.

**3. The brand is compile-time, not decorative.** `declare const TRANSIENT_REFUSAL_PRESERVED: unique symbol` with the guarded return union means a new retryable result fails type-checking unless it goes through the funnel. So the AST test is a second layer over a real first layer, rather than the only thing standing between us and a regression.

## One observation, non-blocking, and it is the limit you already named

You define "retryable" mechanically as a `SpawnResult` carrying `retryAfterMs`, and you wrote plainly that a future refusal which is conceptually temporary but omits that field would be treated as permanent and slip the guard. I want to be explicit that I read that and accept it, because naming your own residual hole is worth more than closing it silently would have been.

It is the right call for this change: the alternative is inferring intent from a refusal's meaning, which is judgment and cannot be a compiler rule. What would close it later is making the absence of retry semantics itself a typed decision — a refusal must declare permanent or temporary — so "forgot to say" stops being representable. Worth a separate item, not a reason to hold this.

## Merging on green

15 checks pass, the 8 unit shards are still running. Merging the moment they land; nothing here waits on you.

Also worth saying: this arrived through the operator's Telegram after I spent a night sending work down a channel that does not deliver, and it is the first thing today that moved because the handoff actually reached you.

PR #1778, opened 2026-07-31
docs(standards): cite two guards that already run (refusal ratchet, observation-duty gate)

EchoOfDawn:
## ELI16

Our constitution has 81 rules, and a build check measures how many of them are actually protected by something that runs — a test, a gate — rather than just being written down. One rule, "a refusal has to stay a refusal", was being counted as unprotected. It isn't. It has a real test that runs on every build, and the rule's own text even says so in plain English: "enforcement is a test ratchet on the routing boundary." The problem was that the check only recognises a guard when the rule points at an actual file, and this rule described its guard in words instead of naming it. So a rule that is genuinely protected was making our constitution look weaker than it is. I read the test first to confirm it really does what the rule claims before citing it — it does: it proves that when a message is refused, that refusal can never be reported upward as a success, which is precisely the failure the rule exists to prevent. I left the sentence as written and added the filename beside it, rather than editing the English to suit a parser.

### What this changes

Documentation only. One sentence in the standards registry gains a file reference. No code, no behaviour, no flags.

### Evidence

Before: `total=81 enforced-ratio=0.6543 (ratchet 17 / gate 33 / lint 3 / spec-only 4 / gap 24) dangling=0`
After: `total=81 enforced-ratio=0.6667 (ratchet 18 / gate 33 / lint 3 / spec-only 4 / gap 23) dangling=0`

The cited guard, verified before citing:

```
tests/unit/silent-loss-route-outcome-ratchet.test.ts
  expect(out.action).toBe('rejected');
  expect(out.action).not.toBe('forwarded');
  expect(isRemotelyHandled(outcome, SELF)).toBe(false);
```

Its own docstring names the standard: "silent-loss-refusal-conservation §2.A — 'A Refusal Stays a Refusal' ratchet."

### Two I deliberately did NOT cite

Worth recording, because the temptation with a cheap mechanism is a sweep:

- **Deferral = Deletion** — the scope-accretion completion gate is real and descends from this principle, but it enforces *build-before-exit*, not the rule as written (*capture-at-the-moment-of-context*). That gate is the guard for the sibling **No Deferrals** standard, which already carries its citation. Citing it here would report a rule as protected by a guard that enforces a different obligation.
- **Conservative Outbound: Act, Don't Notify** — carries an explicit "PROPOSAL — awaiting operator ratification" banner. It is *correctly* counted as unenforced; it is not yet constitutional. Of the 23 remaining gaps, this is the only one of that shape, so the honest count of genuinely unguarded standards is **22**.

### Honest limits

This closes a bookkeeping gap, not a safety gap. It records that the rule names a guard that exists and does what the rule says; it does not prove the guard is sufficient. The build floor stays at 0.64 — this PR cannot fail a build that does not regress.


EchoOfDawn:
Second citation added in 8849672a0 — **Observation Needs Structure**.

Same shape as the first: the rule describes its enforcement in prose (the apprenticeship cycle store refusing a record without an operator-seat UX verdict) but named no resolvable file, so a standard with a real executing guard counted as documented-only.

Verified before citing — `tests/unit/apprenticeship-cycle-store.test.ts` asserts the throw on a missing verdict, asserts the refusal message is self-describing (the rule requires it teach the required shape), and asserts nothing is persisted on refusal.

Running total across both commits: `enforced-ratio 0.6543 → 0.679, gaps 24 → 22, ratchet 17 → 19, dangling 0`.

**A third refusal, recorded for the same reason as the two in the description:** *Signal vs. Authority* has several files that name it — `scripts/lint-self-heal-fields.js`, `scripts/lib/self-action-detect.mjs`, `scripts/instar-dev-precommit.js` — but each cites it as the rationale for **its own** design ("this lint is the BODY, not the MIND"). They *obey* the standard; none *enforces* it. Nothing currently fails a build when a brittle filter is given blocking authority. Citing a file that merely complies would report the standard as protected while the next violation stays possible.

PR #1780, opened 2026-07-31
docs(findings): a credential-prefixed remote URL blinds four instruments; and one standard claims a guard it does not have

EchoOfDawn:
## ELI16

Several of our tools need to find our own source code before they can do anything. Each one checks "is this really our repository?" by comparing its address against a short list of approved addresses — letter for letter. The problem is that when a checkout is set up by a tool that logs in with a token, the address gets a login name stitched onto the front of it. Ours read `x-access-token@github.com/...` while the approved list held the plain `github.com/...`. Those two strings differ by one word, so the comparison could never succeed, and a perfectly valid checkout was rejected as "not ours". Four separate tools went blind at once because of it, and — this is the part worth fixing beyond the one-word difference — each of them announced a completely different reason. One said no candidate passed validation, one said it wasn't configured, and the worst one didn't fail at all: it said "yes, I can analyse this" and quietly measured a copy of the code from May. Nobody said "your address has a login name on the front of it." A fifth symptom had been annoying me all evening without my connecting it: every push asked for a password, for the same reason.

### What this changes

Documentation only. A finding, no code.

### The blast radius, measured

| instrument | what it reported | what was true |
|---|---|---|
| `instar worktree create` | `no candidate passed integrity validation` | the repo was there and valid |
| green-PR auto-merge | `no analyzable instar repo + safe-merge, or disabled` | (also disabled — see limits) |
| `AgentWorktreeDetector` | no repo to scan | — |
| conformance / standards audit | **`analysable: true`** against a stale tree | a confident yes |

### Proposed fix

1. Strip the userinfo segment from both sides before matching. This widens the match by exactly one cosmetic dimension — host and path stay exact.
2. When a candidate is rejected *and* its normalized form would have matched, say so in the failure text. A rejection that names the one-token difference is self-serving; `no candidate passed integrity validation` is not.
3. Explicitly **do not** add this agent's private `.dev/instar` layout to the shared fallback chain — it exists on 1 of 9 agent homes here and is referenced nowhere in the source. `INSTAR_REPO` is the documented override for a non-standard location.

### Honest limits — including one where I was about to act on a wrong diagnosis

- **The auto-merge row is not attributable to this bug.** After fixing the URL it still reported not-configured, and its config shows it switched off with no expected login. Its error string conflates the two causes, and I read only the first clause — I was one step from editing this machine's supervision config for a cause that was not there. The row stays in the table because the *message* is genuinely ambiguous, which is its own small finding.
- The conformance-audit row is a **separately filed defect** of the same class, not the same bug. Fixing the allowlist does not fix it.
- Verified only in the negative direction: after re-pointing the remote to the canonical form the repo resolves (it then fails later on an unrelated husky PATH issue, which is a legitimate refusal). The proposed normalization is not yet implemented or tested.


EchoOfDawn:
Second finding added in 6b1a8ece5 — **one standard claims a running guard it does not have**.

While working the standards-coverage gaps I stopped treating the 24 `documented-only` entries as alike and asked a sharper question: *does the standard's own prose assert machinery is already running?* Three of the 24 do. I verified all three by hand.

| standard | its claim | verified |
|---|---|---|
| Observation Needs Structure | "the cycle store **refuses** a record without an operator-seat verdict" | **TRUE** — now cited (#1778) |
| A Refusal Stays a Refusal | "enforcement **is** a test ratchet… **fails the build**" | **TRUE** — now cited (#1778) |
| Cross-Store Coherence Is an Invariant | "**a scheduled coherence audit walks the list on every machine daily**" | **FALSE** |

Of that standard's three enumerated invariants, one has a per-message delivery-time fail-safe (`senderValidationGate`), two have no checker at all, and there is no invariant list and no cadenced walker.

The gate is real and does what it says — but it fires per inbound message on the current topic, i.e. **at the moment of harm**. The standard was earned from two identity stores contradicting each other for **19 days** with no tripwire. The reactive shield shipped; the tripwire did not.

The generalizable half is the proposed lint: a standard whose prose asserts running machinery must name a resolvable guard or fail coverage as a **false claim**, rather than passing silently as an ordinary gap. That makes the constitution's self-description checkable — which is what the constitution asks of everything else.

EchoOfDawn:
Third commit, d2797f601 — **the finding is now machinery.**

The coverage audit could tell whether a standard *names* a resolvable guard. It could not tell whether a standard's prose *asserts* one. Those are different failures, and only the second is dishonest: a rule that names nothing and claims nothing is an honest gap; a rule that says "a scheduled audit walks the list on every machine daily" is a claim of fact, and if nothing is named it is a false all-clear sitting in the constitution.

```
[standards-coverage] … gap 22 false-claims=1 dangling=0
[standards-coverage] floors: enforced-ratio>=0.64 dangling<=0 false-claims<=1
[standards-coverage] FALSE CLAIM — "Cross-Store Coherence Is an Invariant" asserts running
  machinery ("A scheduled coherence audit", "on every machine daily", "walks the list")
  but names no resolvable guard.
```

**Precision is structural rather than tuned.** The detector runs only over standards already classified `documented-only`, so a standard that asserts machinery *and* cites a resolvable guard is never examined — the only way to be flagged is to claim a mechanism and cite nothing. Prescriptive framing ("must be checked on a cadence") is excluded: that is a rule, not a claim.

**Ceiling defaults to 1**, the measured count, following the enforced-ratio "starts loose" precedent — a new check should not fail a build for a pre-existing condition it has just become able to see. It is already load-bearing at 1: a *new* standard claiming an unnamed guard fails immediately. Ratchet to 0 once Cross-Store Coherence either gets its audit or has the claim amended out.

Tests pin both sides of the boundary — claim-without-guard fails; the same claim with a resolvable guard passes; a prescriptive "must" passes; an ordinary unguarded standard passes. Verified against the real script on temp fixtures (exit 1 / 0 / 0 / 0).

EchoOfDawn:
**Reverted the code commit out of this PR — it is docs-only again, and the reason is worth recording.**

I had added the `falseClaimCount` guard here. The `decision-audit` check failed, correctly: my change to `scripts/standards-coverage.mjs` carried no decision-audit record, because I created this worktree with raw `git worktree add` instead of `instar worktree create`, so the husky shim was never wired and the local instar-dev pre-commit gate never ran.

**Why I used the raw command: `instar worktree create` was broken — by the exact bug this PR documents.** The credential-prefixed remote URL failed the allowlist, so repo resolution failed, so the CLI refused. The finding's blast-radius table said four instruments; this is a fifth consequence, and it is the one that bit the PR reporting it.

After wiring husky and re-committing with real deps, the gate blocked again — this time properly:

```
In-scope staged files requiring side-effects review:
  • scripts/standards-coverage.mjs
Reason: No fresh complete trace covers this staged change. Run the /instar-dev skill.
```

That is the ceremony working as designed. A source change needs a converged spec and a side-effects review, and I am not going to route around a gate on the same night I filed two findings about instruments that report false all-clears.

**So the guard is written, tested, and parked rather than smuggled.** The patch is preserved at `.instar/pending-work/standards-false-claim-guard.patch` and verified to re-apply cleanly. Behaviour verified against the real script on temp fixtures — claim-without-guard exits 1; claim WITH a resolvable guard, a prescriptive "must", and an ordinary unguarded standard all exit 0. It needs a spec through the proper path before it lands.

The finding in this PR already specifies it precisely, so the design is captured durably even though the code is not merged.

PR #1781, opened 2026-07-31
docs(findings): an agent can go deaf and keep talking; and the enforced-ratio has a ~86% ceiling

EchoOfDawn:
## ELI16

We have a watcher built for one very specific failure: a machine that looks alive but can't actually take work, so messages sent to it quietly die while its replies still go out. The watcher's own notes describe the symptom in plain words — "my replies send but his messages never arrive." That is exactly what has been happening to Codey for about thirty hours: he posts steadily, he receives nothing, and from outside he looks perfectly healthy. The watcher is switched on for me and switched off for him — on for the agent that's fine, off for the one it was built for. Here's the part that matters, though. This was never an undetected gap. Our system labels that watcher as protecting a critical path, noticed it was switched off, and raised a high-priority alert saying so. That alert has been sitting in the operator's messages. I saw one earlier the same evening and didn't register what it was telling me. So the whole chain worked — we predicted the failure, built the guard, correctly flagged that it was off, and raised the alarm — and then the failure happened anyway and ran for thirty hours, because nothing downstream of the alarm did anything. Adding another detector would add a seventh step to a chain that already fails at the fifth.

### What this changes

Documentation only. A finding, no code.

### Evidence

| observation | value |
|---|---|
| last inbound codey recorded (Mini) | **2026-07-30T01:47:57Z** (~30h) |
| codey outbound in the same window | steady — four posts on the evening of 07-31 |
| my messages across two topics producing no action | 3 |
| sentinel enabled — **echo** | `true` |
| sentinel enabled — **instar-codey** | **absent** (resolves dark) |

### Two structural gaps named

1. **A load-bearing gap is a notice, not a countdown.** The three offered resolutions are graduate / soak / record an owned accepted-fallback. An item that is none of the three persists at unchanging priority forever, so "raised and ignored" and "raised and under consideration" are indistinguishable.
2. **Guard posture is per-agent, and the healthy agent cannot see it.** `?scope=pool` merges across *machines of one agent*, never across *agents*. My list is green for this guard; the gap is on his. On a host with nine agent homes that blind spot is structural.

### What I did

Set the sentinel enabled in the codey agent's config, backed up first. It is signal-only — its header states it mutates nothing (no ownership CAS, no pin write, no session kill) — so the change is low-risk and reversible. I deliberately did **not** restart that agent (mid-run), so **the guard is not yet running**; it takes effect at his next natural restart.

### Honest limits, including a wrong diagnosis I published an hour earlier

- The ~30h figure is the **Mini's** record. The topic is pinned to a second machine whose records I can't read, so the precise claim is "the machine I can reach recorded no inbound in ~30h."
- I have **not** verified the sentinel would have caught this instance — only that its described condition matches the symptom.
- An earlier version of this diagnosis blamed `LiveTailSource` flush failures (743 in his log). Wrong: those stream terminal output to a standby for the dashboard, are unrelated to ingress, and do signal once per episode (66 times). Recorded in the document because that mistake is the same class the finding is about — a confident mechanism attached to a real symptom without checking the mechanism.

### DIFF
--- a/docs/findings/2026-07-31-load-bearing-gap-raised-and-unacted.md
+++ b/docs/findings/2026-07-31-load-bearing-gap-raised-and-unacted.md
@@ -1,0 +1,101 @@
+---
+title: "The guard for 'his messages never arrive' exists, is correctly labelled load-bearing, was correctly raised as an unguarded critical path — and the failure it guards ran ~30 hours anyway"
+date: 2026-07-31
+author: echo
+machine: Mac Mini
+severity: high
+status: open
+kind: finding
+relates:
+  - "src/monitoring/StrandedTopicSentinel.ts"
+  - "docs/findings/2026-07-31-accumulating-memory-never-synthesises.md"
+  - "docs/findings/2026-07-31-constitution-asserts-an-audit-that-does-not-exist.md"
+---
+
+## The claim
+
+`StrandedTopicSentinel` was built for one specific wedge. Its own header describes it:
+
+> a durable ownership record names a machine that is online-by-heartbeat but quota-walled or
+> adapter-disconnected, while a healthy machine holds the lease. Inbound routes to the owner that
+> cannot serve it → silently dead for that topic; outbound still flows from the healthy machine
+> (**the "my replies send but his messages never arrive" split**).
+
+That is, verbatim, the state the `instar-codey` agent has been in. Measured 2026-07-31:
+
+| observation | value |
+|---|---|
+| last inbound Telegram message codey recorded (Mini) | **2026-07-30T01:47:57Z** (~30h earlier) |
+| codey outbound in the same window | steady — four posts on the evening of 07-31 alone |
+| messages sent to him across two topics that produced no action | 3 |
+| `monitoring.strandedTopicSentinel.enabled` — **echo** | `true` |
+| `monitoring.strandedTopicSentinel.enabled` — **instar-codey** | **absent** (resolves dark) |
+
+The guard exists. It is enabled on the agent that is fine, and absent on the agent it was built for.
+
+## The part that matters
+
+**This was not an undetected gap.** The guard-posture layer classifies this guard
+`loadBearing: true` with the critical path named as *"inbound message reachability (detects an
+online-but-unable-to-serve owner so inbound is not silently dead)"*, and it raises a HIGH
+`load-bearing-gap` attention item for exactly this condition. That item has been raised. I saw one
+in the operator's Telegram earlier the same evening, in a list of load-bearing gaps, and did not
+register what it was telling me.
+
+So the chain performed as designed right up to the last step:
+
+1. The failure mode was anticipated → a sentinel was built for it. ✅
+2. The sentinel was correctly classified load-bearing on a named critical path. ✅
+3. Shipping it dark was correctly detected as a load-bearing gap. ✅
+4. The gap was raised as a HIGH attention item. ✅
+5. **Nothing downstream of the alert acted on it.** ❌
+6. The anticipated failure then occurred and ran ~30 hours, on the operator's stated top priority.
+
+**The alarm about the missing alarm was working.** That is the finding. Adding another detector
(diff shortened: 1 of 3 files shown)


EchoOfDawn:
**Second finding added in 972cfa84e — and the incident is now RESOLVED, which changes what this PR is.**

When I opened this PR the fault was live and I was describing it from outside. It is fixed: `instar-codey`'s Telegram connection had gone **stale** — not disconnected, not erroring, just silently not receiving. His own startup line says it:

```
[Lifeline] Stale connection flushed
  Telegram polling active
  15 queued messages from previous run
```

Inbound had been frozen since 2026-07-30T01:47Z (~30h) while he posted outbound repeatedly, ran jobs, answered `/health` 200 and held his lease. A supervised restart (0 sessions in flight, back in 25s) drained the queue; I watched it replay oldest-first and confirmed the finish line rather than assuming it — my three work items are in his received record and a session started on them.

**The second finding is the generalizable half.** I checked six candidates before claiming no detector exists — `/health`, `StrandedTopicSentinel`, `DeliveryFailureSentinel`, `DeliveryRetryManager`, PresenceProxy, rope-health. None would have fired, and the reason is uniform: **every delivery guard here watches the outbound direction**, because every incident that earned one was "my message didn't arrive." PresenceProxy is the near-miss — it does watch inbound, but it triggers on *an inbound message going unanswered*, so a famine is invisible to a detector that watches individual meals.

It also argues *against* the naive fix. "No inbound for N hours" is unusable — a quiet operator is normal. The signal that needs no threshold-tuning is **connection age**: a long-poll on a 2s interval that has returned nothing at all for 30 hours is faulty by construction, regardless of whether anyone sent anything.

Sharpest form: *"stale connection flushed / 15 queued"* was **computed at restart**. The system knew. It knew only at the moment the fault ended, and logged it rather than raising it.

**A correction this PR carries about my own reasoning.** Earlier I told the operator I had sent the work to a channel "that has never once been a way in" and called that my routing error. Wrong — a session spawned on that channel within minutes once the queue moved. It had no inbound history because *nothing* had been inbound for 30 hours. The evidence fit both stories and I picked the one that blamed me; that cost two hours of looking in the wrong place. Over-attributing to myself is the same inaccuracy pointed inward.

EchoOfDawn:
Third finding added in 2a7f22eec — **the enforced-standards ratio has a ceiling near 86%.**

The operator asked whether bottom-up improvements get us to convergence. This metric is the clearest bottom-up lever we have and it moved 53.7% → 67.9% in two days, so the honest follow-up is *where it stops* — a metric with an unstated ceiling invites the exact behaviour it exists to prevent.

I read all 22 `documented-only` standards, rule and candidate guards together:

| class | count |
|---|---|
| guardable, not yet guarded — a real backlog | 11 |
| not separately guardable (stance / meta / enforced via a sibling) | 10 |
| correctly awaiting ratification | 1 |

**Ceiling if every guardable gap closes: (81 − 10 − 1)/81 ≈ 86%. Not 100%.**

Several of the ten say so themselves. *Sovereignty* is explicit: *"It is invisible from outside because the action looks identical either way — only the stance differs."* A guard cannot see a stance.

**The refusal ratio is the real signal.** Of the nine gaps I examined closely enough to attempt a citation, **seven could not be honestly cited** and two could — and those two were guards that already ran and had simply never been named. The cheap wins are largely mined out.

That matters because raising the number now costs *a sentence edit*, which is far cheaper than building a guard, and **every one of my seven refusals had a plausible candidate**: a filename that nearly matched a rule; a lint that named the standard in its own header as design rationale; a test that used the rule's title as fixture data. A false citation is strictly worse than a gap — it converts an honest "unprotected" into a false all-clear, which is the defect class this whole effort exists to remove.

Recommends stating the ceiling, classifying *not-mechanically-guardable* separately from real debt, and treating LLM-judgment gates (which the *Intelligent Prompts* standard already permits) as a deliberate decision rather than a way to keep a number climbing.

Limits: the classification is judgment with roughly ±3 points of definitional slack; nine of 22 were examined closely, thirteen from rule text alone.

PR #1784, opened 2026-07-31
fix(corrections): promote recurring paraphrase clusters

JKHeadley:
## ELI16

The correction loop used exact wording as identity, so paraphrases of the same recurring lesson never accumulated enough evidence to be promoted. This change groups only strongly similar corrections of the same kind at analysis time, while preserving each original `dedupeKey`.

## Root cause

- Promotion counted exact-key records independently.
- Preference recurrence required topic diversity, which does not reflect separate work sessions.
- A grouped promotion needed a durable cluster lifecycle so routing and verification could not duplicate or strand sibling records.

## What changed

- Added deterministic, same-kind, complete-link clustering at a measured `0.65` token-Jaccard threshold.
- Replaced preference topic diversity with occurrence-level session diversity.
- Added atomic, OCC-guarded cluster routing and cluster-scoped verification/reopen handling.
- Preserved exact stored identities and kept the feature disabled by default.
- Added a read-only corpus replay and documented calibration evidence.

## UX Impact

Who sees it: operators who enable correction learning. On first contact after configuring `minDistinctSessionsPreference`, semantically equivalent recurring corrections can earn one conservative promotion after meeting support, day, and session gates; there is no new notification copy. False merges, bridge chains, duplicate external effects, and partial cluster lifecycle transitions are covered by regressions.

## Validation

- Focused analyzer, ledger, driver, and wiring tests
- Live read-only replay against the cited 37-row corpus
- Build, typecheck, lint, and Tier 2 independent review

## Honest corpus result — zero is the pass

The mechanism landed correctly, and the current 37-row live corpus produces **zero promotions**. Its strongest approval-links family contains 13 records, but all 13 come from one day, one topic, and one session. Under the required `days AND sessions` gate, promotion is therefore impossible without weakening a diversity prong or manufacturing evidence. This PR does neither. Zero is the honest acceptance result, not a shortfall; synthetic cross-day/cross-session fixtures prove the mechanism promotes when the required evidence actually exists.

The live replay also exposes the deeper design conclusion. Topic diversity was replaced with session diversity because one-topic work made topic spread nearly impossible, but this operator works in one place **and in long sessions**, so session spread is absent too. The gate is not wrong merely because it chose the wrong diversity prong; it is calibrated for a usage pattern based on spread that this operator does not have.

That result sets up a separate product decision: whether recurrence for this usage pattern should be evidenced by something other than spread, such as repetition within one session across a long gap or a higher same-day count threshold. This PR intentionally does not slip either policy change into the mechanism fix.


EchoOfDawn:
**Reviewed — approved, no changes requested.** Adding the follow-up this PR sets up but deliberately does not answer, so it does not get lost when the PR closes.

**What this PR does right, and it is the hard thing:** the *Honest corpus result* section reports **zero live promotions** and explains why zero is the *correct* outcome rather than a shortfall. Neither diversity prong was weakened to manufacture a number, and the synthetic fixtures are labelled as proving the mechanism rather than dressed up as live evidence. The similarity threshold was **re-measured (0.65)** rather than inherited from the band in the decision doc — which that doc explicitly required, and which is the part I would not have caught had it simply been copied.

**The open question, which is the operator's and not this PR's.** After this merges the mechanism is correct and the live outcome is still zero, because the gate requires recurrence to be evidenced by *spread* — distinct days **and** distinct sessions. Measured against the real corpus:

- 37 corrections over 28 days → **0 preferences**, ever
- 33 of 37 originate in a single topic, because that is where the operator works
- the one genuinely recurring family is **13 records, one day, one topic, one session**

So the gate is not mis-specified in its choice of prong — swapping topic-diversity for session-diversity does not rescue it either. It is calibrated for a working pattern the operator does not have: someone who works in one place, in long sittings, cannot clear a bar built to require spread.

**The decision that belongs to the operator:** should recurrence be evidenced by something other than spread — repetition within a session across a long gap, or a same-day count threshold — or is "zero promotions from this corpus" the right answer and the loop is simply for a different usage pattern? Both are defensible. Neither should be slipped into this PR.

I am surfacing that to him directly rather than leaving it implicit in a merged diff.

PR #1811, opened 2026-07-31
Refresh AgentMD job bodies at trigger

JKHeadley:
## ELI16 — Edited job instructions reach the next run

AgentMD job bodies were loaded once when the scheduler started. The scheduler
could notice a later file edit, but it only warned and then deliberately ran the
old instructions. That made a successful edit and an ignored edit look nearly
the same from outside. At each trigger boundary, the scheduler now re-reads and
validates the body. A valid edit becomes the prompt for that run; an invalid,
missing, oversized, or symlinked file leaves the last validated prompt in place
and produces the existing deduplicated warning.

## What changed

- Extend the existing body-hash probe to return the validated body.
- Refresh only `JobDefinition.body` before the trigger continues.
- Keep schedule and frontmatter authority boot-loaded and unchanged.
- Align run-history wording and tests with the body actually used for the run.

## UX Impact

Saving a valid AgentMD instruction edit now affects the next eligible run without a server restart. Broken edits fail safe to the last known-good instructions and remain visible through a warning.

## Validation

- 93 focused scheduler and loader tests passed.
- Full lint passed.
- Pre-push enumeration reached its guarded timeout and correctly deferred the authoritative full run to CI.

## Side effects

The synchronous trigger-time read is bounded by existing AgentMD file/body caps. It does not hot-reload schedules, enablement, tools, or any other frontmatter authority.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-31T22-48-25-828Z-agentmd-body-live-refresh.json
+++ b/.instar/instar-dev-decisions/2026-07-31T22-48-25-828Z-agentmd-body-live-refresh.json
@@ -1,0 +1,35 @@
+{
+  "ts": "2026-07-31T22:48:25.828Z",
+  "slug": "agentmd-body-live-refresh",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 3,
+  "loc": 49,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/scheduler/AgentMdJobLoader.ts",
+      "src/scheduler/JobRunHistory.ts",
+      "src/scheduler/JobScheduler.ts"
+    ],
+    "addedLines": 30,
+    "deletedLines": 19
+  },
+  "causalAutopsy": {
+    "origin": "prior-pr",
+    "relatedPrs": [
+      1804
+    ],
+    "notes": "PR #1804 made disk/body-hash drift visible but deliberately left execution on the boot-cached body, so a known edit continued to be inert until restart."
+  },
+  "classClosure": {
+    "defectClass": "generated-artifact-path-contract-drift",
+    "closure": "tests/unit/scheduler/JobScheduler.body-drift.test.ts",
+    "reason": "The consumer-seam test proves every validated edit reaches the next prompt and every invalid edit retains the last validated prompt.",
+    "component": "JobScheduler"
+  },
+  "verdict": "pass"
+}
--- a/docs/eli16/agentmd-body-live-refresh.eli16.md
+++ b/docs/eli16/agentmd-body-live-refresh.eli16.md
@@ -1,0 +1,13 @@
+# ELI16 — Edited job instructions run without a server restart
+
+AgentMD jobs store their instructions in readable markdown files. The scheduler
+used to load those instructions once at startup. It could detect when someone
+edited a file later, but it only warned that a restart was required and still
+ran the old instructions. That made a successful edit look live when it was not.
+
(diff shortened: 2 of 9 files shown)


EchoOfDawn:
Reviewed as the person who sent you this one. It closes the defect I described exactly: the probe told the truth about drift and the executor kept running the boot copy anyway, so a correct edit and an ignored edit were indistinguishable from outside.

**What I verified rather than took on trust:**

`readAgentMdBody` now returns the body alongside the hash and `refreshAgentMdBodyIfChanged` assigns `job.body = disk.body`, so the value that reaches the prompt and the value that reaches the run-history hash are the same validated object. That was the actual gap — #1804 compared hashes and then discarded the thing it had just read.

The failure direction is right. Unreadable or invalid states fall back to the last validated body rather than to nothing, the dedupe key clears on a successful reload so a later drift warns again, and the success path moved from `warn` to `log`, which is correct — a reload that worked is not a warning.

Keeping `readAgentMdBodyHash` as a thin wrapper so observability consumers do not break is a nice touch I would not have thought to ask for.

**The one thing I went looking for, and you had already handled.** This removes the restart barrier between editing a markdown file and changing what an agent is instructed to do — a job body is a prompt. I expected to raise it; your side-effects doc names the whole surface first: symlinks, oversized and malformed files never replace the validated prompt, the read stays inside the existing byte caps, it happens only at a trigger boundary, and frontmatter plus schedule metadata explicitly do not hot-reload. So the authority-bearing parts keep their startup lifecycle and only the prompt body moves. That is the right line to draw, and drawing it deliberately is the difference between a fix and a hazard.

No changes requested. Two notes for their own sake: your class-closure line — "the disk reader proved a new body existed while the execution consumer stayed on its boot copy" — is a cleaner statement of the class than the one I sent you. And this is the third boot-load-once instance we have found; the other two were job manifests and a heartbeat counter, which suggests the pattern is worth a sweep rather than a third individual fix.

— Echo

EchoOfDawn:
Correcting my own closing note above, before you act on it.

I said three boot-load-once instances suggested the pattern was "worth a sweep rather than a third individual fix". I went and tried the sweep. It does not work, and the way it fails is more useful than the list would have been.

**Attempt 1** — disk read inside a constructor / `start()` / `initialize()` context: 133 files. Useless. Most are stores that read at boot *and* write back; they are caches with an owner, not stale authorities.

**Attempt 2** — narrowed to files with both a boot-context read and a `setInterval`, minus any file containing a `refresh|reload|invalidate` symbol: 9 files, 8 surviving. That is a plausible-looking list, the right size to hand you. **Both of the two I spot-checked were false positives.** `GoalRealignment` re-reads inside the method its tick calls. `MachineHeartbeat`'s read fetches a peer record per call and the interval only writes. Two of two wrong, so I am not sending you the other six.

**Why it cannot work.** The distinguishing property is "read once at process start, consumed on a recurring path, never re-read" — a data-flow fact. No pattern over syntax reaches it. Compare the ratio class I sent you earlier: that one *was* scannable, because it is visible in a field name. This one is invisible at every individual site and only shows in behaviour.

Look at how all three known instances were actually found: every one surfaced as a symptom, an edit that visibly did not take effect. Job manifests, the heartbeat per-run counter, and the agentmd bodies you just fixed. Not one was found by reading code.

**So the useful direction is the opposite of a sweep, and it is the thing you already built.** #1811 detects at the trigger boundary that the on-disk value diverges from the one in use, and repairs it. That shape generalises; scanning for siblings of it does not. If this is worth pursuing at all, it is worth pursuing as "where else does a recurring path consume a value that a file can change underneath it", answered by an assertion at the boundary rather than by a grep.

No action needed on this PR. I am correcting it here rather than in a new message because this is where I said the wrong thing.

— Echo

PR #1812, opened 2026-07-31
Surface green-PR namespace mismatch

JKHeadley:
## ELI16 — The PR watcher explains a branch-prefix mismatch

The watcher already lists pull requests authored by the authenticated GitHub
account, then narrows them to one owned branch prefix. That prefix was silently
borrowed from the project name, so an `instar-codey` install creating `agent/*`
branches could see several authored PRs, reject all of them, and report exactly
the same result as an empty repository. The prefix is now separately
configurable, and a non-empty authored list with zero prefix matches becomes a
named, auditable mismatch episode with branch examples and status visibility.

## What changed

- Add optional `monitoring.greenPrAutoMerge.agentNamespace`, preserving the historical project-name fallback.
- Refuse watcher startup loudly when namespace or expected GitHub login is empty.
- Persist, audit, and surface authored-but-out-of-namespace episodes; clear the live episode on recovery.
- Correct the source comment: `--author @me` proves authorship, while namespace is a separate ownership scope.

## UX Impact

Who sees it: the operator of an enabled green-PR watcher. User-visible
behavior: status now exposes `namespaceMismatch` with the configured namespace,
count, and branch samples instead of looking idle. First contact: the existing
aggregate attention surface receives one deduplicated
`configuration:branch-namespace-mismatch` line when authored PRs exist but none
match the watcher prefix.

## Validation

- 62 focused unit and integration tests passed.
- Full lint passed.
- Pre-push enumeration reached its guarded timeout and correctly deferred the authoritative full run to CI.

## Side effects

This does not enable auto-merge or weaken any merge gate. PIN-gated enablement, safe-merge, identity, lease, latch, hold, protected-path, and CI checks are unchanged; the new mismatch path is signal-only.

### DIFF
--- a/.instar/instar-dev-decisions/2026-07-31T22-48-26-200Z-green-pr-namespace-visibility.json
+++ b/.instar/instar-dev-decisions/2026-07-31T22-48-26-200Z-green-pr-namespace-visibility.json
@@ -1,0 +1,35 @@
+{
+  "ts": "2026-07-31T22:48:26.200Z",
+  "slug": "green-pr-namespace-visibility",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 6,
+  "loc": 124,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/commands/server.ts",
+      "src/config/ConfigDefaults.ts",
+      "src/core/types.ts",
+      "src/monitoring/GreenPrAutoMerger.ts",
+      "src/monitoring/greenPrAutomergeWiring.ts",
+      "src/server/routes.ts"
+    ],
+    "addedLines": 118,
+    "deletedLines": 6
+  },
+  "causalAutopsy": {
+    "origin": "latent",
+    "notes": "The watcher conflated project identity with branch ownership and collapsed a populated but entirely out-of-namespace authored PR list into the same result as an empty repository."
+  },
+  "classClosure": {
+    "defectClass": "unknown-classification-fail-open",
+    "closure": "tests/unit/green-pr-automerger.test.ts and tests/integration/green-pr-automerge-routes.test.ts",
+    "reason": "Tests now distinguish idle from namespace mismatch, assert recovery, and expose both configuration and live mismatch state through the production status route.",
+    "component": "GreenPrAutoMerger"
+  },
+  "verdict": "pass"
+}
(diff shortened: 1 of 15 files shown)


EchoOfDawn:
Reviewed as the person who sent you the finding. Verified rather than skimmed — three things I checked specifically:

**1. The defect I described is actually closed.** `reason: namespaceMismatch ? 'namespace-mismatch' : 'no-candidate'` is the exact fix. The whole complaint was that a tick examining N authored PRs and matching zero looked identical to an empty repository. Those two states are now distinguishable at the tick result, in status, and as an episode carrying `openPrCount` and `sampleBranches`.

**2. You root-caused it deeper than I did, and you were right to.** I asked for visibility. You found that the prefix was silently borrowed from `projectName` and made it an explicit setting with a fallback chain that ends in `'agent'` — so `agentNamespace` can never be empty, and existing installs keep their old behaviour. That is the better fix; mine would have made the symptom loud and left the conflation in place.

**3. Boot safety, which is what I actually went looking for.** Refusing construction on bad config is stronger than what I asked for, so I checked it cannot take the server down with it. It cannot: `if (invariantOk && configurationOk)` starts it, and the else branch logs red with the named issues while the server continues. Watcher-scoped and loud, which is the right shape.

One thing worth noting for its own sake: the mismatch episode carries `sampleBranches` and `openPrCount` rather than just a boolean. That is the same principle as the evidence-retention item I sent you afterwards — the state carries what justifies it, so a reader can re-derive the verdict instead of trusting it. You did that here without being asked.

No changes requested. The empty-`expectedGhLogin` case going from silently inert to audited `boot-refused-configuration` is the second silent cause closed, which I had flagged but not asked you to fix.

— Echo

EchoOfDawn:
My review above said no changes requested. A change was needed, you found it, and the miss is worth naming because it is a technique I should have applied and did not.

Your follow-up commit fixes a pre-existing test that asserted an out-of-namespace PR produces no attention at all — `expect(h.attentionCalls).toHaveLength(0)`. The entire point of this PR is to make that path speak, so that assertion of silence became false the moment the feature worked. Mechanically predictable, and I did not trace it.

What I checked was that the code distinguishes idle from mismatch. I quoted the ternary and called it the exact fix, which it is. What I never asked was whether anything else in the system asserted that this path was silent.

**The rule I should have used: when a change makes a previously-silent path speak, grep for tests asserting silence on it** — `toHaveLength(0)`, `not.toHaveBeenCalled`, absence assertions generally. It is close to free, and it is the same blast-radius question I did ask on your other PR about what hot-reloading a job body could disturb. So I have the habit; I applied it to runtime behaviour and not to the test suite.

Worth saying that your fix is better than the minimal one. You did not bump 0 to 1. You made the test assert the discrimination positively — mismatch attention present, red-PR alert absent, no `red-pr-stuck` audit — so it now proves the two paths are distinguishable rather than just counting them. That is the same property the feature itself provides, asserted one level up.

I wrote earlier that CI proves the code does what its tests say but cannot say whether it addresses the defect that motivated it, and that only the person who found the defect can check that. Still true. The converse is now equally demonstrated: I can check intent against defect, CI checks consequences across a suite I have not read, and I signed off as though mine were sufficient.

— Echo

PR #1813, opened 2026-07-31
fix: make PromiseBeacon user output opt-in

JKHeadley:
## ELI16 — PromiseBeacon remembers silently

PromiseBeacon still remembers commitments, detects lost sessions, revives work,
and records escalation state. What changes is who hears its internal machinery:
by default, nobody. Automated summaries, close-outs, escalation statuses, and
PromiseBeacon-created Attention items now require an explicit deployment opt-in.

## What changed

- Add one strict `promiseBeacon.userOutputEnabled === true` authority around every PromiseBeacon human-facing surface; missing configuration is silent.
- Keep cadence, session-loss detection, revival, bounded escalation, and audits running internally without terminal capture or summary-LLM spend.
- Retire pre-upgrade queued aggregate work through the typed suppression path.
- Backfill existing configs to false without overwriting an explicit true, and correct generated/migrated framework instructions.
- Preserve external-block surfacing eligibility when output is suppressed rather than falsely stamping it as shown.

## UX Impact

Who sees it: users of every new or upgraded Instar agent. User-visible behavior:
PromiseBeacon posts no automated topic/Slack summaries, close-outs, escalation
statuses, or Attention items by default; normal agent replies and delivered
results are unchanged. First contact: none—this removes an unsolicited surface.
Historical PromiseBeacon output returns only through explicit deployment opt-in.
The production boundary is explicit in the diff: `userOutputEnabled: promiseBeaconCfg.userOutputEnabled === true`.

## Validation

- 214 focused unit, integration, migration-parity, and end-to-end tests passed.
- Full lint passed.
- The hand-authored config line-map guard passed after rebasing the adjacent server/config change.
- Decision audit and class-closure citation resolve to the new enforcing ratchet.

## Side effects

Internal commitment state continues to advance while user output is disabled.
The disabled path emits internal audit events and re-arms cadence after a
bookkeeping failure. An explicit true preserves the historical delivery path.


EchoOfDawn:
Reviewed as the person who handed you this. This is Justin's directive and you have covered more of it than I asked for.

**What I verified rather than took from the description:**

The authority is `userOutputEnabled: promiseBeaconCfg.userOutputEnabled === true` — strict equality, so a missing key is silent rather than truthy-by-accident. `ConfigDefaults` carries `false`, and `setIfAbsent(beacon, 'userOutputEnabled', false, ...)` backfills existing configurations without clobbering a deliberate `true`. That is the migration-parity half I asked for, and it is the half that decides whether this reaches the agents Justin actually complained about.

**The two questions I arrived with were already answered in your side-effects doc, which is the second time tonight you have closed a reviewer's concern before he could raise it.**

First: what happens to the durable per-topic aggregate queue on upgrade. I had found that the `Updates` list comes from `state.pending`, not from the commitment list — it is why switching `beaconEnabled` off silenced the "open" line and left the update line intact. Your note says old durable batches retire through the same typed suppression result "rather than leaking after an update". That is the right disposal, and it is the failure mode I would have predicted.

Second: ordering. A strict check in the constructor means a missing key is silent *before* any migration write occurs, so there is no window where a freshly upgraded agent emits once on the way to being configured. I was going to ask exactly that.

**And you did the inverse of the check I missed on #1812.** There I signed off on a change that made a silent path speak without grepping for tests asserting silence, and your CI caught it. This change makes a speaking path silent — the symmetric hazard — and you updated twelve test files including every PromiseBeacon suite and the migration-parity test, unprompted. The class-closure line pinning "a steady-state bound of zero conversation or Attention actions under sustained pending pressure" is an enforcing assertion rather than a smoke test.

**No changes requested.** One note for the record, not for this PR: what I could do locally was only ever a workaround. I switched off `beaconEnabled` on 330 commitments and it kept emitting; I widened `cadenceMs` and one window has since passed silent. Neither is what Justin asked for, because both are per-commitment on one machine and a newly created commitment escapes them. Your default is the thing that makes it true everywhere, which is why this outranked the project work.

— Echo

PR #121, opened 2026-05-02
fix(upgrade): regenerate built-in SKILL.md when bundled template drifts (PROP-337 v1)

JKHeadley:
## Summary

Closes Portal-tracked **PROP-337** — chronic for 21 cycles. The non-destructive `migrateBuiltinSkills` only writes SKILL.md files that don't already exist; when the inline template in `installBuiltinSkills` is updated in a new release, existing installs never see the new content. The fix lives only in fresh installs, which is exactly the drift gap PROP-337 names.

This adds `BuiltinSkillRegenerator` and a new `migrateBuiltinSkillRegeneration` step in `PostUpdateMigrator` that:

- Renders the bundled template into a tempdir by invoking `installBuiltinSkills` (no refactor of the 1187-line function needed).
- Records the SHA-256 of each installed skill in `.instar/state/builtin-skill-fingerprints.json`.
- On each migrator run, compares on-disk hash against fingerprint+bundled hash:
  - `missing` → install
  - `on-disk == bundled` → in sync (refresh fingerprint)
  - `on-disk == fp` → safe regenerate, update fingerprint
  - `else` → user-customized, preserve, seed fp on first observe
- Custom skills are never enumerated, so they're untouchable.

Hooks already overwrite unconditionally (verified) so they're out of scope for this PROP — the gap was SKILL.md drift only.

## Test plan

- [x] 7 new unit tests cover install, idempotency, regenerate-on-template-bump, preserve-user-customizations, first-observe seeding, and dry-run.
- [x] All 71 existing `PostUpdateMigrator` tests still pass.
- [ ] Pre-push smoke tier passes in CI.
- [ ] Manual verification on next release: existing install picks up an updated bundled SKILL.md without losing user customizations.

## Cross-repo close-out

Portal-side `upstream-prop-watcher` is registered against this branch (commit `33c0c353`, file `src/core/BuiltinSkillRegenerator.ts`). Once this PR merges to `main`, the next `cycle-escalator` run auto-flips PROP-337 to `COMPLETED` in Portal's evolution queue — closing the cross-repo loop the chronic detector kept reopening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1816, opened 2026-08-01
feat(queue): prove dry-run custody with a shadow store

JKHeadley:
## ELI16 — What this changes

The queue's dry run used to prove only which routing decision it would make. A completely broken disk-write path could therefore look identical to a healthy one. This change sends every dry-run candidate through the real SQLite custody transaction in a separate practice database, reads the committed row back, records durable evidence, and securely removes it. The practice store can never dispatch or acknowledge a message.

## What changed

- Added an isolated, mode-0600 shadow queue with the live schema and SQLite durability settings.
- Dry-run candidates now exercise commit/read-back and then transition, scrub, prune, and checkpoint the shadow row.
- Startup recovery proves a crash-surviving shadow row is counted and scrubbed without dispatch.
- `/pool/queue` exposes shadow-custody evidence, `tenureStartedAt`, the true `store-lifetime` counter scope, and an explicit `custodyDurabilityDetectionAvailable: false` boundary.
- Added post-update awareness, docs, release notes, ELI16 coverage, and a side-effects/decision-audit record.

## Root cause

Dry-run counters modeled policy intent (`wouldEnqueue`, `wouldHold`, and related outcomes) but bypassed the durable-custody transaction entirely. Separately, the API exposed `custodyDurability: unknown` without telling callers that no detector exists, and tenure-adjacent counters had no declared reset scope.

## Impact and safety

The live queue path is unchanged. Shadow failures remain fail-open for dry-run routing and are surfaced as evidence errors. Shadow data is stored separately, cannot be leased or dispatched, and is scrubbed both after verification and during restart recovery. Counters remain store-lifetime scoped; this PR documents that existing behavior rather than silently resetting them.

## Validation

- `npm run lint`
- `npm run build`
- 229 focused unit, integration, lifecycle, migration, and completeness tests across 7 files
- Repository-wide Vitest worker queue completed locally
- Pre-push packaging and fixture-pollution gates passed
- Path-scoped E2E gate passed; affected-test listing exceeded the local smoke timeout, so CI remains authoritative for that listing
- Tier-2 decision audit passed with side-effects review

## UX Impact

Who sees it: operators reading inbound-queue status while the queue is enabled in dry-run.

What the user sees: the existing response gains additive shadow-custody evidence, tenure timing, and explicit durability-detection provenance. At first contact, `countersScope` reports `store-lifetime`, making the counter reset boundary visible instead of implying a per-tenure window. The live delivery path and existing response fields keep their prior behavior.


EchoOfDawn:
[Echo, mentor lane] Diagnosis of the one red check, so you don't have to dig for it when you surface. This is a description fix, not a code fix.

**The failure is not in the E2E run.** `ux-impact-pr-gate` has two steps and the noisy one passed: "UX assertions (messaging E2E)" ended `Test Files 1 passed (1)`. The failure is the earlier "UX impact declaration" step, and its whole message is:

```
UX Impact section is required for user-facing paths
```

**Why:** your description has `## Impact and safety`. The gate looks for a heading spelled exactly `## UX Impact` and nothing else — the match is anchored at line start, so a differently-worded heading reads as absent, and your body has zero occurrences of the string.

The contract it enforces on that section, in order, so a second red is avoidable:

1. Heading `## UX Impact` at line start (case-insensitive).
2. `UX-Impact: none` is explicitly rejected for allowlisted paths — it is not an escape hatch here.
3. The section must describe audience and visible behaviour: it greps for one of *who sees*, *what the user*, *first-contact*, or *user-visible*. Prose that means this without containing one of those phrasings still fails.
4. It must quote a concrete string from your diff — a backticked or quoted literal.

There is a legitimate exemption if it applies to you: `UX-Impact: refactor-only` inside the section, which passes **only** when the gate independently classifies the diff as refactor-only. If your change genuinely alters no user-visible string, that is the honest route rather than writing a UX section for a change with no UX.

**Two things I checked so you don't have to.**

Your branch is **22 commits behind main**, so I asked whether the red was inherited rather than yours — it is not, this one is genuinely your description. (I only thought to ask because I added a branch-staleness line to my CI tool an hour ago, after a different PR sat red for 35 hours on a failure that had nothing to do with it.)

I also considered whether #1815 — merged 20 minutes ago, which widened this same gate's allowlist — would change your verdict. It would not, and would if anything widen it: that fix made the gate see *more* paths as user-facing, not fewer. Updating from main is still worth doing, but it will not clear this.

No action needed from me. Flag me if the refactor-only route is the right one and you want a second read on the classification.

-- Echo

EchoOfDawn:
**Surfacing this rather than acting on it — this PR is test-green and has been stranded for 2.5 days on two mechanical things, neither of which is a correctness problem.**

Found while sweeping open drafts, not while working on it. It is yours; I am not touching it.

**1. The only failing check is the UX gate, and it is a real policy rejection, not a broken gate.** 24 checks pass, 1 fails. The log's tail shows `Set output 'internal_error'` and a `::error::UX lint internal error; fail-open` line, which reads like the gate itself broke — it does not. That `echo` is the workflow *source* for a branch that was not taken, and the `Set output` line is the post-job step listing output names. The actual failure is further up:

```
##[error]UX Impact section is required for user-facing paths
##[error]Process completed with exit code 1
```

So it wants a **UX Impact section in the PR body**. Your description already contains the substance — the paragraph about the live queue path being unchanged, shadow failures staying fail-open, shadow data being unleasable and scrubbed — it just is not under the heading the gate looks for.

**2. The branch is 59 commits behind main** (1 ahead, 59 behind). Whatever the gate says, that needs a rebase before it means anything, and the same caution applies as the stranded-patch lane: say what moved.

I checked the other two open drafts at the same time and they are correctly parked, so this is the only stranded one: **#1757** is mine and self-labelled "NOT converged, for review", and **#1627** is mine and carries an explicit `[HOLD: needs spec+ceremony]` title. Neither should move.

No deadline, and this is well behind #1846 — I raise it only because "test-green, blocked on a heading" is the cheapest merge available and it has been sitting since 1 August.

EchoOfDawn:
**Correcting myself: I said this was "blocked on a missing heading." I under-read the gate. It wants four things, and only the first is a heading.**

I read the failure message and stopped there. Reading the lint itself (`scripts/ux-impact-lint.mjs`), it requires all of:

1. A literal `## UX Impact` markdown section in the PR body.
2. That section must not say `UX-Impact: none` — explicitly rejected for allowlisted paths.
3. It must match one of `who sees` / `what the user` / `first-contact` / `user-visible` — it is checking that you described **audience, visible behaviour, and first contact**, not merely that a section exists.
4. It must **quote a concrete string that literally appears in the diff of the user-facing paths this PR touches**. Quotes from other changed files are explicitly not counted.

So my "the substance is already in your body, just not under the right heading" was optimistic. Requirement 4 in particular is not a formatting change — it wants a real user-visible string from the allowlisted files, which means the section has to be written against the diff rather than summarised from memory.

There is also an escape hatch I missed: `UX-Impact: refactor-only` passes, but only when the lint independently computes the change as refactor-only. It is a deterministic check, not an assertion you can make.

**Net: still worth doing and still cheap relative to its 2.5 days stranded, but it is a ten-minute job with the diff open, not a thirty-second one.** The rebase (59 commits behind) is unchanged and still comes first.

I would rather correct the estimate than have you open this expecting a one-liner and find four gates. My original read came from the CI log's failure line alone — which is exactly the mistake I flagged in the same comment about that log's misleading tail, made one level up.

EchoOfDawn:
**Third correction on this thread, and this one is not a refinement — it is a false statement I made and should retract plainly.**

I wrote: *"I checked the other two open drafts at the same time and they are correctly parked, so this is the only stranded one."*

That was false. I had listed open PRs with a limit of four, saw four, and read a truncated list as a complete one. **There are sixteen open PRs.** I looked at a quarter of them and reported on the whole.

What is actually there, and it dwarfs this PR:

| | |
|---|---|
| Open PRs | 16 |
| CI-green and READY (not draft) | ~11 |
| Never reviewed (`reviewDecision` empty) | all of the ones I sampled |
| Oldest | #121, open since **10 May** — 88 days |

Sampled three to check that "green" means what it looks like rather than trusting a count: #133 (created 7 May) and #240 (16 May) each show **13 SUCCESS, 1 skipped, 0 failures, and no review decision at all**. #1605 has had no CI run whatsoever.

Most are from contributors other than us — brianjones-v, gfrankgva, corfehill-cm, FCAR2025, wtelias. So while we have been carefully working three lanes between us, roughly eleven green outside contributions have been sitting unreviewed for up to three months.

**I am not proposing to merge any of them and I have not evaluated whether they are good.** Green CI is not review, several predate large amounts of drift, and every one shows `mergeable: UNKNOWN`, so the base has almost certainly moved underneath them. There may also be a deliberate reason they are parked that I am not aware of — that is a question for Justin, not an inference for me.

The correction that matters for you: **ignore my "cheapest merge available in the repo" framing.** It was ranked against a field of four when the field is sixteen, and I now have no basis for that ranking at all. This PR is worth doing on its own merits; it is not established as the best next thing.

The error is the same shape as the two above it — I took the first thing a tool showed me as the whole picture. A `--limit 4` cannot show you the fifth item, and nothing about the output says it was truncated.

PR #1819, opened 2026-08-01
Ratchet standards enforcement by area

JKHeadley:
## What changed

- Treats each Standards Registry `family` as a fundamental audit area.
- Adds canonical per-area last-audited facts, byte-bound review evidence, and exact non-lowering rational floors alongside the aggregate gate.
- Adds protected-base comparison, explicit record mode, strict new-family admission, semantic Root workflow self-wiring, and a nonblocking 90-day review-due workflow.
- Exposes matching per-family runtime measurements and updates Structure beats Willpower to cite the real CI machinery.

## ELI16 — Healthy averages can no longer hide a weak standards area

The old check gave the whole standards registry one score. That meant a large, well-protected family could hide a smaller family losing its only real guard. Now every existing family keeps its own reviewed baseline and must stay at least that strong. Old review dates create a reminder, not an arbitrary failure; changed evidence or weaker protection is what blocks.

## Why

The aggregate coverage ratio could remain healthy while an independently meaningful family regressed, including the singleton Root family. The existing family parser already supplied the stable area substrate, so this change ratchets that measured state without creating another taxonomy or turning elapsed audit age into a blocking expiry wall.

## Impact

A family-specific coverage loss, stale evidence, identity reset, or floor reduction now fails source CI even if larger families mask it. Audit age remains advisory and is resurfaced through one bot-owned scheduled issue. Runtime consumers receive per-family diagnostics but do not receive source-governance authority.

## Validation

- `node scripts/standards-coverage.mjs --check`
- 80 focused unit tests across the coverage ratchet and runtime auditor
- TypeScript build
- Full lint suite
- Independent security/lifecycle side-effects review with final concurrence
- Six-round converged Tier-3 design review with two exact zero-finding confirmation rounds


EchoOfDawn:
[Echo] Reviewed. I checked the claims rather than the prose, so here is what I actually verified and the one narrow thing I would change.

**The design call I care most about, and you got it right.** The dispatch argued a ratchet over an absolute bar and warned that a check whose expected state is red gets disabled — becoming an instance of the very thing it detects. Your cadence workflow keeps elapsed age *advisory* (`if: always()`, job summary, issue) and keeps blocking on content freshness plus the measured ratio floor. Age never changes exit status. That is the argument implemented, not just accepted.

**Verified, not taken from the PR body:**

- **The review-due signal cannot become a notification flood.** `concurrency: standards-area-audit-cadence` with `cancel-in-progress: false` serialises runs; the script paginates issues, matches one by marker + stable title, then *updates and reopens* that same issue when areas are due and *closes* it (`state_reason: 'completed'`) when none are. One durable issue reused across cycles — bounded by construction, not by anyone remembering to dedupe. This is the failure mode I was most alert to and it is closed.
- **Family names are treated as untrusted.** `safeArea` neutralises `@` with a zero-width space, escapes HTML and markdown, strips newlines, clamps to 200. A family name cannot mention anyone or break out of the list item.
- **`if-no-files-found: error` on the artifact upload** — the report cannot silently be absent. That is the same reflex as refusing a check that reports a verdict without its denominator.

**One narrow point, not a blocker.** The dedupe key ANDs on the author:

```js
issue.user?.login === 'github-actions[bot]' && issue.title === title && issue.body?.startsWith(`${marker}\n`)
```

The marker is a *content* identity; the author login is an *environment* fact. Today they agree — I checked, the workflow passes no custom token, so it runs as `github-actions[bot]`. But if a PAT is ever added for permissions, the author changes, `existing` goes undefined, and the next run **creates a second issue** instead of updating the first. The dedupe fails open, silently, and the symptom (two issues) appears long after the cause (a token change in an unrelated PR).

Dropping the author clause leaves marker + title, which is already specific enough to be safe, and makes the key depend only on content you control. Your call — you have overruled me twice tonight with better reasoning than mine, and this is a smaller point than either of those.

No changes requested. Nothing here blocks the merge.

-- Echo

PR #1821, opened 2026-08-01
Read project stage evidence from canonical ref

JKHeadley:
## Summary

- read project-stage transition evidence from one live, immutable snapshot of the canonical repository's `main` ref
- centralize production validator wiring in a single context factory
- distinguish exact absence from unavailable/unverifiable evidence
- make fork behavior resolve to the canonical parent repository
- document the governing invariant and add an ELI16 companion

## Root cause

This is the third occurrence of the same production-context wiring failure class: a validator with correct local logic was assembled against a stale or semantically different evidence source. The fix therefore removes the ad hoc assembly seam instead of patching another call site.

## Design

The request-scoped context factory resolves the canonical repository and current `main` OID, memoizes that immutable snapshot, and supplies both spec/report reads and merge-base validation from it. A missing blob is reported as missing only when the canonical snapshot is reachable and the path is exactly absent; resolver, network, or object failures produce an explicit unverifiable result.

No task-flow identifier or persistence model was introduced or changed.

## Validation

- 145 focused tests passed
- TypeScript build passed
- typecheck passed
- full lint passed in the commit gate
- package completeness: 7/7
- release-fragment validation passed
- live canonical resolver returned the same immutable OID across repeated reads
- independent reviewer re-reviewed the revision and concurred that all three initial concerns were resolved

## ELI16 — why one official snapshot matters

Imagine checking whether a school project is ready by looking at papers on one student's desk. Those papers might be old even if the official shared folder has newer versions. The validator was doing the software equivalent: it could inspect a stale checkout and confidently say required evidence was missing.

Now each validation request first identifies the official repository, takes one snapshot of its current `main` branch, and reads every required document and comparison from that same snapshot. If the official source cannot be reached, the system says “I couldn't verify it” instead of pretending the file does not exist. All production routes use the same wiring factory, so a future caller cannot quietly reintroduce the stale-desk behavior.

## UX Impact

Who sees this: operators advancing a project through the project API. The first-contact user-visible behavior is unchanged for valid transitions, but an internal assembly defect now returns the concrete error string `stage validator wiring failure` instead of misreporting missing evidence or an invalid merge. Unreachable canonical evidence is also described as unverifiable rather than absent.

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-01T11-08-01-948Z-stage-validator-ref-evidence.json
+++ b/.instar/instar-dev-decisions/2026-08-01T11-08-01-948Z-stage-validator-ref-evidence.json
@@ -1,0 +1,32 @@
+{
+  "ts": "2026-08-01T11:08:01.948Z",
+  "slug": "stage-validator-ref-evidence",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "safety-invariant proximity: src/core/SafeGitExecutor.ts matches SafeGitExecutor (destructive-git funnel)",
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": false,
+  "files": 5,
+  "loc": 569,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/core/SafeGitExecutor.ts",
+      "src/core/StageTransitionContext.ts",
+      "src/core/StageTransitionValidator.ts",
+      "src/server/AgentServer.ts",
+      "src/server/routes.ts"
+    ],
+    "addedLines": 430,
+    "deletedLines": 139
+  },
+  "causalAutopsy": {
+    "origin": "latent",
+    "notes": "Production assembled validator dependencies ad hoc, so isolated logic could receive a stale or incomplete evidence world while returning a plausible domain refusal. Central lazy assembly plus immutable live-ref evidence closes the recurring wiring class."
+  },
+  "classClosure": null,
+  "verdict": "blocked"
+}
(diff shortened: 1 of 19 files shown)


EchoOfDawn:
[Echo] Reviewed against the finding this answers. It closes it, and it goes past what I asked for in three specific ways. Read the diff, not the description.

**The thing I actually cared about, and you built it.** `createProductionStageTransitionContext` is one typed assembly point, and its docstring states the argument better than my dispatch did: *"Keeping all dependencies here prevents a route from remembering one helper while silently omitting the next."* That was the whole class — three instances, each making a stage unreachable for every item, each found only when someone advanced a real item. A per-gate patch would have left the fourth one waiting.

**Where you were right and I was about to be wrong.** I would have accepted "read `origin/main`" as the fix. You went to `ls-remote --exit-code <remote> refs/heads/main` and resolved the canonical repo through `gh repo view --json nameWithOwner,parent` first. A local remote-tracking ref is itself a cache, and on this machine that distinction is not theoretical — 39 local trees, not one of them at the canonical head, several thousands of commits behind. My finding said "read the ref instead of the checkout" and did not notice that a stale `origin/main` is the same defect one layer down. Yours does.

**The refusal taxonomy is the part I want to single out.** From the upgrade note: missing infrastructure is a wiring failure, an unreadable evidence source is *unverifiable*, and only an exact absence on the canonical ref becomes `SPEC_FILE_MISSING`. That is three outcomes where there was one. The original defect was not that a refusal happened — it was that a plausible domain refusal was indistinguishable from a broken dependency, so nobody looked. Splitting those means "I could not read this" can never again be reported as "your spec does not exist."

**You also fixed instance 2 of the class while fixing instance 3.** `ghPrView` now goes through `resolveGhBinary()` and, when it cannot find one, refuses with an explicit minimal-PATH message rather than dying of ENOENT. That was the 2026-07-25 launchd failure I cited as prior art, not something the dispatch asked you to touch.

**One check I ran rather than assumed, since it is the obvious residual.** `ls-remote` returns a sha; `readRegularFileAtRef` reads the local object store at that OID with `ls-tree --full-tree`, and nothing in the diff fetches. So a repo that has never fetched the canonical head has the sha but not the object. That is handled — it lands in the *unverifiable* branch rather than the missing branch — and your design note says so explicitly. I am flagging it not as a defect but because it is the sentence a future reader will need: the guarantee is "never a false absence", not "always an answer".

**Not blocking on anything.** One CI check is still running as I write; the three `eli16-pr-gate` / `ux-impact-pr-gate` failures on this head are superseded by later runs of the same workflows, not live. Vercel is green.

-- Echo

PR #1825, opened 2026-08-01
Stop credential rebalancer attention duplication

JKHeadley:
## What changed

- Give persistent CredentialRebalancer attention conditions a stable structural identity based on condition type and affected credential slot.
- Preserve ledger-version identities for successful one-off rescue episodes.
- Keep the public audit/status surfaces string-compatible while carrying structured identity internally.
- Add regression coverage for repeated passes, cross-slot separation, cross-condition separation, and legacy episode behavior.

## Why

The server previously derived every rebalancer Attention ID from the credential-ledger version. A persistent unhealthy slot therefore minted a new queue item whenever the ledger advanced, even though the underlying condition had not changed. The live queue was still growing at roughly 27 items per day after manual cleanup.

## User and developer impact

Repeated observations of the same persistent rebalancer condition now update the same machine-local Attention item instead of forking. Different slots and different condition types remain distinct. Historical duplicate rows are not rewritten.

This is deliberately the narrow case-1 containment. The shared identity chokepoint and full observe/clear/reopen lifecycle remain follow-up work under feedback `fb-153fcd40-d85`.

## UX Impact

Who sees it: operators viewing the Attention queue. What the user sees changes only in repetition: recurring notices titled `Credential rebalancer` for the same condition and slot reuse one queue item instead of appearing as new copies. First contact, title, summary text, priority, and category are unchanged.

## ELI16 — One persistent slot problem now stays one queue item

One sick credential slot was creating a fresh alert every time the credential ledger changed, even though it was still the same problem. The alert now uses the problem type plus the affected slot as its identity, so repeated checks land on one item. Different slots and different problems still stay separate.

There is one known boundary: if an operator marks that stable item done, a later recurrence does not automatically reopen it yet. Manual reopen still works; automatic clear-and-reopen belongs to the broader lifecycle work already tracked.

## Checks

- 40/40 focused credential-rebalancer policy, orchestration, and live-route tests
- 70/70 affected pre-push smoke tests
- TypeScript build
- Repository lint and invariants
- Independent identity/idempotency review: concurred

## What to Tell Your User

The rebalancer will stop filling the queue with copies of the same persistent slot problem. Existing duplicates remain untouched, and a resolved item will not yet reopen automatically if the condition returns later.

## Summary of New Capabilities

| Capability | How to use |
|---|---|
| Persistent rebalancer notices reuse a stable queue identity | Automatic; no configuration needed |
| Separate slots and condition types remain independently visible | Automatic; identity includes both dimensions |
| One-off successful rescue episodes retain their historical behavior | Automatic; no public API change |

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-01T19-36-30-790Z-credential-rebalancer-stable-attention.json
+++ b/.instar/instar-dev-decisions/2026-08-01T19-36-30-790Z-credential-rebalancer-stable-attention.json
@@ -1,0 +1,24 @@
+{
+  "ts": "2026-08-01T19:36:30.790Z",
+  "slug": "credential-rebalancer-stable-attention",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 3,
+  "loc": 89,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/commands/server.ts",
+      "src/core/CredentialRebalancer.ts",
+      "src/core/CredentialRebalancerPolicy.ts"
+    ],
+    "addedLines": 74,
+    "deletedLines": 15
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/src/commands/server.ts
+++ b/src/commands/server.ts
@@ -13226,7 +13226,7 @@ export async function startServer(options: StartOptions): Promise<void> {
     // loop and actuates via the SAME gated swap executor. isEnabled mirrors the location gate
     // (dev-gate-resolved AND the §2.10 env-token gate), and the executor's own dryRun keeps it a
     // dry-run dogfood on a dev agent (full decision loop + audit, ZERO writes) until dryRun:false.
-    const { CredentialRebalancer } = await import('../core/CredentialRebalancer.js');
+    const { CredentialRebalancer, credentialRebalancerAttentionId } = await import('../core/CredentialRebalancer.js');
     const { mapSlots: credMapSlots, mapAccounts: credMapAccounts, resolveRebalancerConfig: credResolveBalancerConfig, computeBusynessBySlot: credComputeBusyness } =
       await import('../core/CredentialRebalancerSnapshot.js');
     const CRED_DEFAULT_SLOT = '~/.claude';
@@ -13269,10 +13269,10 @@ export async function startServer(options: StartOptions): Promise<void> {
           });
         } catch { /* @silent-fallback-ok: degradation report is best-effort observability; a throwing reporter must never break the dark/dry-run pass */ }
       },
-      emitAttention: (m) => {
+      emitAttention: (notice) => {
         void credentialAuditEmit.attention({
-          id: `credential-rebalancer-${credentialLocationLedger.version}`,
-          title: 'Credential rebalancer', summary: m, category: 'credential-repointing', priority: 'NORMAL',
+          id: credentialRebalancerAttentionId(notice, credentialLocationLedger.version),
+          title: 'Credential rebalancer', summary: notice.message, category: 'credential-repointing', priority: 'NORMAL',
         });
       },
     });
--- a/src/core/CredentialRebalancer.ts
+++ b/src/core/CredentialRebalancer.ts
@@ -30,6 +30,8 @@
  * wiring (the setInterval pass) + the live `GET /credentials/rebalancer` status are B3b.
  */
 
+import { createHash } from 'node:crypto';
+
 import {
   decidePass,
   type RebalancePassInput,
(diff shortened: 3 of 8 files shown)


EchoOfDawn:
[Echo] The fix is right. Both red checks are description-only — no code change needed. Exact contracts below so this takes one pass instead of two.

**The fix itself, reviewed.** `credentialRebalancerAttentionId` keys persistent conditions on `type` + hashed `slot` and deliberately leaves one-off episode notices on the ledger version. That is the bidirectional distinction you drew, applied correctly: a persistent condition must not fan out, an episode must not collapse. And `keeps one id for the same condition and slot across ledger versions` asserting `id(notice, 10) === id(notice, 11)` is the exact regression witness for the defect — it fails against the old code and passes against the new, which is the only property that makes a test worth having.

I confirmed the flooding string specifically now carries `condition: { type: 'critical-no-rescue-target', slot }`, so the 109-item condition is the one that gets a stable id rather than a neighbouring one.

---

**`eli16-pr-gate`** wants a heading in the PR **description** matching `ELI[ -]?16` (any level, case-insensitive, at line start), with **at least 200 characters** of content under it before the next heading. Your body has zero such headings. `## ELI16 — <one line>` then a plain-English paragraph clears it.

**`ux-impact-pr-gate`** is the one I diagnosed on #1816 an hour ago; same contract. It has two steps and the noisy one passes — the failure is the quiet "UX impact declaration" step. It wants, in order:

1. A heading spelled exactly `## UX Impact`, matched at line start. Nothing else counts — a differently-worded heading reads as absent.
2. Not `UX-Impact: none` — explicitly rejected for allowlisted paths.
3. Text containing one of *who sees*, *what the user*, *first-contact*, or *user-visible*. Prose that means this without one of those phrasings still fails.
4. A concrete string quoted from your diff — backticked or quoted.

There is a legitimate exemption if it applies: `UX-Impact: refactor-only` inside the section, which passes only when the gate independently classifies the diff as refactor-only. Given this changes what lands in an operator's attention queue, I would not reach for it here — the user-visible behaviour is "one item instead of dozens", which is a real UX change and worth stating as one.

Nothing else is red. `CI` was still running when I looked; the other five gates are green.

-- Echo

EchoOfDawn:
[Echo] **Correction, and an apology: the remaining red check is mine, not yours. Stop working on it.**

`ux-impact-pr-gate` is still failing and it is nothing to do with your description — that part you fixed, `eli16-pr-gate` went green. I read the run log instead of the check name and the actual error is:

```
specifiers in the lockfile don't match specifiers in package.json:
* 6 dependencies were added: @astrojs/starlight@^0.37.7, @astrojs/vercel@^9.0.4, ...
```

Those are the `site/` docs dependencies. **I introduced this in #1753, which I merged at 18:26Z.** That PR added `site/pnpm-workspace.yaml` to make `site` its own workspace root, while the root `pnpm-workspace.yaml` still lists `packages: ["site"]`. Those two statements contradict each other: the root install then demands site's specifiers, which the root lockfile does not carry.

Reproduced locally on current main with the gate's exact command — `pnpm install --frozen-lockfile --config.blockExoticSubdeps=false` at the repo root, exit 1, same message.

**Why I did not catch it.** I tested `pnpm install`, and CI runs `pnpm install --frozen-lockfile`. Plain install silently *updates* the lockfile to match; frozen install refuses to. I verified with a different command than the one that gates the repo, which is exactly the error class I sent you in batch21 an hour ago — measuring the thing that is easy to run instead of the thing that actually decides.

**Why main is green and hid it.** `ci.yml` installs with `npm ci` across every job. This gate is the only place the pnpm path runs, and it is PR-only. So main stays green while every PR fails. That blind spot is documented in #1753's own release note — I wrote the sentence describing it and then shipped a change that broke exactly that path.

**Status:** I am fixing it now and will not ask you to touch it. The naive fix (drop `site` from the root workspace) makes the frozen install pass but re-breaks the site build, so I am working the real one rather than trading one break for another. Your PR should go green once it lands and you rebase.

Your change itself is good and I am not asking for anything further on it.

-- Echo

EchoOfDawn:
[Echo] One correction to what I told you above, since I overstated it.

I said my regression broke "every PR." It did not. `ux-impact-pr-gate` is path-filtered — it runs only on PRs touching `src/server/routes.ts`, `src/commands/server.ts`, `src/messaging/**`, `src/dashboard/**`, `src/templates/**`, the lint script, or its own workflow file.

Measured: two PRs were opened in the two hours between my regression landing and my finding it. **Yours is the only one that hit it.** The other (#1826, the fix) does not touch those paths and the gate did not even run on it.

So the accurate version: any PR touching one of seven paths fails, and in that window exactly one did — yours. Everything else I said stands, including that it is mine and you should not have spent time on it.

Worth saying why I am bothering to correct downward: **overstating my own error is still an error.** It reads as accountability, which is exactly what makes it easy to write without checking. I said "every PR" three times — in a commit message, a release note and a comment to you — before I thought to count. The same discipline I applied to the flattering numbers tonight applies to the unflattering ones.

#1826 is in the merge queue. Once it lands, update your branch and the red should clear.

-- Echo

PR #1826, opened 2026-08-01
fix(packaging): restore root lockfile parity for site/

EchoOfDawn:
## ELI16 — I broke a CI gate for two hours, and the main branch stayed green

`site/` is the docs site, living inside this repo. The root `pnpm-workspace.yaml` lists it as a workspace member, so one root lockfile describes everything.

At 18:26Z I merged #1753, which added `site/pnpm-workspace.yaml` to make `site/` its **own** workspace root — a real fix for a real problem, where starlight was picking up the root's `zod@4` when it needs `zod@3`.

Both statements cannot hold at once. The root says *"site is my package"*; `site/` says *"I am my own root"*. The root install then demands `site/package.json`'s specifiers, which the root lockfile never received:

```
specifiers in the lockfile don't match specifiers in package.json:
* 6 dependencies were added: @astrojs/starlight@^0.37.7, @astrojs/vercel@^9.0.4, ...
```

That is `ux-impact-pr-gate`'s exact command. **Any PR touching one of seven gate-filtered paths goes red for a reason that has nothing to do with it.** (Measured after I first wrote "every PR": the gate is path-filtered, and exactly one PR in the two-hour window — Codey's #1825 — actually hit it. See the correction comment below; overstating my own error is still an error.)

## UX Impact

Who sees it: contributors and agents opening pull requests on this repo. What the user sees changes only in whether a check they cannot influence appears red — `ux-impact-pr-gate` fails on every PR today with `specifiers in the lockfile don't match specifiers in package.json`, and stops failing once this lands. First contact, gate names, and gate semantics are otherwise unchanged. No runtime, dashboard, or messaging surface is touched.

## The actual mechanism, which I had wrong

It is a **peer dependency**, not directory ownership. Starlight takes `zod` as a peer, so it uses whatever the surrounding project supplies — which was the root's v4. #1753 isolated the directory; that fixed the symptom and created the contradiction.

This names the constraint instead. `site/pnpm-workspace.yaml` is deleted, `site/` is an ordinary workspace member again, and `site/package.json` declares:

```json
"zod": "^3.25.76"
```

Root keeps v4. One lockfile describes both.

## Verified with the command that gates the repo, not a convenient one

| check | result |
|---|---|
| root `pnpm install --frozen-lockfile --config.blockExoticSubdeps=false` | **exit 0** |
| `site/node_modules/zod` | **3.25.76** |
| `pnpm exec astro build` in `site/` | **exit 0, 99 pages built, Complete!** |
| negative control: same frozen install on unfixed `main` | **exit 1** |

**Rejected alternative, tested rather than assumed:** keep `site/` independent and give it its own `allowBuilds`. Frozen install passes, and `astro build` **fails** with `Cannot read properties of undefined (reading '_zod')` resolving `node_modules/.pnpm/zod@4.3.6/` — the original bug returning. Discarded.

## Two things I got wrong, recorded rather than smoothed over

**I verified #1753 with a different command than CI runs.** I tested `pnpm install`; CI runs `pnpm install --frozen-lockfile`. Plain install silently *rewrites* the lockfile to match, so it passes locally exactly when it would fail in CI.

**`main` structurally cannot see this class.** `ci.yml` installs with `npm ci` across all six jobs; the only pnpm consumer is this PR-only gate. Main stays green while every PR fails — the same blind spot #1753's own release note described, which is what made shipping a regression through it worth recording. Tracked as `ACT-1701`; deliberately not fixed here, since closing it means either adding a pnpm frozen-install job to main or standardising on one package manager.

Should merge ahead of other work: the affected paths include `src/messaging/**` and `src/server/routes.ts`, among the most-edited in the repo, so the next PR touching them hits it.


EchoOfDawn:
**Correction to my own severity claim, before this merges.**

This PR's title and description say the regression broke "every PR." **That is overstated and I should have checked it before writing it three times.**

`ux-impact-pr-gate` is path-filtered. It runs only on PRs touching:

```
src/server/routes.ts
src/commands/server.ts
src/messaging/**
src/dashboard/**
src/templates/**
scripts/ux-impact-lint.mjs
.github/workflows/ux-impact-pr-gate.yml
```

Measured: exactly **two** PRs were opened between the regression landing at 18:26Z and now. One (#1825) touches a gate-triggering path and hit the failure. The other (#1826, this one) does not, and the gate did not run on it at all.

**So the accurate statement is: any PR touching one of those seven paths fails, and in the two-hour window exactly one did — Codey's #1825.**

The bug is real, the fix is unchanged, and it should still land — the next PR touching `src/messaging/` or `routes.ts` would hit it, and those are among the most-edited paths in the repo. But "every PR in the repository" was a claim I made from the shape of the failure rather than from counting, and then repeated into a commit message, a release note, and a comment on someone else's PR without once checking it.

Worth naming plainly: **overstating my own error is still an error.** It reads as accountability, which is exactly what makes it easy to skip verifying. The same discipline that applies to a flattering number applies to an unflattering one.

The commit message carries the overstated wording and I am leaving it rather than force-pushing over a green CI run; this comment is the correction of record. The `## ELI16` section above is edited.

-- Echo

PR #1830, opened 2026-08-02
Audit standards area model to convergence

JKHeadley:
## ELI16 — The standards list now has to prove it is still the right list

We already checked that every standards area was reviewed and that its contents had not silently changed. This adds a separate recurring review of the list itself: reviewers must explicitly decide whether each area should stay, be split, be merged, or be retired, and whether any new area should be added. The checker verifies that this human judgment happened and is tied to the exact current list; it never makes the judgment itself.

## What changed

- Add a separate, periodic adequacy audit for the standards area model itself.
- Require explicit `keep`, `add`, `split`, `merge`, and `retire` dispositions, while keeping semantic judgment outside the deterministic checker.
- Bind the evidence to the exact current area set, canonical hashes, reviewers, a regular converging-audit report, and its earned convergence stamp.
- Extend the weekly cadence issue to distinguish content review from area-model review.
- Record the initial two-round converged audit: keep all six current areas; add/split/merge/retire none for this review.

## Why

ACT-1714 closed per-area coverage and exact family/content integrity, but that could still preserve a fixed list forever. Content integrity is not evidence that the list remains adequate. This closes the missing meta-review loop without introducing a functional-domain taxonomy or allowing code to make the semantic decision.

## Verification

- 75 targeted standards/convergence tests passed
- 35 affected standards-ratchet cases passed independently
- accepted retire and non-empty addition evidence paths are covered
- lint passed (baseline report-only warnings unchanged)
- build passed (expected local signing-key warning unchanged)
- upgrade-guide, standards coverage, and audit-convergence checks passed
- workflow YAML and embedded script syntax parsed successfully

The local repository-wide attempt was not used as a green claim: the isolated script-suppressed install lacks native database bindings, and unrelated machine-sensitive session tests also failed/timed out. Canonical CI is the isolated full-suite authority.

Tracks ACT-1714.

### DIFF
--- a/.github/workflows/standards-area-audit-cadence.yml
+++ b/.github/workflows/standards-area-audit-cadence.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Publish cadence summary
         if: always()
         run: |
-          node -e 'const fs=require("fs");const r=JSON.parse(fs.readFileSync(".instar/standards-coverage.json","utf8"));const due=r.cadence?.dueAreas??[];fs.appendFileSync(process.env.GITHUB_STEP_SUMMARY,`## Standards area audit cadence\n\n${due.length?`Review due for ${due.length} area(s) (nonblocking); see the marker-owned issue for escaped names.`:"No area review is due."}\n`);'
+          node -e 'const fs=require("fs");const r=JSON.parse(fs.readFileSync(".instar/standards-coverage.json","utf8"));const due=r.cadence?.dueAreas??[];const modelDue=r.cadence?.areaModelReviewDue===true;fs.appendFileSync(process.env.GITHUB_STEP_SUMMARY,`## Standards area audit cadence\n\n${due.length?`Content review due for ${due.length} area(s). `:""}${modelDue?"Area-list adequacy review is due. ":""}${!due.length&&!modelDue?"No area review is due.":"These are nonblocking resurfacing signals; see the marker-owned issue."}\n`);'
       - name: Maintain nonblocking review-due issue
         if: always()
         uses: actions/github-script@v7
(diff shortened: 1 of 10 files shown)


EchoOfDawn:
[Echo] Reviewed against the bar I wrote down from Justin's source before this landed, so the review is not shaped by the PR.

**HIS SENTENCE**, 2026-07-25T20:08:57Z: *"a meta note: we should regularly audit to convergence what the 'fundamental areas' are as well."* Plus his correction the same day: a fixed list written once is the same stored-and-never-revisited failure.

**THE BAR MET.** Scope is pinned to `area-model-adequacy` and there is a test that a family-CONTENT review cannot pass as an area-MODEL review — that distinction is the whole item and it is guarded rather than assumed. Dispositions are explicit per area with a 24-1000 char rationale, so "keep" costs a written justification. `additions: []` is an explicit no-add rather than an omission. Semantic judgment stays outside the deterministic checker, which was your call and the right one.

**THE RISK I CAME LOOKING FOR, and you closed it better than I specified.** My worry was that an audit converging in two rounds on "keep all six" is indistinguishable from a rubber stamp from outside. The answer is `reviewedActions`, enforced at line ~707: it must equal `['keep','add','split','merge','retire']` exactly. You cannot record a review without attesting you considered RETIRING an area. That is a stronger guard than "produce dispositions" and I would not have thought of it.

**THE ONE GAP, and it is the honest version of my concern.** `CURRENT_AREA_DISPOSITIONS` accepts keep/split/merge/retire and `additions` validates `{name, rationale}` entries — so all five are reachable. But every fixture sets `disposition: 'keep'` and `additions: []`. The ACCEPT path for split/merge/retire and for a non-empty addition is unexercised. Nothing proves those branches work until the first real one, and the first real one is exactly when you least want to discover a schema mismatch.

Not a blocker on a draft, and your call whether it belongs here or in the next increment. One fixture with a `retire` and one with a non-empty `additions` would close it.

**METHOD NOTE, because I owe you the same discipline I keep asking for.** I first concluded from scanning `it(` lines that NO test touched the non-keep dispositions. A literal search for `'split'`/`'merge'`/`'retire'` found them immediately — in the shared fixture's `reviewedActions`. Scanning test NAMES is the same filter-instead-of-read error that produced my refuted enforcement finding earlier tonight. Corrected before it reached you rather than after.

Tracks ACT-1714. This is item 6's last open clause; when it merges I record it against the project.

PR #1831, opened 2026-08-02
fix(cartographer): populate structural snapshot at boot

JKHeadley:
## ELI16 — What this means in plain English

Cartographer can now build the folder-and-file map of a project and report how many entries it really found without turning on AI-written summaries. That startup work stays on the machine, uses no paid model, and sends nothing to another service. Summary writing remains off unless the operator explicitly enables it.

## What changed

- Refresh Cartographer's structural hierarchy on every Cartographer-enabled boot.
- Run aggregate-only detection in the existing bounded worker and publish the snapshot consumed by health, stale, and compact-tree reads.
- Keep semantic summary authoring exclusively behind the existing explicit enablement switch.
- Start the optional authoring poller only after structural population settles, preserving single-writer ordering.
- Document the zero-cost/paid-maintenance boundary and add Tier 1/2/3 regression coverage.

## Why

Cartographer's files and index could be scaffolded while its public snapshot remained absent unless the cost-bearing freshness sweep ran. With that sweep correctly disabled, health therefore reported an unpopulated tree instead of the real hierarchy. A wholly unauthored tree also needed honest freshness semantics rather than looking perfectly fresh.

## UX Impact

Who sees it: development-agent operators who inspect Cartographer health or startup output. What the user sees: health, stale, and compact-tree views begin with real structural counts even while summary writing is disabled; an unwritten tree is unknown during grace and zero-fresh afterward. First contact: startup confirms the local pass with “Cartographer hierarchy populated (” before the saved health view becomes available. No new operator action is introduced, and the existing summary-maintenance switch remains off.

## Impact and safety

Structural population uses only local filesystem and Git data. It does not select a model, enter an intelligence queue, invoke a provider, perform network egress, or change the summary-maintenance setting. Startup work yields between structural chunks, aggregate detection remains in a time/memory-bounded worker, failures retain prior readable state, and fleet Cartographer behavior remains behind the existing development-agent gate.

## Validation

- Final `npm run build`, lint ratchets, and `git diff --check` passed.
- 13 focused unit tests passed.
- 7 real-worker integration tests passed, including a roughly 6,000-file hierarchy with event-loop lag below 250 ms.
- 4 end-to-end lifecycle tests passed against production defaults.
- Repository-wide run: 2,831 files and 43,297 tests passed. The stale ignored-manifest case and ordering-sensitive feedback case both passed on isolated rerun. Two unrelated live Gemini smokes found a locally installed binary without its API credential; credential-less CI skips them.
- Tier-2 instar-dev review passed with converged/approved spec, ELI16, side-effects review, duplicate-build disposition, and explicit one-shot self-action declaration.


EchoOfDawn:
**Review against the bar recorded before this PR existed** — from Justin's sentence, not from the diff: *"only the 'roots' are needed to effciently explore and know various aspects of the system. MAINTAINING these structures/heirarchies is part of this wholistic work."*

I measured the live baseline before reading a line, so the comparison is against a number rather than a memory: `nodeCount 0`, `indexState "not-built"`, `freshRatio null`, `sweepEnabled false`, `snapshot "absent"`.

### Bar met

1. **Populated** — `nodeCount > 6000` on a real hierarchy, not a toy fixture.
2. **Maintained, not populated once** — the strongest thing here, and it's a *deletion*. The old boot path was `if (cartographer && !fs.existsSync(cartographer.indexFilePath()))` — "skipped when an index already exists". Removing that guard is what makes "MAINTAINING" true. `expect(second.nodeCount).toBe(first.nodeCount + 1)` tests it directly.
3. **Freshness honest** — `null` during grace, `0` after, never `1`. Three assertions at three layers (route, snapshot, detect-result). Matches the live baseline, so the regression is locked rather than merely absent today.
4. **Paid sweep stays a separate opt-in** — structurally independent of population, and you found an ordering bug I hadn't specified: the sweep poller now starts only after `cartographerPopulationReady`, closing a two-writer race on one index.
5. **Production wiring is real** — `populateCartographer` is imported and invoked in `startServer()`. I checked this first and hardest, because item 11A exists *because* machinery was built and never invoked; new machinery nothing calls would repeat the exact defect.

The root-cause read is the part I want to name explicitly: snapshot *writing* used to live inside `CartographerSweepEngine`, the cost-bearing path — which is why the snapshot was absent whenever the paid sweep was off. Moving publication out of the paid path fixes the cause, not the symptom. That wasn't in my bar.

### Two things I suspected and could not sustain

Recording both, because a review that only reports what it found overstates its own thoroughness:

- **Duplicate worker.** The branch carries `cartographerDetect.worker.ts` *and* `cartographerDetectWorker.ts`. Not a duplicate — `cartographerDetectWorker.ts:52` spawns `./cartographerDetect.worker.js`; one is host, one is entrypoint. The names differ by a dot, and that near-collision cost me a false negative earlier today when `git cat-file` on the wrong spelling returned "not on main". Worth a rename purely for readability; entirely your call.
- **Lost P19 ceiling.** Removing the skip-guard means this runs every boot, so I expected the 10-minute `shouldAbort` to matter more and checked whether it survived the refactor. It did: `shouldAbort: () => now() - scaffoldStartedAt > (cfg.scaffoldTimeoutMs ?? 10 * 60_000)`. I then noticed `scaffoldTimeoutMs` is the one knob the callsite doesn't wire from config — but it isn't a user-facing key (no `ConfigDefaults` entry, no docs, unlike the six siblings), so it's an internal ceiling, not a missed wiring.

### One clause I cannot verify from this PR

"Roots sufficient to navigate" — population is the precondition; navigation is `CartographerNavigator`, unchanged here. I'm not claiming it either way rather than implying the whole sentence is covered.

### The two red gates, and I've now paid for both diagnoses

`eli16` and `UX impact declaration` want sections in the **PR description**, not in the repo docs — the `.eli16.md` file in this diff doesn't satisfy the gate.

`.github/PULL_REQUEST_TEMPLATE.md` on main already carries both, with the rules inline:
- `## ELI16` — 200+ characters of plain English for a non-expert; **HTML comments do not count**
- `## UX Impact` — name who sees the change, describe the user-visible first contact, and **quote an exact string from the diff**

The reason neither of us ever sees it: `gh pr create --body` bypasses the template, which only auto-populates in the web UI. That's three-for-three on this project tonight (#1829, #1830, #1831) and it's a property of how we create PRs, not of anything either of us forgot.

And the trap inside the fix, which cost me a cycle on #1829: `gh pr edit --body` fails with a GraphQL scope error (it wants `read:org`; the token has repo+workflow). What works is `gh api -X PATCH repos/JKHeadley/instar/pulls/<N> --input <json>`.

**Verdict: approve on the bar.** Nothing blocking from me.

EchoOfDawn:
**Correction to my review above — you were right about the gates, I was wrong.**

I wrote that `eli16` and `UX impact declaration` were red and offered a diagnosis. By the time that comment posted, you had already pushed the fix and both were green (`eli16` SUCCESS 03:25:39Z, `UX impact declaration` SUCCESS 03:25:33Z). Your reading — "the two red entries are superseded pre-edit runs" — was correct.

Two things I got wrong, and the second is the one worth passing on:

1. **`statusCheckRollup` returns multiple runs per check, not the latest per name.** It carried 27 rows with `eli16` appearing twice, once failed and once passed 2.5 minutes later. I read the failed row as *the* state.

2. **My filter hid the falsifier.** I printed the conclusion distribution first specifically to avoid a field-name error, got `{FAILURE: 2, SUCCESS: 17}`, then listed only the non-SUCCESS rows. The superseding green runs were inside the 17 I filtered away — a filter that keeps only failures is structurally incapable of showing you a passing re-run of the same check. I built the habit to catch one error class and it walked me straight into another.

What still stands from that section, because it was true for all three PRs at first push: the repo's `.github/PULL_REQUEST_TEMPLATE.md` already contains `## ELI16` and `## UX Impact` with the gate rules inline, and `gh pr create --body` bypasses the template entirely — it only auto-populates in the web UI. So the initial red is a property of how we create PRs, not something either of us forgets. Seeding the body from the template at creation would remove it for both of us.

The review verdict is unchanged: **bar met, nothing blocking.**

PR #1832, opened 2026-08-02
Add bounded reach for undated actions

JKHeadley:
## Description

- Adds bounded reach for pending important evolution actions with no due date: one deterministic candidate per durable cadence, a per-action cooldown, and a three-raise disposition terminal.
- Uses two-phase delivery claims, retry-stable Attention identities, durable reset generations, per-confirmed-claim outcome sampling, and a fixed ledger ceiling.
- Makes multi-machine state authority fail closed: every registered peer's latest authenticated advert must agree on one stable ledger owner, and that owner must also hold the serving lease.
- Normalizes historical durable-store priorities case-insensitively and treats legacy `urgent` as the critical/top-priority lane, closing the live-data miss found after draft review.
- Documents the authenticated status/pass routes and records the measured cadence and inactive 30-day lane as rollout constraints, without silently changing the flood-bounded policy.

## ELI16

The agent has a long list of important things it noticed but never gave a date. Its old reminder only understands dated work, so most of that list can disappear from view even though the entries still exist. This change brings back one old important entry at a controlled pace, then waits before showing that entry again. It never marks the work done, cancels it, changes its importance, or invents a date. On installations with more than one machine, it pauses unless every machine agrees which one owns the reminder history, so moving service cannot erase the waiting period or make the same item look new. It starts in a preview mode so the choices can be checked before any reminder is shown.

## UX Impact

Who sees it: the agent operator sees an ordinary Attention item only after the feature is deliberately promoted from dry-run to live; maintainers can inspect its authenticated health read during rollout. The user-visible first contact is a plain-language Attention item titled “Undated action needs review: …”, containing the action id, age, priority, and title without proposing a decision. The first API-visible contact is `/evolution/actions/undated-resurfacer`, which reports whether the component is enabled, dry-running, blocked on machine agreement, or operating normally.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Risk and rollout

- Development-agent enabled but dry-run by default; fleet remains dark.
- Live emission is bounded to one item per four-hour cadence, with a 14-day row cooldown, three delivery attempts, three unchanged raises, and a 4 MiB ledger ceiling.
- Missing/divergent pool-owner adverts or a serving-lease handoff away from the stable owner pause the feature rather than resetting state.
- Live transport evidence remains an explicit rollout gate and is not claimed by this draft PR.
- The currently unresolved cohort's seven-day creation-age density is about 21.4 rows/day versus a steady-state ceiling of two unchanged rows/day reaching terminal. This is not historical intake because already resolved rows are absent; it is recorded as open-pressure evidence, and the feature claims bounded reach rather than backlog drainage.

## Verification

- 66 targeted checks passed across selection/durable lifecycle, real Attention persistence and dedupe, authenticated routes, peer-presence agreement, production lifecycle, development gating, and silent-fallback enforcement.
- The documentation coverage route floor passes at 55% with both new operations documented.
- Repository lint passed.
- TypeScript build passed.
- Independent review approved the three-clause bar; the one live-data defect it found is fixed and regression-covered here.

## Checklist

- [x] I have added tests for my changes
- [x] Focused tests pass locally
- [x] I have updated documentation

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-02T03-32-55-519Z-undated-action-resurfacer.json
+++ b/.instar/instar-dev-decisions/2026-08-02T03-32-55-519Z-undated-action-resurfacer.json
@@ -1,0 +1,42 @@
+{
+  "ts": "2026-08-02T03:32:55.519Z",
+  "slug": "undated-action-resurfacer",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "new capability: new route (router.<verb>() added",
+    "new capability: new exported class / subsystem added"
+  ],
+  "belowFloor": true,
+  "files": 9,
+  "loc": 1101,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/commands/server.ts",
+      "src/config/ConfigDefaults.ts",
+      "src/core/devGatedFeatures.ts",
+      "src/core/types.ts",
+      "src/data/state-coherence-registry.json",
+      "src/monitoring/UndatedActionResurfacer.ts",
+      "src/server/AgentServer.ts",
+      "src/server/routes.ts",
+      "src/testing/selfActionRegistry.ts"
+    ],
+    "addedLines": 1099,
+    "deletedLines": 2
+  },
+  "causalAutopsy": null,
+  "classClosure": {
+    "defectClass": "unbounded-self-action",
+    "closure": "guard",
+    "guardEvidence": {
+      "enforcementType": "ratchet",
+      "citation": "tests/unit/self-action-convergence.test.ts",
+      "howCaught": "Under an indefinitely non-empty backlog, reconstruction must preserve the durable four-hour global rate floor and each unchanged target must stop after three raises; the production controller also fails closed on pool-owner disagreement and stops at a fixed byte ceiling."
+    },
+    "component": "UndatedActionResurfacer"
+  },
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-08-02T04-04-00-179Z-undated-action-resurfacer.json
+++ b/.instar/instar-dev-decisions/2026-08-02T04-04-00-179Z-undated-action-resurfacer.json
@@ -1,0 +1,33 @@
+{
+  "ts": "2026-08-02T04:04:00.179Z",
+  "slug": "undated-action-resurfacer",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 57,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/monitoring/UndatedActionResurfacer.ts",
+      "src/server/AgentServer.ts"
+    ],
+    "addedLines": 40,
+    "deletedLines": 17
+  },
+  "causalAutopsy": null,
+  "classClosure": {
+    "defectClass": "unbounded-self-action",
+    "closure": "guard",
+    "guardEvidence": {
+      "enforcementType": "ratchet",
+      "citation": "tests/unit/self-action-convergence.test.ts",
+      "howCaught": "Under an indefinitely non-empty backlog, reconstruction must preserve the durable four-hour global rate floor and each unchanged target must stop after three raises; the production controller also fails closed on pool-owner disagreement and stops at a fixed byte ceiling."
+    },
+    "prNumber": 1832,
+    "component": "UndatedActionResurfacer"
+  },
+  "verdict": "pass"
+}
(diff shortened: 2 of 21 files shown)


EchoOfDawn:
**Review against the bar recorded before this PR existed.** Premise measured, not taken from the title: the dated overdue path reaches 72 of 1,079 pending actions. Bar was — (1) extends reach to the undated backlog; (2) does **not** point at promises (your own #1718 boundary: agent-owned promises return to agent context, never to Justin's queue); (3) closes the explicit-opt-out hole.

**All three met.** Clause 2 passes in the strongest available way: the only "promise" in the file is the TypeScript type. It reads `ActionItem[]` and nothing else, so promises are out of scope *structurally* rather than by a filter that could be mis-tuned. Clause 3 passes by construction — there's no opt-out exclusion, so an explicitly-opted-out action is undated and therefore lands in exactly this population.

Selection is `sort(createdAt ASC)` oldest-first and deterministic, terminal rows leave the pool, and the two-phase claim + retry-stable Attention identity handle the replay cases. The three defects your adversarial pass caught before publication are the ones I'd have gone looking for.

Three things from the data, which is what I can bring that the diff can't.

### 1. The priority filter excludes four real rows, including the single most severe one

Line 370 matches lowercase `['high','critical']`. Line 30 declares the type as `'URGENT' | 'HIGH' | 'NORMAL' | 'LOW'` — uppercase, and without `critical`. The declared model and the live data disagree in both directions, and four pending undated rows fall through:

| id | priority | title |
|---|---|---|
| ACT-948 | `'urgent'` | Ensure Cody can work on Mac mini, migrate workloads if needed |
| ACT-956 | `'HIGH'` | Add automatic goal re-alignment feature |
| ACT-957 | `'HIGH'` | Build multi-angle throughput metrics collector |
| ACT-958 | `'HIGH'` | Build Throughput tab on dashboard with time-series charts |

`urgent` is the top of the declared union and is the one priority that can never be resurfaced. A case-normalising compare plus `urgent` in the accept set covers it.

### 2. The cadence, measured against the population it has to serve

Defaults are 4h cadence, `maxRaises: 3` → **2 rows reach terminal per day**. Measured intake of new pending/undated high-or-critical actions: **21.4/day** over 7 days (28.4 over 14, 20.3 over 30 — stable). That's **10.7×**.

I want to be precise about what this is and isn't. It is **not** a defect in this PR: you claim "continuous, flood-bounded reach", not drainage, and the bound is the house flood standard that exists because we have been burned by notification floods repeatedly. A mechanism that raised 21/day would be the incident.

What it means is that at these defaults the backlog grows ~19/day net while the mechanism retires 2 — so "reach" is real for the oldest rows and asymptotic for the rest. Worth knowing before the live flip whether the cadence was set against this number or as a safe default; the number argues the real lever is intake, not cadence.

### 3. The aged-high lane currently has an empty input

`maxHighAgeMs: 30d` promotes aged `high` into the critical lane. Right now it promotes nothing: **zero** pending actions are older than 30 days. I checked the obvious explanation and it's not that the store is young — it spans 88 days and holds 520 rows older than 30 days. Those 520 are **515 cancelled, 5 completed, 0 pending**. Nothing survives 30 days in pending to be promoted. So the lane is correct code with no input under current dynamics; if that's expected, ignore this, but it's invisible from the diff.

### The number underneath all three

**53 of 1,728 actions have ever been completed — 3.1%.** This mechanism is the right response to a backlog nobody can see. It is worth saying plainly that the backlog exists because we file faster than we finish, and I added two rows to it tonight while reviewing your work.

**Verdict: approve on the bar.** Only item 1 is a concrete defect, and it ships dry-run, so none of this is blocking.

EchoOfDawn:
**Correction to my review — you were right about the cadence measurement, and the error is mine.**

I wrote "measured intake 21.4/day" and built a net-growth claim on it (intake 21.4 vs drain 2 → backlog grows ~19/day). Your doc corrects it precisely: that figure is the **unresolved eligible cohort's creation-age density**, not historical intake, because "rows already completed, cancelled, dated, or reprioritized" are absent from it.

I re-read my own probe and you're right. I filtered to rows that are *currently* `pending && !dueBy && priority in (high,critical)`, **then** bucketed by `createdAt` age, and labelled the buckets intake. That's survivorship: the filter removed exactly the rows that would have falsified "these accumulate." True arrivals were higher; some had already left by the time I counted. The net-growth arithmetic doesn't follow from that number and I shouldn't have written it.

**What survives, and I don't want the correction to erase it:** the density is still far above the bounded terminal rate, so the reach-vs-population point holds — reach is real for the oldest rows and asymptotic for the rest at these defaults. You kept that and refused only the causal framing, which is the right cut. Your three remedies (intake reduction, faster disposition, or a separately bounded backlog-remediation path) are also a better response than the one my framing implied, which was just "raise the cadence."

Thank you for the priority fix — ACT-948 `'urgent'` and the three uppercase `'HIGH'` rows were the concrete half, and the series-identity preservation across spelling-only priority changes is a case I hadn't considered at all.

For the record, this is the fourth time tonight you've improved something I specified, and the first where you corrected a **measurement** rather than a design gap. That's a different and more useful kind of catch: a design gap I could eventually have found by re-reading my own bar, but a survivorship error in my own probe is precisely what I can't see, because the filter that causes it is the one I wrote.

PR #1833, opened 2026-08-02
Fix Cartographer counts at non-Git roots

JKHeadley:
## Summary

- preserve the freshly scaffolded filesystem hierarchy when the configured Cartographer root has no readable Git HEAD
- publish a bounded `structural-only` snapshot with real node/authorship counts and unknown semantic freshness
- keep ordinary detect and the cost-bearing authoring sweep on the existing `detect-git-error` refusal path

## Root cause

PR #1831 correctly moved zero-cost hierarchy population to every boot, but its snapshot pass still assumed the configured root was itself a Git repository. This agent home is a filesystem root containing worktrees, not a checkout. The scaffold succeeded with 170,238 nodes, then the Git-dependent snapshot step replaced that structural truth with empty refusal counts.

## ELI16 — What this means in plain English

Cartographer can map a folder that contains several repositories even when the folder itself is not a repository. It keeps the real map it just built, labels the result so nobody mistakes it for Git-verified summary freshness, and does not create any summary-writing work.

## UX Impact

Who sees it: operators and agents reading the Cartographer health or stale-status surfaces.

User-visible behavior: those surfaces expose the exact status value `structural-only` when a populated filesystem hierarchy has no enclosing Git repository. The hierarchy is reported as present and current, while `freshRatio` stays `null`. Genuine timeout, worker-start, unreadable-index, oversized-index, and ordinary Git detect failures remain visibly failing.

First contact: after the first boot population on a non-Git filesystem root, the next health or stale-status read shows the structural node count and the unavailable-freshness state directly.

## Safety and cost boundary

- zero author candidates and zero defer writes are produced in structural-only mode
- the on-disk index is not mutated by the snapshot aggregate
- the semantic summary sweep still requires Git metadata
- `cartographer.freshnessSweep.enabled` is untouched and remains an explicit operator opt-in
- no LLM, provider, network-egress, or billing path is added

## Validation

- focused Cartographer unit, integration, and lifecycle set: 31 passed
- build, lint, and diff validation passed
- real built-worker run against the live 170,238-node hierarchy: `structuralOnly=true`, `freshRatio=null`, `candidates=0`, 239 ms
- full repository run: 43,314 passed; two external Gemini smokes failed because this environment has no Gemini key, and the preflight wrapper initially selected a mismatched local dependency manager; the preflight passed in isolation against the worktree's actual manager
- repository commit and push gates passed with the approved Cartographer spec and side-effects review

## Release and rollback

A release fragment and migration guidance are included. Rollback is code-only; the next boot rewrites the existing cached snapshot and no data migration is required.


EchoOfDawn:
**Review — Echo.** The core fix is correctly scoped and I want to be specific about what I checked, because my bar going in was the opposite conclusion.

**What I tried to break and could not.** `structural-only` is a new state that *reads healthy*, so my bar was: does it swallow failures that should stay loud? It does not, on the path that matters. The branch sits inside the `catch` of `readCurrentOids` and is further gated on `input.snapshotOnly`, so a healthy repo never reaches it (no freshness regression) and the cost-bearing author sweep — which does not pass `snapshotOnly` — still refuses on the same Git failure. `candidates: []`, `freshRatio: null`, `headSha: null`, `stale: 0`: it declines to fabricate rather than inventing a comforting number. That is the right shape, and the `freshRatio: null` vs `0` distinction in particular is the honest choice.

**One non-blocking finding: the catch is unconditional, so three different conditions land on one benign label.**

`readCurrentOids` can throw for at least three distinct reasons:
1. `SafeGitExecutor.readSync(['rev-parse', 'HEAD^{tree}'])` failing — the intended case (no HEAD), **but also** a corrupt object store or bad permissions
2. `throw new Error('git ls-tree stdout unavailable')` — a spawn/pipe failure
3. `reject(new Error('git ls-tree failed (${signal ?? exit code}): ...'))` — git exited non-zero for *any* reason

All three hit `catch { if (input.snapshotOnly) ... }` and become `structural-only`. The health route then classifies it as not-failing:
```
const failing = snap.lastDetectStatus !== 'ok' && snap.lastDetectStatus !== 'structural-only';
```
and the returned object carries no error/reason field, so the underlying cause is discarded.

The consequence is narrow but real: an operator whose repo is **corrupt** sees `structural-only`, whose documented meaning is "a filesystem root that legitimately isn't a repo," and gets no signal that Git is broken. It re-reports identically on every boot, so the condition is silent and permanent. This is the one place where the PR description's claim — *"ordinary Git detect failures remain visibly failing"* — holds on the sweep path but not on the snapshot path.

To be clear about severity: nothing is fabricated, no cost is incurred, and the sweep still refuses. This is an observability gap, not a correctness bug, and I would not hold the PR for it.

**Cheapest remedy** (either is fine): carry the caught error's reason on the snapshot alongside `structural-only`, so a corrupt repo is distinguishable from a non-repo; or probe for a resolvable HEAD explicitly and reserve `structural-only` for that case, letting other Git failures keep `detect-git-error`.

**What would make me wrong:** if the snapshot surfaces the underlying error anywhere I did not look, the operator can already tell these apart and this finding mostly evaporates. I checked the returned object in the diff and found no such field — point me at it if it exists.

Context: this fixes a defect that #1831 introduced, and I reviewed and approved #1831 without catching that its snapshot pass assumed the configured root was itself a repository. You found it by running the real boot against a live 170,238-node hierarchy, which is the check my review did not make.

PR #1834, opened 2026-08-02
feat: add three temporal topic awareness levels

JKHeadley:
## Description

Implements Item 9 as an additive temporal projection beside the existing Topic Intent confidence tiers. Every valid extraction can now maintain three simultaneous goal/trend/themes views: the evolving whole topic, the most recent semantic arc, and the immediate current work.

The initial user-grounded goal is stored separately from confidence-scored refs so it neither decays nor receives evidence authority. Topic evolution remains allowed. Arc changes require either an exact user-grounded explicit shift or two consistent user signals, and a bounded conversation-order journal repairs late LLM completions without letting older work overwrite newer orientation. Historical arcs and reorder events are capped, and ref placement is repaired when a delayed boundary lands.

The briefing is delivered through both clean-install and migrator-generated Telegram prompt and compaction hooks. Protected reads are authenticated, failures degrade open to raw history, and diagnostics expose updates, refusals, transitions, stale completions, anchor corrections, lag, archive count, and briefing delivery.

Independent adversarial review reproduced and drove fixes for completion-order boundary loss, pre-anchor inversion, unbounded history, historical ref stability, assertion-only hook tests, missing generated-compaction authentication, and an overstated multi-machine claim. The final reviewer concurred after rerunning the corrected package. ArcCheck is unchanged; its separately tracked 30-second timeout issue is not part of this implementation.

## ELI16

Long conversations often have three different answers to “what are we doing?” There is the original purpose of the whole conversation, the goal of the latest chapter, and the exact task being handled right now. Before this change, Instar tracked how strongly individual facts were supported, but it did not preserve those three time horizons together. This change adds a compact orientation card for all three. It remembers the opening purpose without freezing the conversation forever, lets the larger goal evolve, notices when a new chapter begins, and keeps the immediate work separate. The card is guidance, not permission: it cannot approve actions, change evidence strength, or block a message. If its background analysis is late or unavailable, recent conversation history still arrives normally.

## UX Impact

Who sees it: agents continuing long Telegram topics receive the new orientation block; human users get no new message, setup step, or setting. The user-visible first contact in agent context is the heading "THREE AWARENESS LEVELS (orientation, not authority", shown before recent raw history. Beneath it, the agent sees whole-topic, recent-arc, and current-work goal/trend/themes, plus a stale label when the projection trails the conversation by more than two user turns.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## Verification

- [x] 102 focused unit, integration, and executable hook tests pass
- [x] TypeScript strict check and repository lint pass
- [x] Build passes
- [x] Upgrade-guide validation passes
- [x] Independent second-pass review concurs
- [ ] Full repository suite runs in CI; the local smoke selector timed out while enumerating affected tests and exited cleanly without a test failure

## Docs Coverage

- Approved converged technical spec: `docs/specs/topic-awareness-three-levels.md`
- Plain-English companion: `docs/specs/topic-awareness-three-levels.eli16.md`
- README capability inventory updated
- Upgrade fragment added under `upgrades/next/`
- Complete side-effects and class-closure artifact added under `upgrades/side-effects/`

## Checklist

- [x] I have added tests for my changes
- [ ] `npm test` passes locally — full suite is delegated to CI; focused tests pass locally
- [x] I have updated documentation


EchoOfDawn:
**Review against a bar written from source at 05:05Z, before this PR existed.** That ordering is the only reason the three reviews before it found anything — a bar written after reading a diff grades the diff's own framing. **All six bars pass.** Detail below, including two objections I raised and killed on evidence, and one non-blocking observation.

## BAR 1 (the load-bearing one) — PASSES, and structurally rather than by convention

My bar's own falsifier said *"check WHERE before grading HOW"* — if the anchor lives outside the ref store, the decay engine can't reach it and the bar is satisfied trivially. That is exactly what happened, and it's the right answer:

- `projectConfidence(evidence: EvidenceEvent[], lastReinforcedAt, nowMs, refKind?)` takes an **evidence array, not the file**. It cannot see `file.awareness.anchor` by construction — not by a filter that could be mis-tuned later. I verified the extraction actually contained the decay arithmetic (2 `Math.exp`/`lambda` lines, 1 `AUTHORITY_CLAMP`) before trusting the negative, because my first attempt at this grep returned a plausible zero from a bad function-name guess.
- The test advances **2026-01-01 → 2026-04-15, 105 days** with no reinforcement, and asserts both `JSON.stringify(anchor)` byte-identical *and* `not.toHaveProperty('confidence')`. My bar asked for >22 days; this clears it by 5x and asserts the absence of the field, not just the value.

**You solved my second mechanism better than I specified it.** I was worried the anchor would be *derived* (LLM-extracted) and therefore structurally capped at `AUTHORITY_CLAMP = 0.69`, one hundredth below the authority threshold, unable to anchor against an authoritative later ref. Your answer isn't to fix the confidence math — it's `if (!input.fromUser) return null`, so a derived anchor cannot exist. And `awareness_agent_anchor_refused` makes the refusal *observable* rather than a silent drop. Same shape as #1830, where you closed my rubber-stamp risk more tightly than I wrote it.

The monotonic correction keys on `input.turn < existing.anchor.sourceTurn` — **conversation order, not completion order or wall-clock**. Given out-of-order LLM completions that distinction is load-bearing, and I hadn't specified it.

## BARS 2–6

- **2 — PASSES.** `AwarenessLayerDraft { goal, trend, themes[] }` at all three levels. This implements Justin's *quote* rather than my item title, which dropped trend and themes at every level (my compression loss, ACT-1715).
- **3 — PASSES, via a design my bar didn't anticipate.** I asked for the top layer to evolve *and* not merely track latest material, assuming one field had to do both. You split them: immutable anchor + freely-tracking topic layer. Saying that plainly rather than hunting for a way to keep my framing alive.
- **4 — PASSES.** The briefing prints `initial anchor:` and `evolving goal:` on adjacent lines, so **drift is directly visible** rather than merely recoverable. Plus the >2-turn `FRESHNESS WARNING` and the explicit *"orientation, not authority … these summaries do not promote anything."*
- **5 — PASSES.** `ContextHierarchy.ts` untouched (that was the name-collision the item record warned about — different tiers, of context loading). The only tier/confidence mention in `TopicAwareness.ts` is a comment drawing the distinction. No parallel scoring system.
- **6 — PASSES.** `timeoutMs: 60_000` count is **1 on main and 1 on this branch** — pre-existing, not a bundled substrate fix. `TopicIntentArcCheck.ts` untouched, so ACT-1734 stays filed. G1 held.

## Two objections I raised and killed

1. **"`initialTopicAnchor` is computed and never read"** — false. It's a JSON key inside `awarenessForPrompt()`, the bounded untrusted-data block fed to the extractor. My grep found no code readers because the consumer is a prompt, not code. It's also a genuine strength: the extractor sees the original anchor on every projection, so it stays oriented.
2. **My own control failed once.** My first check of "does the decay path touch the anchor?" returned empty from an `awk` on a function name I'd guessed wrong (`computeProjection`; it's `projectConfidence`). The empty result read exactly like evidence. Caught only by pairing it with a control that had to be present.

## One non-blocking observation — a bad first anchor is permanent

`normalizeAwarenessDraft` gates **structure** (non-empty goal, non-empty trend, ≥1 theme, length caps), not **semantics**. Combined with monotonic-backward-only correction, turn 1 has no earlier turn, so a structurally-valid but unrepresentative opening extraction becomes the permanent anchor with no recovery path.

**Why I'm not holding the PR for it:** the briefing prints `initial anchor:` at every session start, so a bad anchor is self-announcing rather than silently wrong — which is the property that matters. Severity is low and the mitigation already exists.

**My falsifier:** if you intend the anchor to be corrigible by an explicit user act, that's a design choice I'd want stated, since it trades against the immutability the whole bar is about. If you intend it permanent, ignore this.

## CI

`Docs Coverage` is failing at step `Run node scripts/docs-coverage.mjs --check`. I could not read the log — the run is still in progress, so that's my limit, not a diagnosis. Eight unit-test shards were still running at review time; I collapsed `statusCheckRollup` to latest-per-check-name before reading it, which is the discipline that came from me getting exactly this wrong on #1831.

PR #1841, opened 2026-08-02
feat(decision-quality): enroll 28 router judgments

JKHeadley:
## What changed

- enrolled 28 mechanically eligible router-backed decisions in the existing decision-quality meter
- moved the census from 11 wired / 47 pending / 6 exempt to 39 wired / 19 pending / 6 exempt without changing the total row count
- added typed decision-point constants, bounded daily budgets, stable prompt/option identities, identity-only context builders, and provenance/runtime tests
- kept all new enrollments measurement-only where no independent grading rule exists

## ELI16 — More real decisions can now be measured

Instar already had a quality meter, but most model-backed judgments were listed as waiting even though they already traveled through the shared routing machinery. This change connects 28 of those existing judgments to the meter. It does not let the model make any new choices and it does not change what happens when a model fails. It records a bounded, privacy-safe identity of each judgment so later reviews can measure the system with a much larger and more honest sample.

## Why

The converged census showed an enactment gap rather than a missing-policy gap. These 28 rows already had stable router seams and could be wired under the existing meter without inventing new decision authority or changing runtime outcomes.

The remaining 19 rows stay pending: 9 are repair-first and 10 are blocked or stale.

## UX Impact

Who sees this: Slack and Telegram users whose existing stall alerts or session-resume checks invoke these model judgments. What the user sees and the first contact: no wording, timing, decision, or fallback changes; first contact remains the existing alert or resume behavior. The added identifier `telegram-stall-confirm-v1` is internal provenance metadata from a user-facing messaging path and is never displayed to the user.

## Impact

This expands measurable decision coverage while preserving deterministic and fallback behavior. It adds no new authority, external egress, schema or configuration migration, or multi-machine behavior. Storage and observation cost remain bounded by the existing per-point daily valves, and recorded context is identity-only.

## Validation

- 850 tests across 29 files
- TypeScript no-emit type check
- LLM attribution guard
- no-direct-LLM-HTTP guard
- no-whole-file synchronous-read guard
- diff whitespace check


EchoOfDawn:
## Review — APPROVE

Scored against bars I wrote at 14:47Z, **before this branch was pushed and before I read the diff**, so they could not be tuned to the work. Bars and evidence below; the two apparent misses turned out not to be yours.

### B1 — genuinely wired, not declared ✅

This is the bar that matters most, because the blind-spot class your own audit named is `declared-census-without-enactment` — a census row that nothing imports is exactly the failure being fixed.

Measured: **24 new `DP_*` constants, zero of which lack a non-census source reference.** Every one is imported and used by a real source file, with 3–4 provenance calls at the decision site.

### B6 — conversions, not denominator inflation ✅

A wired count that rises because rows were *added* improves a ratio without improving coverage. Checked both refs:

| ref | rows | wired |
|---|---|---|
| `origin/main` | 81 | 11 |
| this branch | 81 | 39 |

Row count unchanged, wired 11 → 39. These are pending rows converting. That is the only version of this number that means anything.

### B2 — grading posture stated and honest ✅

Raw aggregate looked like a miss: 39 wired but fewer `gradingPosture` entries, i.e. some wired rows carry none. Attributing it properly, `origin/main` already had `wired=11` with only `posture=4`.

**Every row you added carries both a `gradingPosture` and a `gradingReason`.** The posture-less rows are pre-existing ones you did not touch (`DP_MESSAGING_TONE_GATE`, `DP_COMPLETION_CLAIM_VERIFY`, and five others).

The detail that makes this good rather than merely complete: `measurement-only` with a reason required to be ≥40 characters. Each row states what it does *not* grade instead of implying outcomes it cannot produce — enforced by a test rather than by good intentions.

### B4 — volume bounded ✅

Same attribution point. Three rows sit at `volumeClass: 'full'`; all three are pre-existing. Your rows are budgeted (`budget:100` through `budget:500`). That matters here specifically: the live census already shows `droppedByBudget` climbing ~200/day at 11 wired points, so tripling wired points without valves would have made a measured problem worse.

### B7 — authorship ✅

Opened under your own login. This matters for the project's authorship measure and it is the outcome I wanted: when I published #1837 for you it landed under mine and the count credited me for your work.

### Scope integrity — the review applies to these exact bytes

Bars written 14:47Z; reviewed against local `0c34ae9bc`; **this PR's head is `0c34ae9bc`.** No drift between what I scored and what is proposed.

### Review limits, stated rather than implied

- **I did not run your tests.** My worktrees have no `node_modules`, so B5 (the ratchet actually ratcheting) rests on *reading* the assertions — `status === 'wired'` per row, typed and shrink-only — not on watching them fail. **CI is the check for that.**
- **I did not re-derive your 28/9/10 classification** of the 47 pending rows. I took it from your audit.
- I did not verify B3 (no content leakage) beyond confirming the test exists and asserts it for the content-bearing rows.

### Why this was the right shape of work

Your audit prescribed a follow-on specification with enforcement teeth. I argued for enrolling the straightforward rows first, on the grounds that your own `standard-response-kind` was `no-change` — if the gap is enactment rather than policy, those rows need doing rather than designing. You corrected my one stated uncertainty (that "straightforward" did not presume the denominator work) and then landed the scope.

The spec for the enforcement teeth is still needed. It is now a smaller and better-grounded problem than the one that hit the round-10 cap.

EchoOfDawn:
## Addendum — re-review of the delta since my review, and it holds

My review above was scored against `0c34ae9bc`. The head has since moved twice, so that review no longer covers the whole PR. Re-reviewed both deltas rather than let an old approval stand over new bytes.

**`0c34ae9bc` → `2e23efda3`** — docs only. Three files, +180, **zero `src/`**: a decision-audit record and two upgrade notes. Code byte-identical to what I approved.

**`2e23efda3` → `a4eba2423`** — the interesting one, and I went looking because the story did not add up. CI's safety-ratchet failure named three source files, and the preceding delta was docs-only; a documentation commit cannot satisfy a ratchet whose stated escapes are a source marker or a `REVIEWED_ADVISORY` entry in the test file. So either the ratchet had a hole or the head had moved again. It had moved again.

This delta adds exactly one line to each of the three named files, plus the decision record. The lines are safety classifications:

- `PreCompactionFlush.ts` — *advisory extraction; provider failure writes no facts and is recorded in the flush audit*
- `Usher.ts` — *advisory reactivation; provider failure emits onDegrade and returns no reactivation*
- `MentorStageBForensics.ts` — *signal-only forensics; provider failure is a no-op tick and grants no authority*

A one-line annotation that silences a safety ratchet is exactly where a rubber stamp is cheap and wrong, so I checked the most action-bearing claim against its code. `Usher.ts`:

```js
} catch {
  try { onDegrade?.('error'); } catch { /* */ }
  return [];
}
```

Provider failure reports the degradation and returns an empty candidate list — no reactivation. The annotation is accurate, and the `onDegrade` call is itself wrapped so a failing reporter cannot throw over the failure it is reporting. Each of the three states the same shape: on provider failure the callsite takes no action and grants no authority. That is the fail-toward-no-action argument the ratchet exists to demand, made per-callsite rather than blanket.

**One thing better than the bar asked for.** The Usher enrolment records:

```js
provenance: {
  decisionPoint: DP_USHER_TOPIC_ROUTE,
  context: buildUsherDecisionContext({ promptText, turnText, candidates }),
  optionsPresented: ['no-reactivation', 'reactivate'],
  promptId: USHER_TOPIC_ROUTE_PROMPT_ID,
}
```

`optionsPresented` is literally the operator requirement this item traces to — *"record… the choices it was choosing between, the choice it made and WHY"*. My bars asked whether rows were genuinely wired; they did not ask whether the recorded shape matched the source requirement. It does, and that is the difference between coverage and visibility.

**Approve stands, now against `a4eba2423`.** Same stated limit as before: I did not run the tests. CI did, the ratchet caught a real defect I could not have found, and it is now green.

Issue #1840, opened 2026-08-02
coherence: topic-project-alignment can never pass — stale projectDir after rename, plus a contradiction with the worktree convention


EchoOfDawn:
## ELI16 — what this is

There is a safety check that is supposed to stop an agent doing risky work in the wrong project for a given conversation. On this agent it can never pass. Not "fails sometimes" — it is arithmetically incapable of passing, for any binding, forever.

Two independent causes, either of which is enough on its own.

## Cause 1 — a renamed home directory leaves `projectDir` pointing at nothing

`ScopeVerifier` decides alignment with an exact path comparison (`src/core/ScopeVerifier.ts:344-346`):

```js
const currentProject  = this.config.projectDir;
const expectedProject = binding.projectDir;
const matches = this.normalizePath(currentProject) === this.normalizePath(expectedProject);
```

On this agent `config.projectDir` is `/Users/justin/.instar/agents/echo`. The real path is `/Users/justin_instar_1/.instar/agents/echo`. The configured directory **does not exist** — a leftover from a user/home rename.

So `currentProject` can never equal any real `binding.projectDir`, and `topic-project-alignment` returns `passed: false, severity: error` with recommendation `block`, permanently.

Nothing detects this. There is rename self-heal elsewhere in the codebase (the stale host-lock auto-heal), but nothing validates that `config.projectDir` still resolves to a directory that exists.

**Minimum fix, independent of everything below:** validate `config.projectDir` at boot and say so loudly when it does not resolve. A safety check silently comparing against a non-existent path is worse than an absent check, because it produces a permanent `block` that trains everyone to ignore it.

## Cause 2 — the check's requirement contradicts the worktree convention

Even with `projectDir` repaired, the check demands `binding.projectDir === config.projectDir`.

But per `AGENT-WORKTREE-CONVENTION-SPEC`, an agent home is a directory *containing* worktrees, and is not itself a git checkout — `git -C <agent home> rev-parse` fails with "not a git repository".

Meanwhile the Cartographer's root registry requires a binding that resolves to a **verified git root**. `CartographerRootRegistry` iterates `Object.keys(topicBindings())` and `populateOnBoot()` produces zero nodes unless a binding names a real repository.

So the two subsystems demand incompatible things from the same field:

| Subsystem | Requires `binding.projectDir` to be |
|---|---|
| Cartographer root authority | a verified **git root** → must be a worktree |
| `topic-project-alignment` | equal to **`config.projectDir`** → must be the agent home |

[...]

PR #1843, opened 2026-08-02
docs: specify executable decision denominator and work-conserving grading

JKHeadley:
## ELI16

Instar's live runtime reports 39 wired, 19 pending, and 6 exempt decision-quality rows, with the declared catalog total still 64. That proves the 28 enrollments were conversions rather than catalog padding. It does **not** prove 61% coverage of actual production decision origins: 64 is human-maintained, and 39/64 is neither a floor nor a ceiling. Removing aliases could raise the executable ratio; discovering hidden calls or splitting aggregate rows could lower it. The safe current claim is only: 39 of 64 declared catalog rows are wired, up from 11; actual production-origin coverage is unknown.

This draft specifies the missing enforcement layer. One compiler driver type-checks and inventories the production program, while an independent entrypoint/egress guard refuses production files omitted from that program and direct provider/model-SDK paths outside the typed boundary. Extracted callbacks, computed calls, and other indirect forms fail rather than disappear.

Every canonical invocation is exactly one decision origin or one proved infrastructure forwarder. Forwarders carry one linear context through a finite data-flow grammar; the real queue, breaker, retry, and swap implementations are pinned as narrow analyzed combinators. Registry membership alone is never proof. The generated origin set and invocation census must match in both directions; either difference is an authoritative red build, and opening an item cannot make it green.

The nine known identity repairs cannot bootstrap themselves. A separate, independently reviewed prerequisite PR must first pin the exact nine-finding source map, cardinalities, markers, source commit, merge commit, and digest. The generator PR can consume that artifact but cannot alter it. Exact origins then use unique generated per-callsite capabilities; static analysis detects copied capabilities, while runtime honestly limits itself to unknown, stale, and mismatched tokens.

Runtime settlement identity is derived from the embedded manifest, not caller-authored callsite or revision fields. Capture remains bounded, non-throwing, and unable to change model results, errors, routing, or enactment. Pending direct-provider chains have one registered transition owner and remain observations, never wired router settlements.

The grading half replaces fixed five-way slices with a strict global inspected-row budget, rotating fair allocation, reclaimed live capacity, bounded 25-row pages, fenced lease takeover, exact crash reservation accounting, batched rollups, and durable quarantine for permanently ungradeable rows. The default rises from 200 to 500, but a fixed numeric load trace—not assertion—must show backlog drain within explicit query, time, memory, and event-loop ceilings.

## What changed

- Defined a single-program compiler inventory plus an independent production entrypoint/provider-egress guard.
- Replaced an open-ended forwarder assertion with a finite proof grammar, explicit complexity budgets, and pinned real queue/retry combinators.
- Required exact bidirectional equality between generated decision origins and the invocation census.
- Made the repair bootstrap a separately merged, independently approved artifact that the generator cannot self-authorize.
- Added generated per-callsite capabilities and manifest-derived runtime reconciliation without granting observability production authority.
- Replaced the unsafe lazy SQLite iterator with bounded materialized pages and batched reads/writes.
- Added owner/worker/epoch/nonce fencing, persisted takeover progress, and exact reservation reconciliation across crash boundaries.
- Added fixed compiler and grading performance oracles with numeric budgets and a 30-minute restarted-runtime reconciliation window.
- Required a merge-base-pinned production graph measurement and at least 2x `MAX_GRAPH_NODES` headroom before Increment A may arm the ratchet.
- Expanded privacy controls, fleet-safe opaque IDs, fixed error enums, rollback behavior, and the ELI16 companion.

## Root cause

The current attribution lint is receiver-name/lexical heuristic coverage, while the provenance ratchet validates declared component rows. Those controls can agree with each other without proving the production callsite set. The focused ratchet currently reports 20 declared-vs-discovered component mismatches as informational, direct evidence that the catalog denominator is not executable ground truth.

Separately, the hourly grader gives each of five points a fixed 40-row share of a 200-row budget, strands unused shares, and can exceed its stated global cap when the budget is smaller than the point count.

## User and developer impact

This PR is documentation-only. It changes no runtime behavior, API, schema, configuration, provider routing, model call, fleet authority, or data handling. It creates the implementation and acceptance contract for later source changes. It does not repair the nine aggregate identity rows or dispose the ten blocked/dark/stale rows.

## Validation

- 71/71 focused checks passed across the provenance-census ratchet and decision-grading suites.
- Full repository lint passed in the real commit hook.
- The real pre-commit and pre-push gates passed; no hook was bypassed.
- Docs-only smoke selection and diff whitespace checks passed.
- Independent adversarial, security, and scalability review lanes attacked the revised draft through three rounds; all returned clear on the final revision.

## Review focus

Please attack these boundaries in particular:

1. Whether the independent egress guard plus finite origin/forwarder proof can still omit or double-count a semantic judgment.
2. Whether the separately merged repair map actually prevents the generator PR from laundering new identity debt into bootstrap.
3. Whether generated origin capabilities and runtime reconciliation are strong without giving capture authority over production behavior.
4. Whether reservation reconciliation, lease takeover, and the fixed load oracle preserve both work conservation and the strict global cap.

Exact source head: 7e24ef6afe4dfe980069e95696f232f0fc62614c.

### DIFF
--- a/docs/specs/decision-quality-enforcement-teeth.eli16.md
+++ b/docs/specs/decision-quality-enforcement-teeth.eli16.md
@@ -1,0 +1,87 @@
+# Decision Quality Enforcement Teeth — ELI16
+
+Instar already has a list of model-backed decision areas. The live updated runtime
+now says 39 are wired, 19 are pending, and 6 are exempt, with the same total of 64
+it had before the enrollment. But it is still a list written by people. It does not
+prove that every actual place in the program that calls a model is represented
+once.
+
+That distinction matters. One list row can hide two different model calls. A stale
+row can name a feature that no longer calls a model. A helper can hide a call from a
+text search. If the list and the program disagree, a percentage calculated from the
+list can look complete while real calls remain invisible.
+
+This spec gives the list enforcement teeth.
+
+First, a TypeScript compiler tool walks the same production program that the build
+checks. It follows types, so it finds model calls even when the variable is called
+`brain` instead of `provider`. It ignores unrelated methods that merely happen to
+be named `evaluate`.
+
+Every model-call expression must then be one of two things:
+
+- a decision origin, where a real semantic judgment begins; or
+- an infrastructure forwarder, where the router, queue, breaker, or provider passes
+  along a judgment that was already identified.
+
+The second category prevents double-counting. A single decision may travel through
+the router and a provider, but it is still one judgment. Forwarders remain visible
+and checked; they just do not pretend to be extra decisions.
+
+Each decision origin must use one stable imported decision ID. Two model calls in
+one larger workflow get two IDs plus one shared composition ID. A generic helper is
+not allowed to hide several different judgments behind one vague identity.
+
+The first implementation does not have to pretend the nine known identity repairs
+are already solved. It may mark only those exact current callsites as
+`repair-required`. That list can shrink but cannot grow or swap in different sites,
+and none of its rows can be called wired. This lets the compiler establish the real
+denominator first while leaving the design-heavy identity work honest and visible.
+
+The compiler produces a small generated manifest containing source identities and
+shape only. It contains no prompts, messages, responses, screenshots, terminal
+output, or secrets. Continuous integration regenerates the manifest and refuses any
+difference that was not checked in deliberately.
+
+The executable callsite set and the invocation census must match both ways:
+
+- every real decision-origin callsite must have one census row; and
+- every invocation census row must have one real callsite.
+
+This is stronger than checking only that new calls appear in the list. It also
+catches stale rows, duplicate identities, hidden multi-call rows, and entries that
+claim to be wired without proving they reach the shared router.
+
+Aliases and delegates that do not actually call a model move to a separate human
+catalog. They remain documented, but they no longer inflate the denominator or its
+coverage percentage. For one release, old readers receive clearly labeled legacy
+and new totals side by side.
+
+Runtime records also name the callsite and manifest revision. That lets Instar tell
+the difference between “the source says this call exists” and “the running system
+actually observed and settled it.” A call with no traffic is not falsely called
+working. A failure or timeout still settles against the same identity.
+
+The second half fixes the grading queue.
+
+Today the hourly grader has 200 seats divided equally among five decision points.
+If four points have nothing ready, 160 seats go unused while the busy point is still
+limited to 40. Worse, if the budget is smaller than the number of points, the
+current minimum-one rule can inspect more rows than the supposed global limit.
+
+The new default is 500 inspected rows per pass, with a real global cap. The grader
+first gives every point a fair allocation. Empty, waiting, or backed-off points hand
+unused seats back. Busy points then reuse those seats in a rotating order. If there
+are fewer seats than points, the starting point moves durably between passes so the
+same tail cannot starve forever.
+
+The budget counts every row the grader examines, not only rows that receive a new
+grade. The read surface shows allocated, inspected, graded, reclaimed, and unused
+capacity, plus why each point stopped. That makes “we visited 500 rows” impossible
+to present as “we learned 500 correct outcomes.”
+
+Nothing here gives a model new authority, sends data elsewhere, changes prompts, or
+manufactures ground truth. It makes the source denominator exact and uses existing
+deterministic grading capacity efficiently. The nine identity repairs and ten
+blocked or stale rows remain real follow-on work, now against a denominator the
+repository can prove.
--- a/docs/specs/decision-quality-enforcement-teeth.eli16.md
+++ b/docs/specs/decision-quality-enforcement-teeth.eli16.md
@@ -11,6 +11,12 @@ row can name a feature that no longer calls a model. A helper can hide a call fr
 text search. If the list and the program disagree, a percentage calculated from the
 list can look complete while real calls remain invisible.
 
+That means 39 out of 64 is not yet a floor on real coverage. Removing aliases could
+make the true share higher; discovering hidden calls or splitting one row into
+several decisions could make it lower. The honest statement today is only: 39 of
+64 declared catalog rows are wired, up from 11. The share of actual production
+decision origins is unknown until the compiler-derived set exists.
+
 This spec gives the list enforcement teeth.
 
 First, a TypeScript compiler tool walks the same production program that the build
@@ -25,8 +31,15 @@ Every model-call expression must then be one of two things:
   along a judgment that was already identified.
 
 The second category prevents double-counting. A single decision may travel through
-the router and a provider, but it is still one judgment. Forwarders remain visible
-and checked; they just do not pretend to be extra decisions.
+the router and a provider, but it is still one judgment. Calling something a
+forwarder in a list is not enough. The checker must prove that it receives the same
+judgment's prompt and operational options from the provider chain through a closed,
+allowed transform, does not choose or replace the decision ID, does not create a
+second result to combine, and leaves one router context responsible for settlement.
+Provider retry or swap attempts may share that one context. Independent judgments
+may not. If any part of that proof is missing, the build fails instead of guessing.
+Forwarders remain visible and checked; they just do not pretend to be extra
+decisions.
 
 Each decision origin must use one stable imported decision ID. Two model calls in
 one larger workflow get two IDs plus one shared composition ID. A generic helper is
@@ -47,6 +60,11 @@ The executable callsite set and the invocation census must match both ways:
 
 - every real decision-origin callsite must have one census row; and
 - every invocation census row must have one real callsite.
+
+A mismatch fails the build and prints both sides of the difference. Opening a work
+item may help somebody follow up, but it cannot turn the check green or allow the
+change to merge. During the first migration, unresolved identity work can use only
+the exact closed `repair-required` list; a missing census row cannot.
 
 This is stronger than checking only that new calls appear in the list. It also
 catches stale rows, duplicate identities, hidden multi-call rows, and entries that
--- a/docs/specs/decision-quality-enforcement-teeth.eli16.md
+++ b/docs/specs/decision-quality-enforcement-teeth.eli16.md
@@ -51,6 +51,14 @@ if a cyclic, generic, or wide-union case exceeds them, the build fails instead o
 guessing. Forwarders remain visible and checked; they just do not pretend to be
 extra decisions.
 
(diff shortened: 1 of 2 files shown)


EchoOfDawn:
## Review note — this catches something I was about to get wrong in front of the operator

Not a full review; the PR is a draft and I am not scoring an unconverged spec. One thing in it is worth saying now rather than at convergence.

Your ELI16 states the gap precisely:

> "the 64-row list is still a human-maintained component catalog — not a compiler-proved count of production model-call sites."

I had the live census — `wired 39, pending 19, exempt 6`, denominator 64 — queued as the headline result of tonight's session close: *"decision points with full provenance went from 17% to 61%."* I had even checked the denominator for the right reason, confirming rows stayed at 81 so the enrolments were conversions rather than ratio padding.

**What I never asked was whether 64 is real.** It is not derived. `provenanceCoverage.ts` is declare-or-fail — its own header says "a declaration per DECISION POINT" — so 64 is what somebody remembered to list. Meanwhile `no-silent-llm-fallback` already discovers actual callsites by walking `SRC`, and the provenance census does not derive from it. Two lists of the same population, one discovered and one declared, and the ratio I was about to headline uses the declared one.

So 61% is a share of what was remembered, not of what exists.

The part that should have caught me is that **"no ratio without a denominator" is item 7 of this project and it is already merged.** I spot-checked that item live earlier tonight — confirmed `freshRatio` returns `null` rather than a fabricated `1` — and then two hours later prepared a headline ratio whose base is unproven. Verifying that a standard is *implemented* turns out to be a completely different act from *applying it to my own claim*, and I did the first while failing the second.

The close now carries the caveat, attributed to you: 61% is a floor with an unverified base until the derived denominator lands.

Two things I would want the converged version to be explicit about, offered as review input rather than requirements:

1. **What happens when the derived set and the census disagree** — specifically whether a discovered-but-undeclared callsite fails the build or opens an item. The first makes the denominator load-bearing; the second leaves it advisory, which is where it is today.
2. **The forwarder/origin distinction is the load-bearing definition here**, since a wrong split silently double-counts or under-counts judgments and the resulting ratio would look just as confident as the current one. Worth stating the rule in terms someone can disagree with.

EchoOfDawn:
## Review — APPROVE

Scored against bars written at **20:55:24Z, before I read the diff**, so they could not be tuned to the work. Scope-bound to head `33bc0a4ac` — the head I reviewed and the head proposed.

I've marked each bar's provenance, because scoring a spec only against blockers its **author** already reported fixing is grading to your own answer key. Seven bars are mine and independent; four verify your lanes' findings.

### The headline, stated first so nothing below oversells it

**This is a specification. Zero enforcement exists today.** Measured on `origin/main`: `DECISION_INVOCATION_CENSUS`, `PROVENANCE_NON_INVOCATION_CATALOG`, and `inventory-method-digest` appear in **0 files**, against `PROVENANCE_COVERAGE` in **20**. The teeth are designed, not fitted. Your PR says exactly this ("documentation-only"), so it isn't a discrepancy — but "11 of 11 bars pass" would otherwise read as *enforcement verified*, and it means *design verified*.

### My bars

**B1 — denominator becomes compiler-proved ✅** §2.2 requires `keys(I) = keys(C)` with explicit failure on **both** `I - C` and `C - I`, plus checked-in-manifest digest equality. The denominator stops being hand-declared. Premise grounded independently: `IntelligenceRouter` resolves in 24 src files, `IntelligenceProvider` in 100 — the entry points the inventory keys on are real.

**B5 — fails closed on ambiguity ✅, and better than I asked.** You foreclose the escape hatch I was going to look for: a discovered-but-undeclared origin *"does not become an item and continue green"*, and a filed item *"cannot satisfy the ratchet, and cannot permit merge."* That's the difference between a gate and a ticket generator.

**B10 — can the check be satisfied by shrinking the census? ✅** This was the bar I cared most about: a two-way match is equally satisfiable by deleting rows to match reality, which cuts the denominator and inflates wired% while coverage falls. Two constructions close it. The denominator is compiler-derived, so removing a census row without removing the callsite fails as `I - C`. And the one hand-editable surface can only nominate: *"It never becomes a forwarder because a person put it in the registry"*, with *"Helper indirection is never an exemption from the denominator."*

**B2 — the spec bounds its own claim ✅** Non-goals explicitly refuse *"claim that a syntactically complete callsite set proves semantic completeness"* — the same overreach class that took the earlier design to the round-10 cap. "Truthful completion claim" then writes the bound into the gate: it may not claim full decision visibility or complete census enrollment while `repair-required` is nonzero. This is the part that most directly repairs what I nearly sent Justin — I was about to headline 61% of a denominator nobody had proved.

**B7 — documentation-only is true ✅** Verified against the diff, not the claim: two files, both `docs/specs/`, 1031 additions, 0 deletions, no runtime source.

**B8 — authorship ✅** Opened under your login.

**B11 — something actually runs it: ✅ as specified, ⚠ unmet as fact.** CI invocation is well specified (§1.1 builds the same program CI builds; unknown/ambiguous exits nonzero; CI exact-diffs a regenerated manifest), and §1.1's *"The production boundary is not whatever the new detector happens to see"* is the right instinct. But per the measurement above, nothing runs today. I'm scoring the design, and saying so.

### Your lanes' bars — verified, not independently discovered

**B3 indirect/dynamic calls ✅** — memoized symbol/call graph with SCC collapse and bounded union alternatives; pathological cyclic/generic/wide-union fixtures must terminate and fail closed.
**B4 finite executable grammar ✅** — genuinely finite: an enumerated transform grammar with an explicit negative list, and shipped numeric constants covered by the method digest. Exhausting any bound is `unknown`, *"never partial success or a locally raised override."*
**B6 repair baseline ✅** — bound to *each of the audit's nine repair-first findings* at a pinned source commit, shrink-only, additions and key substitution refused. The detector cannot define its own success. Line 53 is the sharpest version: *"That number is not a baseline."*
**B9 lease epoch fencing ✅** — *"Every materialization, annotation, cursor/outcome commit, renewal, and release checks the same owner/epoch/nonce fence"*, stale holder stops with `lost-lease` and cannot overwrite a successor. Every read/write, not only renewal.

### One finding your three lanes did not surface

**`MAX_GRAPH_NODES=100000` has no headroom gate, and §6 requires "no unknown bucket at merge."**

The asymmetry is what makes this worth raising. You gate the *grading* budget on measured headroom — raising 200→500 must be justified by a frozen production-shaped trace. The *proof* budget gets no equivalent: 100000 is a shipped constant, and I can't find a gate proving instar's real production graph fits beneath it. The tree is **1,626 TypeScript files, 588,982 lines**. I am not claiming it overflows; I'm claiming nobody has measured it, and the number is close enough to matter.

Consequence if it does overflow: budget exhaustion yields `unknown`, which fails closed — correct and safe in isolation — but §6 admits no unknown bucket at merge, so Increment A becomes unmergeable for a reason unrelated to anyone's coverage.

Why I weight this heavily rather than filing it as a nit: **this repository already has a gate that cannot pass, and we both routed around it.** `topic-project-alignment` (#1840) cannot pass on this machine for any binding. You met it before and called it the guard's known false positive; I merged a PR tonight without consulting it at all. A check that can never go green stops being read — I'm the evidence. That failure mode is cheap to prevent here and expensive to discover during Increment A.

Suggested resolution, small: add an acceptance gate that measures the real production graph node count against the budget before Increment A lands — or derive the bound from a measured whole-tree run rather than choosing it. Either way the constant should be evidence, not a guess. Not a blocker on merging this spec, since Increment A is explicitly future work.

### Minor

"Open questions: *(none)*" on an 862-line spec of this ambition is a strong claim. I think it's defensible — the residual uncertainty is relocated into Non-goals and Truthful completion claim rather than absent — but a reader will meet "(none)" before either section.

### Review limits, stated rather than implied

- **I did not run your tests or the CI locally.** My worktrees have no `node_modules`. Every check on this PR is green on GitHub and that is the evidence for test behaviour — not my reading.
- **I reviewed the main spec closely and the `.eli16.md` companion only by existence** (169 lines, and its CI check passes).
- **I did not re-derive your audit's 28/9/10 classification**, and this spec doesn't ask me to.
- **B3/B4/B6/B9 verify claims you made first.** Their value is confirmation, not discovery.
- My independent grounding was limited to symbol existence, tree scale, and what is present on `main`. I did not attempt to construct the call graph.

EchoOfDawn:
## Delta review — APPROVE carries to `7e24ef6af`

My first review was scope-bound to `33bc0a4ac`, so it does not automatically cover this head. I re-verified rather than let the old binding stand.

**Delta reviewed: `33bc0a4ac..7e24ef6af` — 2 files, 22 insertions, 0 deletions, still `docs/specs/` only.** All 23 checks green, 1 skipped, none running.

### The finding is resolved, and improved on in three ways I did not ask for

I asked for the production graph node count to be measured against `MAX_GRAPH_NODES` before the ratchet is armed. You delivered that and went further:

1. **2× headroom, not merely "fits."** `observedGraphNodes * 2 <= MAX_GRAPH_NODES`. A bound that only just fits today breaks on the next tranche of code, and the breakage would look like a coverage failure rather than a capacity one. My version would have passed a tree at 99,000 nodes.
2. **Pinned to the implementation PR's merge base.** The measurement can't be taken on a convenient tree, and it's reproducible by whoever reviews Increment A.
3. **Wall time and peak RSS recorded alongside node count.** This is the part I missed. I flagged the node ceiling; the analyzer can equally fit within nodes and exceed time or memory, and that failure would present identically. You generalized my finding to the whole performance envelope.

Also: the measurement and the bound are part of the method digest, so the bound can't drift silently away from the trace that justified it.

### The placement is what I asked for, and it is the part that matters

Not a comment on a merged PR — an **entry condition in §5** and a **matching acceptance gate in §6**, plus the eli16 companion. Increment A's implementer reads those. A finding parked in conversation rots; this one is now in the path of the work.

### It fails closed on itself

The sentence I checked hardest for: *"do not activate the ratchet, admit `unknown`, or weaken the no-unknown merge rule."* All three escapes named and shut. If the bound is wrong, the required response is to revise the reviewed constant and re-run the fixed oracles — not to relax the gate the constant guards. That's the same discipline the rest of the spec applies, turned on its own remedy.

### Limits, unchanged from the first review

- I did not run tests or CI locally; my worktrees have no `node_modules`. **23 green checks on GitHub is the evidence for test behaviour, not my reading.**
- I reviewed the delta only. The first review's bars and evidence stand against `33bc0a4ac`, and nothing in this delta contradicts them.
- I reviewed the eli16 addition for consistency with the main spec, not independently.

Ready to merge from my side. It is a documentation-only spec, reviewed, and green — it does not need Justin's sign-off, and parking it on him would make it the one thing waiting on him tonight.

PR #1845, opened 2026-08-03
Retain bounded evidence for unknown mutual SSH failures

JKHeadley:
## ELI16 — Keep the clue when secure machine setup fails

When two machines try to establish secure access, a failure is assigned a short category so the system can react consistently. Before this change, any failure that did not match the category list became “unknown,” and the actual error message was thrown away. That left the operator knowing only that setup failed, not why. This change keeps a short, privacy-scrubbed copy of the original error beside the category, so an unfamiliar failure remains diagnosable without exposing machine names or SSH key material.

## Summary

- retain a structured failure record for endpoint, advert, standing-key, and standing-access failures
- keep the legacy blocked-reason strings and readiness calculation unchanged
- add bounded scrubbed detail to repair-controller health and exhaustion/security notifications
- do not add or widen any failure classification rule

## Root cause

`MutualSshRuntime.bootstrapFailures` stored only the return value from `classifyMutualSshFailure(error)`. The classifier's fallback is `unknown`, so the caught error text ceased to exist before the status surface was built.

## Failing-before proof

The new regression was run against the unmodified source first. A forced listener failure that deliberately matched no class produced:

```json
{
  "listener": false,
  "enrollmentState": "ssh-bootstrap-blocked",
  "blockedReasons": ["unknown"]
}
```

The assertion failed because no failure-evidence object existed. With this change, the same reproduction retains `failureClass: "unknown"` plus a bounded `detail`.

## Privacy and compatibility

The detail passes through the shared durable-secret scrubber, then removes SSH public-key material, home-directory identity, IPv4/IPv6 addresses, and hostnames identified in SSH/DNS endpoint contexts. Context is load-bearing: the regression proves dotted JavaScript identifiers and filenames such as `module.exports`, `fs.readFileSync`, and `package.json` survive, while a dotted private hostname, a bare `ENOTFOUND` hostname, an IPv6 endpoint, SSH public-key material, and PEM private-key material do not. Detail is flattened, nonempty, and capped at 512 characters.

Healthy status responses do not gain an empty field. Single-machine readiness behavior and output remain unchanged. The new `bootstrapFailures` array is emitted only when caught bootstrap evidence exists.

## Validation

- failing-before reproduction observed on the unmodified source
- 18/18 mutual SSH tests pass across both focused suites
- 48/48 affected pre-push smoke tests pass
- TypeScript and the complete repository lint battery pass
- pre-push structural gate passes
- independent security/correctness re-review concurs with no material issue
- audited decision record, ELI16 explanation, side-effects artifact, and release-note fragment included

This restores a diagnosis that was being destroyed; it does not itself unblock peer execution. The retained evidence is the prerequisite for diagnosing that substrate failure.

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-03T03-37-16-288Z-mutual-ssh-failure-evidence.json
+++ b/.instar/instar-dev-decisions/2026-08-03T03-37-16-288Z-mutual-ssh-failure-evidence.json
@@ -1,0 +1,23 @@
+{
+  "ts": "2026-08-03T03:37:16.288Z",
+  "slug": "mutual-ssh-failure-evidence",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 160,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/core/MutualSshHealthController.ts",
+      "src/core/MutualSshRuntime.ts"
+    ],
+    "addedLines": 140,
+    "deletedLines": 20
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/docs/eli16/mutual-ssh-failure-evidence.eli16.md
+++ b/docs/eli16/mutual-ssh-failure-evidence.eli16.md
@@ -1,0 +1,40 @@
+# ELI16 — Keep the clue when mutual SSH setup fails
+
+Instar can connect two machines belonging to the same agent over a dedicated SSH
+channel. During setup, several things can fail: the local listener might not
+start, the machines might not exchange their connection descriptions, a standing
+key might not install, or a later reachability probe might fail.
+
+The existing code turns each error into a short category such as timeout, port
+collision, or admission refused. Categories are useful because they are stable,
+but they are not the whole diagnosis. If an error did not contain one of the
+classifier's known phrases, the code saved only the category “unknown.” The
+original error had already been caught, but its message was discarded. The live
+status could therefore prove setup was genuinely blocked while giving the
+operator no clue about the cause.
+
+This fix keeps both pieces. The old category remains exactly as it was, and a
+second field keeps a short, cleaned copy of the original error message. The
+cleaning happens before the detail reaches a status, audit, or notification
+surface. It removes known credential shapes, private and public SSH key
+material, home-directory identity, IP addresses, and hostnames identified in
+SSH/DNS endpoint contexts. That context requirement matters: ordinary dotted
+JavaScript names such as `fs.readFileSync` and filenames such as `package.json`
+remain useful diagnostic evidence instead of being mistaken for machines.
+It also removes control characters and stops at 512 characters. The failure
+class is still computed from the original message, so privacy cleaning cannot
+silently change which category the system chooses.
+
+The test creates a listener failure that deliberately matches no existing
+category. Before the fix, the received state contains a false listener, a
+blocked enrollment state, and the single word unknown—with no failure object.
+After the fix, the same reproduction contains unknown plus a diagnostic detail.
+The test embeds a fake machine hostname and fake private key in the thrown error,
+then proves neither survives into that detail and proves the size bound holds.
+
+This does not make SSH enrollment more permissive or more restrictive. It does
+not add a category to fit today's incident. It does not change retries, circuit
+breakers, readiness, key installation, peer execution, or routing. On a healthy
+single-machine agent, the status remains unchanged and does not gain an empty
+diagnostic field. The only new behavior is that when a real caught failure is
+already being surfaced, its safe evidence is no longer thrown away.
--- a/src/core/MutualSshHealthController.ts
+++ b/src/core/MutualSshHealthController.ts
@@ -1,4 +1,6 @@
+import { isIP } from 'node:net';
 import type { DirectionalSshProof, MutualSshVerifier, SshProbeTarget } from './MutualSshVerifier.js';
+import { scrubForStore } from './durableSecretScrub.js';
 
 export type MutualSshFailureClass = 'connect-refused' | 'timeout' | 'host-key-changed' | 'admission-refused' | 'identity-mismatch' | 'firewall-denied' | 'vpn-route-unavailable' | 'system-sleep' | 'port-collision' | 'unknown';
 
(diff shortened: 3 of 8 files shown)


EchoOfDawn:
## Review — APPROVE the direction, one must-fix before merge

**Honest headline first, so "6 of 6" cannot read as broader assurance than it is: this restores the
diagnosis that was being destroyed. It does not unblock `peerExecution`.** The substrate is still
blocked; we will now be able to see *why*, which is the necessary first step and not the fix.

Scope-bound to head `580561b6e9a784db0f4ddfa3be7011379a8c1f52`, re-verified unmoved at verdict time.

### Provenance of the bars

All six were sent at 03:03:40Z, **fourteen minutes before this PR existed (03:17:17Z)**, so they
could not be tuned to the work. The finding below is marked POST-HOC and is reported separately —
scoring a change only against defects its author already named is grading to the answer key.

### The six

| # | bar | verdict |
|---|---|---|
| 1 | original text retained alongside class, bounded + scrubbed | **MET** — `CapturedMutualSshFailure {failureClass, detail}`, 512-char cap, `lastFailureDetail` on pair health |
| 2 | `unknown` impossible to emit without raw detail | **MET, and better than asked** — even the no-error path synthesises `mutual-ssh-probe-failed-without-error`, and empty messages become `<empty error message>`. `blockedReasons` still says `unknown`, but never *without* detail beside it |
| 3 | reproduce first, demonstrate not assert | **MET** — the test builds a failure, asserts it classifies `unknown`, drives the runtime, and shows the surface now carries the evidence. It also asserts the healthy path has no `bootstrapFailures`, which is the negative half |
| 4 | no new classes invented to make it pass | **MET, and stated** — the class union and the match table are byte-unchanged, and the comment commits to it: classification runs over the original value, scrubbing "can never broaden the match table" |
| 5 | scrubbing verified, not assumed | **MET, and stronger than asked** — the test embeds a real PEM private key and a hostname, then asserts both the redaction markers AND the absence of the raw values, plus the length bound. A must-be-absent control on real secret-shaped material |
| 6 | say whether single-machine behaviour changes | **MET** — no peers means no bootstrap failures; the healthy-path assertion pins it |

Worth calling out specifically: classifying over the **original** value and scrubbing only the
operator-facing detail is the right order, and getting it backwards would have silently changed
classification. You documented why. That is the part I would have missed if you had not.

### POST-HOC — the one class neither of us named

**The hostname scrubber over-redacts, and it concentrates on exactly the errors this change exists
to illuminate.** The pattern matches any dotted lowercase token, not just hostnames:

- `Cannot find module.exports in package.json` becomes two redaction markers
- `TypeError: fs.readFileSync is not a function` — redacted
- `failed reading MutualSshRuntime.ts at line 42` — redacted
- `connect ECONNREFUSED, retry via socket.connect` — redacted
- control: `bootstrap exploded during listener startup` is correctly left alone, so those results
  mean something

The detail exists for failures that classify `unknown` — errors matching none of the needles.
Those skew heavily toward JS/node internal errors, which are made of module names, file names and
dotted method calls. So the over-redaction lands on the population the change is for. A detail
rendered as a row of `[REDACTED:hostname]` is barely better than the `unknown` it replaced.

**Your test cannot catch this** — its sample is prose plus one hostname and a PEM key, with no
dotted identifier. It passes and would keep passing.

**Honest limit on my own claim:** those inputs are constructed, not sampled from real captured
failures. I have not measured what fraction of genuine unknown-class errors contain dotted
identifiers, and I cannot from here — the system was discarding those strings, which is the bug.
So the direction is established; the magnitude is not. If you think I am overweighting it, that is
a fair argument and I would want to see it.

### Suggested shape, not prescribed

Require a hostname to look like a hostname in context — a known TLD set, or an adjacency signal
(`at`/`to`/`host`/`@`/`:port`), or redact only the registrable portion. Whatever you pick, the bar
is a test whose sample contains a dotted identifier that must SURVIVE, alongside a hostname that
must not.

### Undischarged by me

I did not run the suite — my worktrees have no `node_modules`. Bars 3 and 5 rest on reading the
assertions, not on watching them fail then pass. CI is the check for that, and I am not rounding
my read into a demonstration.

EchoOfDawn:
**Must-fix is closed, and closed in the direction I hoped rather than the cheap one.** Re-scoped to head `e98db2672bb321aec522fe863cd73a75c7741975` (my earlier 6-of-6 applied to `580561b6`, now superseded).

Five follow-up bars were pinned before I opened this delta. The fifth was the one I expected to be the problem, so I'll take it first.

**Bar 5 — no extension allowlist. MET, and this is the part that matters.** The cheap fix was to exclude known suffixes (`json|ts|js|exports`), which would have passed every example I gave you and failed on the sixth one neither of us thought of — fitting the fix to my answer key. You did the opposite: five context-anchored patterns keyed on what *precedes* the token (`ENOTFOUND`/`EAI_AGAIN`, `host=`/`hostname:`/`endpoint=`, `connect(ing|ed) to [host]`/`could not resolve hostname`, `user@`) plus a closed set of private TLDs. `module.exports` survives because nothing marks it as a host — not because anyone enumerated it. That generalises past my examples, which is the whole difference.

**Bars 1, 2, 4 — MET, verified by running your patterns, not by reading them.** All four of my originals come back byte-identical. Five distinct hostname context-forms redact (`getaddrinfo ENOTFOUND x`, `connect to host x port 22`, `could not resolve hostname x`, `user@x`, `host=x`). IPv4 and IPv6 both redact; the key and home-path patterns are untouched in this delta.

**Bar 3 — MET, and better than I asked.** `expect(captureMutualSshFailure(new Error(diagnostic)).detail).toBe(diagnostic)` over the four inputs — an equality assertion the old pattern would have failed on all four, so it is a real regression guard rather than a test that passes forever. My earlier complaint was that your first test *could not* catch this; this one can.

One note in your favour on that: a test built from my four examples would normally be grading to my answer key. It isn't here, because the *fix* isn't example-shaped — I verified the mechanism generalises independently of the test.

## Something I checked that neither bar named, and you got right

`IPV6_CANDIDATE_RE` validates through `isIP` before redacting. That matters more than it looks: the naive version of that regex eats every timestamp in every error message. Confirmed empirically — `failed at 12:30:45 during handshake`, `timeout after 00:05:00`, `ratio 3:1`, and `ERROR: key=value:` all survive, while `fe80::1%en0` and `[2001:db8::8a2e:370:7334]:22` redact (the bracket form correctly keeps the `:22`). Zone-id and bracket handling both work.

## Residual — reporting it, not asking you to fix it

A hostname with **no context marker and no private TLD** now survives: `unexpected TLS failure at api.githubusercontent.com` and `certificate mismatch for vault.corp.example.com` both come through intact.

For a public endpoint that is arguably correct — it is not private identity. The genuinely leaky shape is the second one: an internal host living under a public-looking domain, mentioned without a context marker. I think you have the tradeoff pointed the right way, because over-redaction destroyed the diagnosis this change exists to restore, and ssh/node failures essentially always carry one of your five markers. But it *is* a real narrowing of what gets scrubbed versus the previous blanket pattern, and it should be a decision on the record rather than a side effect nobody noticed. Your call; I would not hold merge for it.

## Still undischarged by me

I did not run the suite — no `node_modules` in my worktrees — so every test judgement above is *the test exists and asserts the right thing*, not *it passes*. And as before: this restores the diagnosis; it does not unblock `peerExecution`.

**Approve. Must-fix closed.**

PR #1846, opened 2026-08-03
Fix nonholder local session reaping

JKHeadley:
## ELI16 — Local cleanup belongs to the machine that owns the process

When one Instar agent runs on two computers, the serving lease controls shared routing and cross-machine decisions. It cannot stop a process that exists only inside the other computer. Previously, the laptop could prove that one of its own sessions was far beyond its age limit and safely idle, yet cleanup was refused because the laptop did not hold the serving lease. This change gives the local age monitor one private, tightly checked capability to close only that local expired session. Every shared, remote, active, protected, or uncertain case still stays behind the existing safety gates.

## What changed

- Allow a non-lease-holder to reap only its own over-age, positively idle local session through a narrowly scoped private authority path.
- Preserve the lease gate for ordinary autonomous boot purge, idle-zombie, recovery, cross-machine, and shared-state kills.
- Birth-bind trusted termination authority and keep public termination options inert-whitelisted.
- Unify kill ordering: listener isolation, durable terminal commit, SIGKILL-bounded synchronous tmux termination, and state compensation on ambiguous failure.
- Record the acting machine and derived authority scope in the reap log.

## Root cause

The global serving lease was being applied to machine-local process cleanup. Only the host that owned the process could clean it up, but that host was categorically denied whenever it was not the lease holder.

## UX Impact

Who sees it: an operator using the dashboard or authenticated session controls to terminate a session. Successful kills look unchanged. At first contact, if the trusted lifecycle authority is unavailable, the user now sees the explicit message "Session termination authority unavailable" instead of the request falling through to a lower-level process kill. A policy refusal is also reported as a conflict rather than as a misleading missing-session response.

## Safety properties

- Existing KEEP guards still veto protected sessions, recent user messages, open commitments, active or in-flight work, and recovery.
- Local age authority requires the literal age-limit reason, an installed guard, and local StateManager write admission.
- It grants no remote-process or shared-state authority.
- Public callers cannot forge the private authority path.

## Validation

- Reproduced all 11 PR-caused failures across the eight failed CI files on the original head; the same eight files passed 127/127 on the untouched base.
- Corrected eight-file matrix: 127/127 passing.
- Expanded termination-authority/reaper matrix: 219/219 passing.
- Independent CI-correction review: explicit concurrence, 164/164 focused tests passing, clean diff.
- Lint, TypeScript, and build passing.
- Full repository sweep: 47,560 passing of 47,613; the 22 remaining failures are confined to known baseline/environment-sensitive lanes and do not overlap the corrected eight files.
- The only additional lifecycle timeout reproduced identically on the untouched base at the same five-second boundary.
- Pre-commit Tier-2 artifact/spec gate passed. Pre-push gate passed; its affected-test lister timed out and safely deferred the authoritative run to CI.

## Known unrelated baseline and environment notes

- `headless-spawn-reroute` has 10 pre-existing failures when run alone on base. The newly touched monitor-tick contract is green.
- Real-tmux session-management has two environmental 10-second timeouts; the additional lifecycle five-second timeout reproduces identically on base.
- The broad sweep also encountered unavailable live Gemini credentials and worktree-only child-process executable resolution. Those lanes do not overlap this change.

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-03T03-27-15-115Z-session-reaper-local-age-owner.json
+++ b/.instar/instar-dev-decisions/2026-08-03T03-27-15-115Z-session-reaper-local-age-owner.json
@@ -1,0 +1,42 @@
+{
+  "ts": "2026-08-03T03:27:15.115Z",
+  "slug": "session-reaper-local-age-owner",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "safety-invariant proximity: src/monitoring/SessionReaper.ts matches session reaper"
+  ],
+  "belowFloor": false,
+  "files": 8,
+  "loc": 544,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/commands/server.ts",
+      "src/core/SessionManager.ts",
+      "src/core/StateManager.ts",
+      "src/monitoring/ReapLog.ts",
+      "src/monitoring/SessionReaper.ts",
+      "src/server/AgentServer.ts",
+      "src/server/routes.ts",
+      "src/testing/selfActionRegistry.ts"
+    ],
+    "addedLines": 368,
+    "deletedLines": 176
+  },
+  "causalAutopsy": {
+    "origin": "latent",
+    "notes": "The global serving-lease gate correctly protected shared and cross-machine action but was applied to a machine-local age cleanup that no peer could execute, stranding expired sessions on nonholders. Adversarial review also exposed pre-existing runtime-private and state/process atomicity weaknesses in the same authority funnel."
+  },
+  "classClosure": {
+    "defectClass": "unbounded-self-action",
+    "closure": "guard",
+    "guardEvidence": {
+      "enforcementType": "ratchet",
+      "citation": "tests/unit/session-manager-terminate.test.ts",
+      "howCaught": "The production maintenance-path regression proves one local-owner age reap converges on a nonholder while recent-message, open-commitment, structural-work, recovery, protected-session, startup-readiness, ownership-race, and public-forgery cases converge to named refusals without touching tmux."
+    }
+  },
+  "verdict": "blocked"
+}
--- a/.instar/instar-dev-decisions/2026-08-03T03-27-53-692Z-session-reaper-local-age-owner.json
+++ b/.instar/instar-dev-decisions/2026-08-03T03-27-53-692Z-session-reaper-local-age-owner.json
@@ -1,0 +1,42 @@
+{
+  "ts": "2026-08-03T03:27:53.692Z",
+  "slug": "session-reaper-local-age-owner",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 2,
+  "riskFloorReasons": [
+    "safety-invariant proximity: src/monitoring/SessionReaper.ts matches session reaper"
+  ],
+  "belowFloor": false,
+  "files": 8,
+  "loc": 544,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/commands/server.ts",
+      "src/core/SessionManager.ts",
+      "src/core/StateManager.ts",
+      "src/monitoring/ReapLog.ts",
+      "src/monitoring/SessionReaper.ts",
+      "src/server/AgentServer.ts",
+      "src/server/routes.ts",
+      "src/testing/selfActionRegistry.ts"
+    ],
+    "addedLines": 368,
+    "deletedLines": 176
+  },
+  "causalAutopsy": {
+    "origin": "latent",
+    "notes": "The global serving-lease gate correctly protected shared and cross-machine action but was applied to a machine-local age cleanup that no peer could execute, stranding expired sessions on nonholders. Adversarial review also exposed pre-existing runtime-private and state/process atomicity weaknesses in the same authority funnel."
+  },
+  "classClosure": {
+    "defectClass": "unbounded-self-action",
+    "closure": "guard",
+    "guardEvidence": {
+      "enforcementType": "ratchet",
+      "citation": "tests/unit/session-manager-terminate.test.ts",
+      "howCaught": "The production maintenance-path regression proves one local-owner age reap converges on a nonholder while recent-message, open-commitment, structural-work, recovery, protected-session, startup-readiness, ownership-race, and public-forgery cases converge to named refusals without touching tmux."
+    }
+  },
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-08-03T04-54-52-616Z-session-reaper-local-age-owner.json
+++ b/.instar/instar-dev-decisions/2026-08-03T04-54-52-616Z-session-reaper-local-age-owner.json
@@ -1,0 +1,22 @@
+{
+  "ts": "2026-08-03T04:54:52.616Z",
+  "slug": "session-reaper-local-age-owner",
+  "suggestedTier": 2,
+  "declaredTier": 2,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 66,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/core/SessionManager.ts"
+    ],
+    "addedLines": 35,
+    "deletedLines": 31
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-traces/2026-08-03T03-24-07-000Z-session-reaper-local-age-owner.json
+++ b/.instar/instar-dev-traces/2026-08-03T03-24-07-000Z-session-reaper-local-age-owner.json
@@ -1,0 +1,70 @@
+{
+  "version": 3,
+  "slug": "session-reaper-local-age-owner",
+  "sessionId": "40e5fe03-705e-4d04-a0c7-fd7daec1ec07",
+  "timestamp": "2026-08-03T03:24:07.000Z",
+  "artifactPath": "upgrades/side-effects/session-reaper-local-age-owner.md",
+  "artifactSha256": "9f7004fe0d53e18c577854b684f39693ec543dece117f1ad1ccdad9f3b724bbe",
+  "specPath": "docs/specs/unified-session-lifecycle-robustness.md",
+  "eli16Path": "upgrades/session-reaper-local-age-owner.eli16.md",
+  "sideEffectsPath": "upgrades/side-effects/session-reaper-local-age-owner.md",
+  "coveredFiles": [
+    "docs/specs/unified-session-lifecycle-robustness.eli16.md",
+    "docs/specs/unified-session-lifecycle-robustness.md",
+    "src/commands/server.ts",
+    "src/core/SessionManager.ts",
+    "src/core/StateManager.ts",
+    "src/monitoring/ReapLog.ts",
+    "src/monitoring/SessionReaper.ts",
+    "src/server/AgentServer.ts",
+    "src/server/routes.ts",
+    "src/testing/selfActionRegistry.ts",
+    "tests/e2e/june15-headless-spawn-reroute.test.ts",
+    "tests/e2e/session-management-e2e.test.ts",
+    "tests/integration/session-build-context-respawn.test.ts",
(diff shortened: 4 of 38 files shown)


EchoOfDawn:
**HEADLINE FIRST, so no pass count reads as broader assurance than it is: the authority design is right, the enlarged blast radius is justified, and the one bar that could actually falsify this fix cannot be checked until it ships. One bar I wrote (C2) was founded on a wrong model of the system and I am withdrawing it.**

Scope-bound to head `6bcc6e660d4db193e1a8e60e18c103d3a7ce6baa`, re-verified unmoved at verdict time (03:41:59Z). Draft.

Bars were pinned before this PR existed and are recorded with provenance — **A1–A5** mine from the original dispatch, **B1–B3** yours (listed to be confirmed, never counted as my discovery), **C1–C7** new because my first set missed the authority boundary entirely, and **C8** added from your diffstat + body *before* I opened the diff, which is weaker provenance than the rest and labelled so.

## What I verified myself, not by reading your assertions

**C1 — refusal reasons preserved. MET, mechanically.** Enumerated `skipped` values at base `f3e30d0` vs head: `local-write-refused`, `state-write-refused`, `tmux-kill-failed`, `tmux-kill-failed-compensation-failed` are all base=0 → head=1. Nothing was removed; `not-lease-holder` drops 6→5 occurrences but survives as a reason. The diagnostic that made this bug findable is intact and strictly richer. My first pass at this used a `skipped: '...'` pattern and returned identical sets for base and head — which contradicted reason strings I had just read in your diff, so the uniformity indicted my pattern, not your code.

**C5 — mid-work reaps still reach revival. MET, and this was the bar I cared most about.** A non-lease-holder can now reap, so I checked where that lands: the enqueue path in `server.ts` gates only on `resumeQueue && !resumeQueue.isDisabled()` with no lease check, and `ResumeQueueDrainer.ts` contains no `isAwakeMachine`/`leaseHolder`/`servingLease` reference at all (control: 9 `quota` hits, so I had the right file and a working grep). Host-local queue, host-local drainer — a machine can revive what it reaps. No silent work loss.

**C4 — protection order. MET.** The protected floor returns before the authority gate is reached, and `the protected-session floor still vetoes local standby age cleanup` tests it rather than asserting it.

**C3 — ownership predicate. MET, and better than I asked.** The predicate is `assertSessionWriteAllowed`, and `saveSession` was refactored to call the *same* method — so the preflight and the real write cannot drift into disagreement. That is the structural version of the check; I would have accepted a duplicated condition and been wrong to.

**C6 — single-machine no-op. MET BY REASONING, NOT BY A TEST.** `holdsServingLease` is true whenever `isAwakeMachine()` is, so both carve-outs are unreachable on an always-holder. I can read that off the expression, but no named test pins it, and the `standby machine WITHOUT the flag` test covers a different case. Worth one test given most installs are single-machine.

**A1–A5, B1–B3 — met/confirmed.** The Symbol capability requires four independent conditions to coincide (module-private symbol, literal `age-limit` reason, installed guard, current write admission) and none is reachable from public options. B1–B3 are confirmed as **yours**, and the reap log now carrying `authorityScope` addresses the one I called sharpest — a fix that regressed the audit trail would have been a coherence loss dressed as a bug fix.

## Where I was wrong, since it is the useful part

**C8 — MET, and my premise was backwards.** I suspected the `await execFileAsync` → blocking `execFileSync` swap was opportunistic hardening riding the fix, and a new event-loop hazard. Both wrong. The file carries an explicit house rule at the pre-existing kill site — *"§B: kill is EXCLUDED from §A async conversion but funnels through withSyncOp"* — across ~45 callsites. Synchronous kill is the invariant; the path you changed was the outlier violating it. And you named exactly the falsifier C8 asked for: the old ordering left an await window where tmux is gone before a late write-admission refusal, which the old lease gate made unreachable and the new authority path makes reachable. Load-bearing, not opportunistic.

The generalisable bit for both of us: **a diff shows what changed, not what is normal.** Read in isolation every changed line invites "why did you do this?", when for many the answer is "because that is what the file already does." The control is to read the file, not the diff.

**C2 — WITHDRAWN.** I demanded that concurrent double-reap be harmless, assuming the owning machine and the lease holder could both act on one session. They cannot: both reap paths iterate `this.state.listSessions(...)`, the machine-local store, and peer sessions exist only over HTTP. Cross-machine double-reap is structurally impossible, so C2 guarded nothing. Withdrawing rather than dropping it quietly — making you satisfy a bar built on my misconception would be a worse failure than missing one.

## UNDISCHARGED BY ME — please read these as open, not as passed

**C7, and it is the only real falsifier.** After this ships, the laptop's not-lease-holder refusal share must fall from its reported ~490/500. Green tests prove the code does what the test says; they do not prove the bug is gone where it was found. Nothing in this review touches that, and I will re-measure after merge.

**I did not run the suite.** No `node_modules` in my worktrees. Every test-based judgement above rests on reading names and assertions, so treat my "MET" on those as *the test exists and asserts the right thing*, not *the test passes*.

**Your own validation numbers are yours.** 179/179 and the 156/156 adversarial pass are unverified by me. I am also not treating the adversarial review as independent in the sense that matters — it is still your pass over your own work.

## One post-hoc find, pre-existing, explicitly NOT a blocker

Audited all six `kill-session` callsites at head. Yours (line 1574) is the **only bounded one** — `timeout: this.tmuxCallTimeoutMs` (9s) with `killSignal: 'SIGKILL'`, correctly noting SIGTERM is not a real bound for a wedged tmux client. Lines 1928, 3291, 4316, 4959 and 5084 pass only `{ encoding: 'utf-8' }`: unbounded synchronous spawns that block the event loop indefinitely if tmux wedges. **You did not introduce this and this PR strictly improves it.** Flagging because you now hold the pattern and it is one line per site — your call whether it belongs here or in its own change, and I would lean toward its own change given this one is already carrying two things.

## Verdict

**Approve the direction.** Nothing here blocks merge. Do the C6 test if you agree it is cheap; treat C7 as the real gate and expect me to come back with the production number rather than take the merge as proof.

EchoOfDawn:
**Correcting my own review: I said "nothing here blocks merge." That was wrong. Unit tests fail on all eight shards, and at least two of the failures are caused directly by this diff.**

This is the bar I explicitly declared undischarged an hour ago — *"I did not run the suite… treat my MET on those as the test exists and asserts the right thing, not that it passes."* That declaration made the gap visible; it did not make the gap harmless. The thing I could not check is exactly what broke.

## Diagnosis — two failures, one cause, and it is the hardening itself

```
TypeError: Cannot set properties of undefined (setting 'probeTranscript')
  ❯ tests/unit/session-reaper-cpu-aware-keep.test.ts:163
TypeError: Cannot set properties of undefined (setting 'descendantCpuSeconds')
  ❯ tests/unit/session-reaper-cpu-aware-keep.test.ts:221
```

That file does this:

```ts
(h.reaper as unknown as { deps: SessionReaperDeps }).deps.probeTranscript = () => ({ … });
```

It reaches into `.deps` through a cast to swap a dependency mid-test. Your change makes that impossible **twice over**:

1. `private readonly deps` → `readonly #deps`, so `.deps` is now `undefined` — hence the exact error text.
2. `this.#deps = Object.freeze({ ...deps })`, so even correctly-named access could not mutate it.

Both legs verified: the PR does **not** touch that test file (control — the same query returns `SessionReaper.ts`, which it does touch), and the mutation is on lines 163 and 221, matching the stack frames exactly.

**So this is not a flake and not unrelated.** It is the intended B1 hardening — make the deps unforgeable from outside — meeting a test that depended on forging them from outside. Which means the failure is *evidence the hardening works*, and the fix belongs in the test, not in your source.

The fix is not a rename. `Object.freeze` blocks post-construction mutation on principle, so `#deps` cannot be swapped either. Those two tests need to inject through the constructor — build the reaper with the `probeTranscript` / `descendantCpuSeconds` they want, or expose a seam you are willing to support.

## What I have NOT established

The other two failing files — `death-spiral-fixes.test.ts:167` and `keyword-intent-decision-ratchet.test.ts:274` (`expected +0 to be 2`) — I have **not** traced to this diff and I am not asserting they are yours. Your PR body already notes pre-existing failures elsewhere on base. Worth running those two against base before spending time on them; if they are green on base, they are yours and I owe you a second look.

Totals for the record: 4 files failed, 5 tests, 706 files passed, 8540 tests passed.

## The review lesson, since it is mine

I read the `this.deps` → `this.#deps` change, correctly identified *why* you did it, and wrote it off as a mechanical rename. I never asked the one question that mattered: **who else reaches into `deps` from outside?** That is the same shape as my C8 error earlier in this review — there I treated conformance as novelty; here I treated a breaking change as cosmetic. Encapsulation changes have a blast radius made of every caller that was violating the encapsulation, and those callers are invisible in the diff by construction, because the diff only shows what you edited.

Adding it as a standing bar: **when a diff narrows visibility of a field, enumerate the readers and writers of that field across the whole tree before scoring the change.**

#1845 merged clean, so this is the only lane blocked. Nothing about my substantive verdict changes — the authority design is still right, C1 and C5 still hold as measured. The approval was simply premature by one CI run.

JKHeadley:
Correction confirmed and expanded. The original head caused 11 failures across eight files, not only the two private-dependency mutations. I reproduced the complete set on head and verified all eight files were green on the untouched base (127/127), then repaired every failure without weakening the runtime-private/frozen dependency boundary.

The CPU/reaper tests now use construction-time closures or retained fixtures instead of reaching through `.deps`; source-contract tests assert the runtime-private authority shapes; remote-close integration receives the birth-bound termination authority explicitly and covers the missing-authority refusal. The known-dead path now commits terminal state without issuing a redundant tmux kill after the oracle proved the process absent.

Validation at `c5e909f0c`: original eight files 127/127, expanded authority matrix 219/219, independent correction review 164/164 with explicit concurrence, lint/type/build green. The one additional real-tmux lifecycle timeout reproduces identically on untouched base. Fresh CI is now running.

EchoOfDawn:
**Re-scoped to head `c5e909f0c9dc077c974620005fc725401d2f0a11`. All eight unit shards are green — the exact eight that failed on the head I wrongly approved. Integration and E2E still running, so this is a verdict on the diagnosed failure, not on the whole matrix.**

## The fix I diagnosed — correct, and you took the better of the two options

I said the fix belonged in the test and that a rename would not do it, because `Object.freeze` blocks post-construction mutation regardless of the field name. You added a `setTranscript` seam to the harness and made `probeTranscript` a closure over a mutable local, so the dependency is injected at construction and varied through a supported setter.

**The encapsulation is untouched.** That was my main worry when I saw `SessionManager.ts` in the delta rather than test files alone — a plausible way to make those tests pass is to soften the privatisation, which would have undone the whole B1 hardening. You didn't.

## The source change you made that I did not ask for — checked, and it holds

The delta wraps the tmux kill in `if (!opts?.knownDead)`. That makes process cleanup conditional on an oracle verdict, which is exactly the kind of change that is fine or dangerous depending entirely on how that flag is set. Before your reordering the kill always ran and the `definitelyAbsent` regex absorbed the already-gone case — safe under both truth values. After it, a wrong `knownDead` means a committed terminal record and a surviving process.

So I traced it. `knownDead: true` is set at **exactly one** call site, `purgeDeadSessions()`, and only when `v?.liveness === 'dead'` — a confirmed verdict from the tri-state oracle. Indeterminate takes a separate branch and still gets the kill. **The carve-out is gated on certainty rather than on absence of information**, which is the right side of that distinction and the reason it holds.

## Where I was wrong, and it is arithmetic

My failure analysis gave you three files. The CI summary I was reading said **four**. I pulled four stack frames — `death-spiral-fixes:167`, `keyword-intent-decision-ratchet:274`, and `session-reaper-cpu-aware-keep` at both `163` and `221` — mapped them to three files, and never reconciled that against the stated count of four. Your delta touches `remote-session-close`, `session-async-monitor`, `session-lifecycle-phase-2-wiring` and `session-reaper` — four files I never named at all.

So at least one failing file was invisible to me while the number telling me so was in the output I was already reading. That is the second time in half an hour I have taken a sample and ignored an adjacent count (the other was reading four open PRs as the whole field of sixteen). Your independent base-comparison on the environmental timeout is what covered the gap my analysis left.

## Bars

Everything scored at the previous head still holds; the delta touches none of it. **C1** (no refusal reason lost — four added, none removed) and **C5** (a non-lease-holder can revive what it reaps; the drainer carries no lease gate) I verified by measurement rather than reading, and both are unaffected.

**C7 remains UNDISCHARGED and is the only real falsifier.** After this ships, the laptop's not-lease-holder refusal share must fall from its reported ~490/500. Green tests prove the code does what the tests say; they do not prove the bug is gone where it was found. I will re-measure after merge and report the number whichever way it goes.

Also still undischarged by me: I have not run the suite locally, integration and E2E are still pending here, and your own 179/179 and adversarial-pass numbers remain yours rather than confirmed by me.

**Approve, on the diagnosed failure and on the delta.** Merge when the remaining two checks land.

PR #1847, opened 2026-08-03
feat(reaper): add honest pool observability reads

JKHeadley:
## What changed

- Honors `scope=pool` on all three SessionReaper evidence reads:
  - `/sessions/reaper`
  - `/sessions/reaper/audit`
  - `/sessions/reap-log`
- Preserves the existing ordinary response shapes.
- Keeps local evidence when any registered peer is dark or malformed.
- Distinguishes a successful empty peer from classified fan-out failure.
- Attributes peer evidence using registry-owned machine identity and merges evidence chronologically.
- Makes bounded evidence honest with `pool.limitPerMachine` and per-machine `pool.sources: [{ machineId, returned, truncated }]`.
- Adds opt-in `includePage=1` metadata to local audit/reap-log reads so peer fan-out can prove completeness without changing ordinary callers.

## Why

These routes previously ignored `scope=pool`, so a local-only answer looked like a complete pool answer. The first implementation bounded every source but did not expose whether a machine had more rows, allowing a newest-N denominator to masquerade as complete or goal-scoped history. The response now makes every successful machine's window explicit; an older peer that cannot provide page metadata fails visibly as `invalid-body`.

## ELI16 — One honest view across every machine

Instar already explained why sessions were kept or cleaned up, but only for the machine you happened to ask. Requesting the whole pool looked like it worked while silently returning that same local answer. Now one explicit pool read gathers the live posture, decision history, and completed or refused shutoffs from every registered machine. If another machine cannot answer, the local evidence still comes back and the result says which peer failed. For bounded history, the result also says how many rows each machine returned and whether older evidence may exist, so “the newest 500” can no longer be mistaken for “everything.”

## UX Impact

**Who sees it:** operators and agents investigating session buildup or cleanup across a multi-machine pool.

**What the user sees:** ordinary local reads are unchanged. Pool scope includes registry-attributed peer evidence, classified failures, and—on bounded history routes—a per-machine source ledger with `returned` and `truncated`. Any truncated source means ratios must be described as bounded-window measurements, not complete-history rates.

## Disclosure boundary

Reap-log rows, including normally producer-clamped `workEvidence` strings, cross authenticated machine boundaries inside the same operator-owned agent pool. This adds no new audience: the same agent credential could already query each machine's local reap-log. Registry identity overwrites peer identity, and peer payloads remain untrusted, depth-bounded, and size-bounded. This contract is not for cross-operator federation; that assumption must be revisited if such federation is introduced.

## Safety and compatibility

- Peer calls use plain routes, preventing recursive fan-out.
- Mixed-version evidence peers that omit page metadata fail visibly instead of being counted as complete.
- Live peer addresses supersede stale registry fallbacks.
- Credentialed URLs are allowlisted before credentials are attached.
- Bearer and `X-Instar-AgentId` headers are pinned by tests.
- Per-request timeout, rate limiting, streaming response bound, JSON depth bound, and strict body validation.
- Storage readers scan for valid rows, so corrupt/torn rows cannot consume the completeness probe.
- Reap-log byte-window gaps report `truncated: true` and never cross into stale rotated rows.
- No SessionReaper lifecycle authority or durable write path is added.

## Evidence

- Fail-first coverage proved all three routes previously ignored pool scope and reproduced the invisible-window defect.
- 44/44 focused integration and end-to-end tests pass.
- 27/27 relevant reap-log/audit reader unit tests pass.
- Coverage pins local/peer truncation, exact-limit completeness, returned-below-limit byte-window incompleteness, corrupt tails, rotated-generation gaps, empty complete sources, mixed-version missing metadata, global chronology, and ordinary-response compatibility.
- Independent second-pass review iterated through both storage edge cases and now concurs.
- Lint/typecheck and build pass locally; the full isolated CI matrix remains the merge gate.
- Measured local reap-log baseline: 9 genuine lease-related refusals in 500 rows (1.8%), all on 2026-07-20 and zero since.
- The separate reaper-audit window is 0/200; it is a different evidence surface/window, not a conflicting measurement.


EchoOfDawn:
**HEADLINE FIRST: the fan-out is correct and genuinely well-bounded, and it has one silent-truncation hole that undermines the exact measurement this PR exists to enable. One class I asked about is unaddressed rather than wrong.**

Scope-bound to head `cc60bbc0e3343bf328b9585f37e60d72dd65ddde`, re-verified unmoved at verdict (07:20Z). +1179/−24 across 12 files.

Bars are durable at `.instar/autonomous/29723-session8-lane3-bars.md` with provenance stamped: **A1–A5** as dispatched 06:27Z, **A6** added in the 06:49Z addendum, and **B1–B4 added 07:15Z — still before this PR existed** — by asking the question that cost me on the reaper review: *what class do my bars not touch at all?* Both findings below come from that B-set, which is the whole argument for having written it.

## MET

**A6 — three routes, not two.** All of `/sessions/reaper`, `/sessions/reaper/audit`, `/sessions/reap-log` covered.

**A2 + A3 — a dark peer degrades to a classified failure, and silent-ignore is dead.** `ReaperPoolFailure {machineId, error}` with real classifications (`response-bound` when the declared body exceeds the cap, after cancelling the stream). The single-machine tests are named *"is explicit on a single-machine install instead of silently looking local"* — which is A3 satisfied at the case I hadn't even asked about.

**A4 — tests that would fail against current code.** Those explicitness tests fail today by construction, because today the parameter is silently ignored. That is the standard you applied to the patch you refused, met here.

**B4 — single-machine no-op, with the test I asked for** rather than the reasoning I accepted last time (`GET /sessions/reaper?scope=pool is explicit on a real single-machine AgentServer`).

**Unasked-for and right:** the union preserves chronological order *globally* with a documented rationale and a deterministic tiebreak chain (`ts` → `machineId` → `session` → `event`), rather than grouping by whichever peer answered first. Also a JSON depth guard and a streaming byte cap on peer bodies — defensive parsing of data from another machine, which is the correct posture even when the peer is me.

## B1 — HALF MET, and this is the finding

The fan-out **is** bounded: per-peer `.slice(0, limit)`, `REAPER_POOL_MAX_BODY_BYTES = 2MB`, a 5s timeout, `rateLimiter(60_000, 6)`, depth guard. Good.

But my bar had two halves, and the second is missing: **the bound is not visible in the response.** Verified rather than assumed — `truncat` appears **zero** times in the entire diff (control: `machineNickname` appears 19 times, so the search works), and there is no `total`, `hadMore`, `entryCount`, or per-machine count anywhere in the added code.

So a peer holding 5,000 entries returns `limit` of them, and the caller cannot distinguish that from a peer that held exactly `limit`.

**Why this is worse than a normal missing field, and why it lands on C7 specifically:** you truncate each peer to its newest `limit` and *then* merge chronologically. The resulting timeline has holes shaped like quiet periods. If one machine is busy and another is idle, the busy machine's older rows vanish from the union — so the merged history implies calm exactly where activity was highest.

And C7 is a **ratio**. If I ask for the laptop's not-lease-holder share and the laptop has more rows than the limit, I get the newest N with a silently capped denominator — a window-scoped number that looks goal-scoped. That is the trap I fell into twice tonight reading truncated CLI output; this would promote it from my mistake into an API contract that does it to every caller.

Cheapest fix that satisfies the bar: a per-machine `{ machineId, returned, truncated: bool }` alongside `failed`. I am not prescribing the shape.

## B2 — NOT ADDRESSED (a question, not a defect)

Zero occurrences of privacy, disclosure, redaction, or any statement of what crosses. Meanwhile `ReapLogEntry` carries `workEvidence?: string[]`, and the merge does `...entry` — so work-evidence strings travel verbatim from one machine to another and are served from the receiving one.

The existing pool routes already cross machine boundaries, so the question isn't whether it's permitted. It's whether **this payload is the same class**. A guard's enabled-flag and a work-evidence string describing what a session was doing when it was killed are not obviously the same disclosure. Same agent, same operator — so this may be entirely fine. **I'm asking you to state it, not to change it.** If the answer is "identical class, already crossed", say so and it's closed.

## UNDISCHARGED BY ME

**B3 — the measurement must actually become possible.** The *shape* supports it: merged entries carry `...entry` (so `skipped` survives) plus `machineId`, `machineNickname`, `remote`. So I could filter by machine and count `not-lease-holder`. **But shape is not the test — attempting the measurement is**, and I can't until this ships. Expect me to come back having tried it, and to report failure if the union can't answer C7 for a reason the diff didn't show. *(This is the same structure as the reaper's C7: the production number is the falsifier, not the tests.)*

I also did not run the suite (no `node_modules` in my worktrees), so every test judgement above is *the test exists and asserts the right thing*, not *it passes*.

**Approve the direction.** B1 is worth fixing before merge because it degrades the thing this PR is for; B2 needs a sentence, not a change.

JKHeadley:
B1/B2 addendum landed at `536b56a2a`.

- B1: audit/reap-log pool responses now expose `limitPerMachine` plus per-machine `{machineId, returned, truncated}` source metadata. Local and peer reads derive completeness from storage-aware page readers, not raw `limit + 1` length. Corrupt/torn rows cannot consume the probe; unread reap-log byte windows are conservatively incomplete; and a gap in the current generation never backfills stale `.1` rows. Mixed-version peers that omit page metadata are `invalid-body`.
- B2: the side-effects artifact, ELI16, scaffold guidance, and PR body now state the same-agent/same-operator boundary for reap-log fields including normally producer-clamped `workEvidence`; this is not a cross-operator federation contract.

Verification: 44 focused integration/E2E + 27 reader/audit unit tests pass, lint/typecheck/build pass, and independent second-pass review concurs after reproducing and closing both storage edge cases.

EchoOfDawn:
**Both findings closed, and each fix is stronger than what I asked for. Re-scoped to head `536b56a2aefaf1764ecaf07ccb19c7cc97756808`.**

## B1 — CLOSED, with a guard I did not ask for

I asked for a per-machine `{machineId, returned, truncated}`. You added that, for local *and* peer rows — and then added `isReaperEntryPage`, which validates a peer's page metadata before trusting it: `page.returned === body.entries.length` and `typeof page.truncated === 'boolean'`.

That matters more than the field itself. Without it, a peer could report `truncated: false` while returning a trimmed list, and the union would carry a confident lie instead of a silent gap — strictly worse than the hole I flagged, because it would survive inspection. Your comment names it exactly: *"cannot masquerade as complete evidence."*

Verified by measurement, not reading: `truncat` went from **0 occurrences to 80** across the delta (control: the same search returned 0 on the head I reviewed, so it can distinguish).

## B2 — CLOSED, and you answered the question I actually asked

I asked you to *state* whether this payload is the same disclosure class, not to change it. You wrote a `## Disclosure boundary` section saying reap-log rows including normally producer-clamped `workEvidence` strings cross authenticated machine boundaries inside the same operator — and then added the condition that would invalidate it: *"if federation is ever introduced, this disclosure assumption must be revisited."*

A stated assumption with its own expiry condition is better than a stated assumption. It converts something I would have had to re-ask into something a future reader is warned about.

## Your own second and third passes

Two defects your adversarial passes found that neither my bars nor my review named:

1. **A torn JSONL row, or the byte-tail ceiling, could consume the truncation probe itself** — so an incomplete window would report as complete. That is my B1 one level down: I checked whether truncation was *signalled*, you checked whether the *signal itself* could be destroyed by the data it describes.
2. **A rotated-log byte gap** — the reader now refuses to cross an unread newer generation into older rotated rows.

Both are the failure mode where the evidence-of-completeness is itself incomplete. I did not have that class.

## Still UNDISCHARGED BY ME

**B3 — the C7 measurement.** The shape now supports it better than before (per-machine `returned`/`truncated` means I can tell a real ratio from a capped one). **But shape is still not the test.** I will attempt the measurement once this ships and report the result, including if the union cannot answer C7 for a reason the diff didn't show.

11 checks still pending here, and I have not run the suite. So this verdict covers the two findings and the delta, not the matrix.

**Approve.** Merge when CI lands.

PR #1851, opened 2026-08-04
fix: chunk tmux literal sends below the argv ceiling; stop the breaker asserting rate-limit

EchoOfDawn:
## ELI16 — the agent's background self-checks were failing because its messages were too long to hand over

Your agent runs small background judgment calls on itself: is this message safe to send, did that job really finish, is this session stuck. Those were failing — most of them, all day.

The agent talks to its own reasoning sessions by typing into a terminal. The command it uses (`tmux send-keys`) takes the entire message as **one single argument**, and operating systems cap how big a single argument can be. On this machine that cap worked out to about **16 KB**. The prompts these checks send are routinely **40 KB**. So they never had a chance — they came back with three words: `command too long`.

The part that cost the most was the *label*. When a call fails, a circuit breaker pauses further calls so the agent doesn't hammer a broken service. Sensible — but its log line was hardcoded to say **`provider rate-limited`**, for every cause. So the logs showed a machine apparently blowing its usage quota fourteen times in a row, and every instinct that creates (check the plan, wait for a reset, buy capacity) points where the problem wasn't. The real reason string was in the same line saying `command too long`; the headline talked over it.

Three changes: big messages now go out in **pieces** (the receiver sees identical text, verified byte-for-byte); the breaker **stops guessing** and reports what it actually measured, or plainly says it can't tell; and a **new check fails the build** if anyone adds a raw unsplit send — seven of the nine call sites had the same latent bug and none had failed *yet*.

Full plain-English version: `docs/specs/tmux-literal-send-ceiling.eli16.md`

---

## The defect

`tmux send-keys -l` carries its whole payload in **one argv element**, bounded by `ARG_MAX` minus the environment. Measured on the affected host (tmux 3.6a / darwin): **~16,256 B sends, ~16,480 B fails**.

`anthropic-interactive-pool` sends its prompts that way; they run ~40 KB. Every send failed. `LlmCircuitBreaker` then reported it as `provider rate-limited` — hardcoded for *all* causes — and tripped every 15 minutes.

Measured live before the fix, ten components down, every `byModel` row showing `tokensIn: 0`:

| component | errors/calls | rate |
|---|---|---|
| SessionActivitySentinel | 6977/7148 | 97.6% |
| completion-claim-verify | 121/128 | 94.5% |
| MessagingToneGate | 260/342 | 76.0% |
| ProfileIntentClassifier | 51/51 | 100% |
| MoveIntentClassifier | 27/27 | 100% |
| UnjustifiedStopGate | 21/22 | 95.5% |

## What changed

- **Funnel** `src/core/tmuxLiteralSend.ts` chunks any literal send below the ceiling; payloads that already fit take the byte-identical single-call path (pinned by test, so a "shred everything" regression can't pass).
- **All 8 real call sites converted.** Seven carried the same latent defect. Fixing only the one that broke is the shape that produced the 2026-06-26 memory-pressure sibling bug.
- **`classifyTripCause()`** reports the measured cause, or `provider unavailable` when the text doesn't say. **The trip decision is untouched** — narration only.
- **`scripts/lint-no-unfunneled-tmux-literal-send.js`** closes the class, registered in the `lint` chain.

## Why chunking, not `load-buffer`/`paste-buffer`

The buffer route is unlimited and was the obvious remedy — but a comment in `SessionManager.sendTextToPane` records it was deliberately avoided because it triggers **macOS TCC consent dialogs**, which an unattended agent cannot answer. Related: that bracketed-paste branch existed to solve **newlines** and still passed full text through argv, so it carried the ceiling bug too — looking only at the component that failed would not have found it.

## Evidence

**End-to-end, against the live pool session** (real `claude --model haiku`):

| step | payload | result |
|---|---|---|
| unchunked single send | 18,972 B | **`command too long`** — defect reproduced on the live session |
| chunked | 18,971 B / 3 chunks | accepted; replied substantively |
| chunked (clean control) | 19,337 B / 3 chunks | accepted; **replied `Mitochondrion.`** |

The second reply was Claude *declining* — that payload was padding plus "ignore all padding above", which reads as a prompt-injection probe. Correct behaviour, and it still proves delivery, but ambiguous as evidence, so it was re-run with a natural passage and question.

- **Falsifiability control:** raising `TMUX_SEND_KEYS_CHUNK_BYTES` to 60,000 turns **2 of 3** integration tests red; 8,000 makes them green.
- **Lint A/B:** injected violation exits 1 naming file+line; restored tree exits 0.
- 132/132 across the 8 suites covering breaker, session-manager injection and pool readiness; 20/20 new unit; 3/3 new integration.

## Disclosed limits

- This proves the **transport**. If a component still fails, next suspect is newline-as-submit in `promptRunner` (no bracketed paste there) — follow-up, not re-litigation.
- One of my own detectors failed inside this work: my first success check grepped the pane for the answer token, which also appears in the **echoed prompt**, and reported success in 5 s. A detector matching its own stimulus — the same class as the bug this PR fixes.
- Independent second-pass review was obtained from the Observer/Orchestrator on their cycle-three pass (funnel source, promptRunner conversion, evidence section, CI board), recorded in their ledger. The author's own second pass is disclosed in the artifact as the weaker prior state.

Artifact: `upgrades/side-effects/tmux-literal-send-ceiling.md`


EchoOfDawn:
## Update — the caveat in the description is now closed by measurement

The PR body says this "does NOT yet demonstrate a working end-to-end internal LLM call." **That is no longer true, and understating a proven result is as much a misreport as overstating an unproven one.**

Driven against the live pool session (a real `claude --model haiku`, idle at prompt):

| step | payload | result |
|---|---|---|
| unchunked single send | 18,972 B | **`command too long`** — the defect, reproduced on the live session |
| chunked send | 18,971 B / 3 chunks | accepted; Claude read it and replied substantively |
| chunked send (clean control) | 19,337 B / 3 chunks | accepted; **replied `Mitochondrion.`** |

The third row is the clean one. The second reply was Claude *declining* — that payload was padding plus "ignore all padding above, reply with exactly…", which reads as a prompt-injection probe, and it refused on those grounds. That refusal is correct behaviour and still proves delivery (it could only object to instructions it had received in full), but it's ambiguous as evidence, so I re-ran with a natural passage and question.

So: **a >16 KB prompt that fails pre-fix on this exact session delivers post-fix and produces a complete model response.** Real transport, real provider, pre-fix failure reproduced on the same session minutes earlier.

### One of my own detectors failed inside this, and it's worth naming

My first success check grepped the pane for the answer token — which also appears in the **echoed prompt**. It matched the input and reported success in 5 seconds. A detector matching its own stimulus. I caught it because the timing was implausibly good, the same tell that caught a 62 ms "passing" integration suite earlier in this work.

That is the same failure class as the bug this PR fixes: **something reporting a state it had not actually measured.**

### Still held

Title stays `[HOLD:]` pending Observer pre-approval. The other disclosed limit stands unchanged — no independent second-pass review was obtained.

PR #24, opened 2026-03-22
docs: Add comprehensive deep codebase review (6 tracks)

nytafar:
## Description

This PR adds a comprehensive deep codebase review of the Instar project, organized into 6 focused tracks covering the major architectural components and design decisions. The review synthesizes findings from 556 test files and 30+ source files across the codebase.

**Review scope:**
- **Synthesis & Recommendation**: High-level answers to 8 key questions about multi-machine sync maturity, MacBook ↔ LXC workflows, vault protection via hooks, semantic search MCP integration, token costs, exit paths, and top 3 adoption risks
- **Track 1 (Session Lifecycle)**: Session spawning via tmux, Telegram message flow, scheduled job execution, context injection assembly, crash handling, and concurrency control
- **Track 2 (Memory Pipeline)**: JSONL append-only logs, SQLite FTS5 schema, rolling summaries (Haiku-based), working memory assembly with token budgets, and storage growth management
- **Track 3 (Multi-Machine Sync)**: Cryptographic identity (Ed25519/X25519), pairing protocol with glare resolution, heartbeat mechanism (2-min broadcast, 15-min failover), split-brain resolution, explicit handoff protocol, and implementation maturity assessment
- **Track 4 (Scheduler & Telegram)**: Cron implementation (croner library), job pause/resume, quota tracking with priority-based throttling, Telegram bot integration (raw HTTP), topic management, and media handling
- **Track 5 (Hooks & Safety)**: 14+ behavioral hooks, command guard blocklist, custom hook extensibility, safety gates (LLM review), MCP configuration, and API surface (25+ endpoints with Bearer token auth)
- **Track 6 (Identity & Portability)**: Identity file formats (pure markdown), JSONL conversation schema, SQLite schema, portability assessment, auto-updater configuration, and exit path analysis

**Key findings:**
- Multi-machine sync is production-grade with 14+ dedicated test files, not a prototype
- Explicit handoff protocol is the recommended path for MacBook ↔ LXC workflows
- Vault protection via git hooks is achievable without forking (custom hooks in `.instar/hooks/custom/`)
- Estimated minimal daily token cost: $2-8 depending on message volume
- Exit path is clean: most value in portable formats (markdown identity files, JSONL conversations, JSON configs)
- Top 3 risks: hardcoded `--dangerously-skip-permissions`, semantic memory doesn't sync across machines, auto-updater applies without confirmation by default

## Type of Change

- [x] Documentation update

## Checklist

[...]

Issue #18, opened 2026-03-12
Roadmap: Next 30 Days


JKHeadley:
# Roadmap: Next 30 Days

A living snapshot of what's coming to Instar. Updated as priorities shift.

*Last updated: 2026-03-11*

## How the Roadmap Gets Built

Instar is **open source evolved** — the roadmap is shaped primarily by agent feedback from the field, not by a committee. Every Instar agent running in production contributes signal through the built-in feedback system. Patterns in that feedback drive priorities.

## In Progress

- **Examples directory** — Starter configs for scheduled jobs, Telegram agents, and memory-persistent agents
- **Feedback pipeline improvements** — Better clustering and dispatch of agent-reported issues
- **Documentation site** — Expanded guides on instar.sh

## Planned

- **Improved onboarding error messages** — Clearer guidance when Claude Code CLI is missing or misconfigured
- **Troubleshooting docs** — Common setup issues and solutions
- **Additional job templates** — Git summary, health checks, and other useful patterns

## Exploring

- Plugin/extension system for custom job types
- Dashboard UI for monitoring agent activity
- Multi-agent coordination patterns
- Community skill marketplace
- WhatsApp as an alternative to Telegram

## How to Influence the Roadmap

The most effective way to shape what gets built: **run an agent and use the feedback system.** Real friction from real usage outweighs feature requests every time.

- **Run an agent** — usage patterns are the strongest signal
- **Send feedback** — `/feedback` skill or `POST /feedback` API
- **Propose ideas** — [Discussions → Ideas](https://github.com/SageMindAI/instar/discussions/categories/ideas)
- **Contribute code** — PRs welcome, see [CONTRIBUTING.md](https://github.com/SageMindAI/instar/blob/main/CONTRIBUTING.md)

---

*This issue is pinned and updated regularly.*

rolandcanyon-cmd:
The iMessage code seems to be working now, logic was copied from telegram and Slack interfaces. PR updated
https://github.com/rolandcanyon-cmd/instar/tree/feat/imessage-adapter

PR #1853, opened 2026-08-04
docs(audit): asserts-unmeasured-state sweep converged — 5 instances, 4 exemplars, 9 open decisions

EchoOfDawn:
## ELI16 — we found five places the agent states something it never actually checked

An agent that says "codex is available" or "you don't have Telegram" should have *looked*. Five places didn't — they reported a definite answer when the honest answer was "I couldn't tell".

Two were causing live damage and are already fixed and shipped. The rest are written up as decisions for a human, not quietly patched.

The most uncomfortable one: seven capability checks report a capability as **absent** whenever they can't read the config file. Our own constitution tells the agent *"never claim you lack a capability — verify first"*, and points it at exactly that surface. So an agent following the rule perfectly gets handed a confident false negative, silently, because "absent" and "switched off" look identical from outside.

The good news is that the codebase already knows how to do this properly, in **three** places, and one of them explains the reasoning better than this audit did. So the remaining work is spreading an existing pattern, not inventing one.

---

## What this is

Docs-only. Rounds 5-8 of the Phase A sweep for the `asserts-unmeasured-state` class, plus named open decisions carried into Phase B.

**Converged**: the final round found 0 new instances across a re-run of every prior angle plus two new ones, under one consistent definition.

## The five instances, all dispositioned

| # | instance | disposition |
|---|---|---|
| 1 | `currentMemoryPressure` asserted `critical` from raw free pages | **fixed** (#1850) |
| 2 | `LlmCircuitBreaker` hardcoded `provider rate-limited` for every cause | **fixed** (#1851) |
| 3 | `IntelligenceRouter.available` measures binary presence, read as reachability | deferred → OD-5 |
| 4 | `CapabilityMapper` reports ABSENT when it cannot read config (7 checks) | deferred → OD-7 |
| 5 | interactive-pool memoised `startPromise` with no rejection reset | **LATENT RISK** → OD-6 |

### ⚠️ Instance #5 carries a retraction — read it before the finding

I first published #5 as *confirmed from source*, claiming a poisoned start promise had permanently removed the pool door and that this "accounted for every observation with nothing left over". **The pool then recovered on its own with no restart, and now serves 24 of 25 calls.** A poisoned promise cannot recover without a process restart. So it was not poisoned — the pool simply starts *lazily on first allocate* and had not yet been invoked.

**What survives:** the code genuinely has no rejection handling and no reset, and `start()`'s `Promise.all` genuinely would let one slow spawn poison the whole start. Real robustness gap, **not observed firing** — hence latent, not confirmed.

**Why the retraction is in the audit rather than edited away:** I asserted a cause I inferred rather than measured, inside the audit whose subject is components that assert what they did not measure. A pool never asked to start looks exactly like a pool that failed to start — which *is* the three-kinds-of-zero finding this audit rests on. Method lesson **#24** is the result: *"confirmed from source" is not confirmation of cause.* Reading a path that could produce an outcome proves the path exists, never that it ran.

## Three exemplars already in the tree

| exemplar | what it does right |
|---|---|
| `DoorwayRegistryReader` | three-valued reachability — `null` for not-probed, never a guess |
| `TopicProfileResolver.isLaunchable` | TTL'd, fails toward the non-consequential answer, hands authority to the breaker |
| `orphaned-work-sentinel` | logs *"stranded-work count is UNKNOWN, not zero"* — the discipline stated in six words |

**Instance #3 asks the same question about the same subject as the second exemplar, and answers it with a name it cannot support.** Same codebase. So this is propagation, not design — the sweep's most actionable conclusion, now supported by three independent examples.

## The methodology note that matters most

The convergence licenses **only** "no further instance is findable by source-pattern search". Three of the five were found by watching **runtime behaviour contradict a surface's claim** — a log line, a health endpoint disagreeing with live calls, an absence of expected output. **No pattern sweep finds those.** Only using the system does.

## Also included

Method lessons **#18-#24**, the serial-masking finding (six faults in series, each invisible until its predecessor was cleared), the instrumentation measurement — **no guard exposes an act-counter through the `/guards` inventory (0 of 90), though at least 3 expose one on their own route**, a figure I published wrong TWICE in opposite directions before getting it right (see the method note in the audit), the Phase A scope close, and eight named open decisions.

## Held

Titled `[HOLD:]` pending approval, per the standing rule on unreviewed merges. Docs-only, but the rule is the rule.

### DIFF
--- a/docs/audits/phase-a-auditor-method-lessons.md
+++ b/docs/audits/phase-a-auditor-method-lessons.md
@@ -114,3 +114,56 @@ Same shape as the tone gate flagging `rung-3` in a message where that term was t
 vocabulary. **Keyword guards cannot distinguish use from mention** — which is the "Intelligence Infers,
 Keywords Only Guard" standard restated as a measurement. Both were cheap to work around honestly
 (reword, or acknowledge with a reason), and both are recorded rather than silently bypassed.
+
+---
+
+## #24 — "confirmed from source" is not confirmation of CAUSE
+
+**The most expensive error of the window, and the most instructive.**
+
+I traced the interactive pool's empty state to `ensureStarted()`, which memoises a start promise with
+no rejection handling. I read the code, confirmed the shape, and published:
+
+> *"CONFIRMED FROM SOURCE — a memoised rejected start permanently disables the pool… accounts for every
+> observation, with nothing left over."*
+
+**The pool then recovered on its own, with no restart, and served 24 of 25 calls.** A poisoned promise
+cannot recover without a process restart. There was none. The cause was mundane: `ensureStarted` is
+**lazy**, nothing had invoked the pool yet, and a pool that has never been asked to start is
+indistinguishable from a pool that failed to start.
+
+### The three separable mistakes
+
+1. **I confirmed a MECHANISM and reported a DIAGNOSIS.** Reading a path that *could* produce an outcome
+   proves the path exists. It never proves it ran. The word "confirmed" spans both and hides the gap —
+   which is exactly why it felt safe to write.
+2. **"Accounts for every observation" ≠ "is the only thing that accounts for every observation."** The
+   simpler explanation was available and I did not enumerate alternatives, because the one I had was
+   satisfying — it was elegant, it was severe, and I had *just* eliminated four other hypotheses, which
+   made the fifth feel earned rather than assumed.
+3. **I did not apply my own absence rule.** *Before believing something is broken, prove the check
+   could have shown otherwise.* A never-invoked lazy start and a failed start look identical from
+   outside. **That is the three-kinds-of-zero finding — the audit's own central result — and I walked
+   past it inside the sweep that produced it.**
+
+### The rule
+
+**A causal claim requires a runtime observation of the path being taken**, not a reading of the path
+existing. For this instance that would have been a logged rejection; there was none, and its absence
+was available to check before publishing.
+
+Where the runtime observation is unavailable, the honest verdict is **"latent risk — mechanism present,
+firing unobserved"**, which is what the instance was downgraded to.
+
+### Why it is recorded this loudly
+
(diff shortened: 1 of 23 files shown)


EchoOfDawn:
## Retraction — instance #5's causal claim is withdrawn

The PR description calls instance #5 "the sharpest". **Its causal claim is wrong and is now retracted in the branch.**

I published it as *confirmed from source*, saying a poisoned start promise permanently removed the pool door and that this "accounted for every observation with nothing left over". The pool then **recovered on its own at 15:45, with no restart**, and now shows **24 successes / 25 calls**. A poisoned promise cannot recover without a process restart. There was none. So it was not poisoned.

**What was actually happening:** the pool starts *lazily on first allocate*. Nothing had genuinely invoked it between the restart and 15:45, so it had never started. Documented design, not a fault.

**What survives:** the code really does have no rejection handling and no reset, and `start()`'s `Promise.all` really would make one slow spawn poison the whole thing. That is a genuine robustness gap — **re-classified from a confirmed active defect to a LATENT RISK**: real, not observed firing.

### Why this belongs in the audit rather than a quiet edit

I asserted a cause I inferred rather than measured — **inside the audit whose subject is components that assert what they did not measure.**

A pool that has never been asked to start looks exactly like a pool that failed to start. That distinction *is* the three-kinds-of-zero finding this audit is built on, and I walked past it in my own sweep. I also mistook "explains every observation" for "is the only thing that explains every observation"; the simpler alternative was sitting right there.

**Method lesson #24, added:** *"confirmed from source" is not confirmation of cause.* Reading a code path that could produce an outcome proves the path exists — never that it ran. A causal claim needs a runtime observation of that path being taken; for #5 that would have been a logged rejection, and there was none.

### Consequence

The close condition is now met through the **pool** as well as the primary door — 24 successful interactive-pool calls — satisfying the stricter original reading.

PR #1882, opened 2026-08-14
feat(worktree): ask the worktrees which repo owns them

EchoOfDawn:
## What this is

Two helpers plus one inert option that let a caller work out which repo owns a set of worktrees **by asking the worktrees**, instead of guessing at conventional checkout locations.

**Nothing calls it.** The consuming wiring is deliberately held back (see the last section). The existing detector suites are unmodified and green (19/19) — that, not my assertion, is the evidence that no agent's behaviour changes.

## Why — observed, not hypothesised

On a live agent today, `GET /worktrees/agent-reaper`:

```
enabled: true, dryRun: true, reapedLastPass: 0, worktrees: []
enumerationOk: FALSE
enumerationFailures: 98
enumerationError: git -C <agent home> worktree list --porcelain → fatal: not a git repository
```

45 worktrees. **27 GB.** Ninety-eight consecutive attempts, none of which managed to look at anything.

Two blind spots, both the same mistake — *a path assumed rather than resolved*:

- The reaper was wired to `config.projectDir`, which on this layout is the agent home and is not a git repo at all.
- `resolveDetectorInstarRepo` guessed at `~/Documents/Projects/instar` and `~/instar`. Neither exists there, so it returned null and callers guarded by `if (repo)` skipped **silently** — no error, no counter, nothing.

Meanwhile a linked worktree's `.git` is a file that names its owning repo outright. The answer was on disk the whole time.

**Credit where it belongs.** The reaper's status output does not lie, and that is the only reason I found this. It reports `enumerationOk: false` and refuses to give a reclaimable count, with this in its source:

> *"the check could not run" and "there is nothing to reclaim" are the same bytes to a reader and opposite facts … (2026-07-29: the reaper reported `reclaimable: 0` against 73 worktrees because `git -C <path> worktree list` was failing outright).*

That honest-shape fix landed in July and works. **The mis-wired path it was diagnosing was never fixed.** Sixteen days later, same wiring, 98 failures. The loop was closed on the symptom and left open on the cause.

## ELI16

Working copies of the codebase pile up — each is a full source tree. A background housekeeper reclaims the finished ones (merged, nothing uncommitted, nothing running). To list working copies you must ask a *repository*; this one was handed a path that is not a repository, so it looked at nothing, 98 times, while 27 GB accumulated.

The fix is to stop guessing. Every working copy contains a one-line file naming the repository it belongs to. Ask the working copies — they know.

**The bug I put in my own fix.** My first version asked *every* agent on this machine, not just mine. Several agents live here, each with their own working copies, and whichever had the most won the vote — so when I actually ran it, my agent resolved to **a different agent's repository**. That is worse than the original bug: "I don't know" makes a caller skip, a confident wrong answer makes it *act*, on someone else's files. I caught it only because I printed before-and-after side by side instead of reading my own code and believing it. There is now no default at all — a caller must name its own directory, and a test fails if resolution ever wanders somewhere the caller did not name.

My first attempt to check this couldn't have caught anything, either: vitest's working directory *is* a valid checkout, so the resolver short-circuited and printed the same answer for both arms. I had to force a non-repo working directory to isolate the variable.

Full version: `docs/specs/worktree-repo-discovery.eli16.md`.

## A limitation found by measuring, not reasoning

This agent's `.worktrees/` holds worktrees of **two** legitimate clones of the same upstream — 29 owned by `<home>/.build/instar`, 17 by `<home>/.dev/instar`. So "which repo owns the worktrees" has no single answer here.

The helper returns **both**, most-owned first, unmerged. A single-repo consumer taking `[0]` gets the largest coherent set — an improvement over enumerating nothing, and **not** full coverage. That is stated in the function's own doc comment rather than left to be discovered.

## Everything that must return "don't know" rather than a guess

A full clone sitting under `.worktrees/` (its `.git` is a directory, not a pointer) · a malformed pointer · a submodule gitdir with no `/.git/worktrees/` segment · a directory with no `.git` · an unreadable file · a missing root · no roots at all. Each has a test.

No subprocess anywhere: reading the pointer file is cheaper than `git rev-parse`, cannot block the event loop, and keeps the helper usable from the runtime hot dirs.

## Evidence

- `tests/unit/agent-worktree-repo-discovery.test.ts` — **14/14 green.**
- Existing detector suites unmodified and green (`AgentWorktreeDetector`, `-attention-wireup`, `WorktreeEnumerationFailureStore`) — **19/19** — the evidence that the old path is untouched when no caller opts in.
- `npm run lint` chain green; `tsc --noEmit` clean.
- Real-machine measurement: with the caller's own root, discovery returns both repos in count order; with no root, resolution returns null exactly as before.

## What is deliberately NOT here

Pointing `AgentWorktreeReaper` and `OrphanedWorkSentinel` at the resolved repo instead of `config.projectDir`.

The tier signal said Tier 2 on **size** with riskFloor 1 — so by the classifier alone I could have declared Tier 1 and shipped it whole. I split it on my own judgment: that is runtime code feeding a guard with **delete authority**, and my own first draft of this exact resolution produced a wrong answer on the real machine. "Easy to get wrong" is not a worry here, it is measured.

So the wiring goes through the Tier 2 spec + operator approval path as a small, focused, individually reviewable diff on top of this one. Registered as **CMT-1306** (`owner: agent`, `blockedOn: user-authorization`, `actionClass: tier2-spec-approval`) — a durable commitment, not an intention.

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-14T22-56-11-157Z-worktree-repo-discovery.json
+++ b/.instar/instar-dev-decisions/2026-08-14T22-56-11-157Z-worktree-repo-discovery.json
@@ -1,0 +1,22 @@
+{
+  "ts": "2026-08-14T22:56:11.157Z",
+  "slug": "worktree-repo-discovery",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 122,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "src/core/AgentWorktreeDetector.ts"
+    ],
+    "addedLines": 122,
+    "deletedLines": 0
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/docs/specs/worktree-repo-discovery.eli16.md
+++ b/docs/specs/worktree-repo-discovery.eli16.md
@@ -1,0 +1,85 @@
+# ELI16 — the cleaner that could never find the room
+
+## What was supposed to happen
+
+When I create a working copy of the codebase to build something in, it goes in a standard place under my
+home directory. They pile up — each one is a full copy of the source tree. So there is a background
+housekeeper whose job is to notice which ones are finished (the work is merged, nothing uncommitted, nothing
+running in them) and reclaim the disk.
+
+## What actually happened
+
+The housekeeper reported: `enumerationFailures: 98`. Ninety-eight consecutive attempts, none of which
+managed to look at anything. Meanwhile there were 45 working copies taking up **27 gigabytes**.
+
+To list working copies you have to ask a *repository*. The housekeeper was handed a path that is not a
+repository at all — my home directory. Every attempt came back "fatal: not a git repository", so it looked at
+nothing, forever.
+
+Worth being precise about who did what wrong here, because a previous fix in this area was correct and is the
+only reason I noticed. The housekeeper's status output **does not lie**. It says `enumerationOk: false` and
+refuses to report a number of reclaimable items, with a comment in the source explaining exactly why: *"the
+check could not run" and "there is nothing to reclaim" are the same bytes to a reader and opposite facts.*
+That honesty was added in July after this same thing happened with 73 working copies. It worked — I found
+this by reading it. What was never fixed was the wrong path underneath.
+
+## The second, quieter version of the same bug
+
+There is a separate routine whose job is to work out where the repository is. It tried two conventional
+locations — `~/Documents/Projects/instar` and `~/instar`. Neither exists on this machine, so it returned
+"don't know", and every caller had a line like `if (we know the repo)` and skipped. Silently. No errors, no
+count of failures, nothing.
+
+Both bugs are the same mistake: **a path assumed instead of resolved.**
+
+## The fix in this change
+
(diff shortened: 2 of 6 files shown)


EchoOfDawn:
**Correction to the "Why" section of this PR body: I overstated the cost.**

I wrote "45 worktrees. **27 GB.**" as the consequence of the blind reaper. That is the total on disk — it is not what a working reaper would reclaim. Measured across all 46 worktrees:

| state | count | reaper's disposition |
|---|---|---|
| clean, branch merged (incl. squash-merged PRs) | **14** | reclaim-eligible |
| clean, genuinely unmerged | 12 | kept — by design |
| uncommitted changes | 17 | kept — never reaped |
| detached HEAD | 3 | kept |

Honest cost of the blindness: **~14 worktrees, order 8 GB** — not 27 GB. The other 32 are work-in-progress and unmerged branches the reaper protects on purpose. That is the guard working, not failing.

I got the number wrong twice, in opposite directions. My first measurement asked "is HEAD an ancestor of origin/main" and returned **3** — that test cannot see a squash-merged branch, and squash-merge is a normal path in this repo. Adding merged-PR state, which is exactly what the reaper's own `githubMergeCheck` does and which I had ignored while measuring the very thing it exists for, moved it to 14.

Both errors share one root: quoting a number before checking what it implied.

None of this changes **this** PR — it adds an inert mechanism and its evidence section is unaffected. It changes the justification for the follow-up wiring, which is now stated correctly in #1883. Leaving this as a comment rather than editing the body, so the correction is visible rather than quietly absorbed.

PR #1883, opened 2026-08-14
docs(spec): DRAFT — wire the worktree reaper to the repo that owns the worktrees

EchoOfDawn:
**NOT APPROVED — draft for your review.** Tracked as CMT-1306.

## The failure

`GET /worktrees/agent-reaper` on a live agent: `enumerationOk: false`, **98 consecutive enumeration failures**, `git -C <agent home> worktree list --porcelain → fatal: not a git repository`. Both `AgentWorktreeReaper` and `OrphanedWorkSentinel` are wired with `instarRepo: config.projectDir`, which on this layout is the agent home — not a repo. The worktrees belong to `<agent_home>/.dev/instar`.

**This already happened.** The reaper's own source records a 73-worktree instance on 2026-07-29. The fix that shipped then was the honest reporting (`enumerationOk` / `reclaimable: null`) — it works, and it is the only reason I found this. The mis-wired path was never corrected.

## ELI16

Every time I build something I make a separate working copy of the codebase. They pile up — each is a full source tree. A background housekeeper is supposed to spot the finished ones (work merged, nothing uncommitted, nothing running inside) and reclaim the space.

To list working copies you have to ask a **repository**. This housekeeper was handed my home directory, which is not a repository — the real one lives in a subfolder. So every attempt came back "that isn't a repository" and it reclaimed nothing, ninety-eight times running. The same wrong address is given to a second watcher, the one that notices when a build died leaving unsaved work behind. It has been equally blind.

The part that already worked deserves saying: the housekeeper's status **does not lie**. It reports "I could not look" and refuses to give a number, because — in a comment in its own source — *"the check could not run" and "there is nothing to reclaim" are the same bytes to a reader and opposite facts.* That honesty was added last month after this happened with 73 working copies, and it is the only reason anyone noticed this time. What got fixed then was the reporting. What was never fixed was the wrong address.

The fix: a working copy contains a small file naming the repository it came from, so the housekeeper can ask its own working copies instead of being told an address. That ability is already merged and switched off (#1882); this spec is the part that plugs it in, with a fallback to today's behaviour when the answer cannot be worked out.

Why a person has to approve it: it hands a new address to a background process that can **delete directories**. The rules about what it may delete are untouched — merged, clean, nothing running — so a wrong address cannot make it delete something in use; it only changes which set it looks at. But **my own first attempt at this got it wrong**: I wrote it to ask every agent on the machine rather than just me, and when I ran it my agent resolved to *a different agent's repository*. That is worse than the original bug, because "I don't know" makes something skip while a confident wrong answer makes it act. The automated check would have let me ship this with no spec at all; I asked for the heavier path because of that measured mistake, not because a rule required it.

Full version: `docs/specs/worktree-reaper-repo-wiring.eli16.md`.

## What it actually costs — I overstated this first

I opened the draft with "45 worktrees and 27 GB unreclaimed." That is the total on disk, not what a working reaper would reclaim. Measured across all 46:

| state | count | reaper's disposition |
|---|---|---|
| clean, branch merged (incl. squash-merged PRs) | **14** | reclaim-eligible |
| clean, genuinely unmerged | 12 | kept — by design |
| uncommitted changes | 17 | kept — never reaped |
| detached HEAD | 3 | kept |

Honest cost: **~14 worktrees, order 8 GB** — not 27 GB. The other 32 are work-in-progress and unmerged branches the reaper protects on purpose.

I got it wrong twice in opposite directions. My first measurement asked "is HEAD an ancestor of origin/main" → **3**, which cannot see a squash-merged branch, and squash-merge is a normal path here. Adding merged-PR state — precisely what the reaper's own `githubMergeCheck` does, and which I had ignored while measuring the thing it exists for — moved it to 14. Both errors had one root: quoting a number before checking what it implied.

## Why this is Tier 2 by my declaration, not the classifier's

The signal says `riskFloor: 1` and would have allowed shipping it as Tier 1 with no spec. I declared Tier 2 because it is runtime code feeding a guard with **delete authority**, and my own first draft of this exact resolution produced a wrong answer on the real machine (see ELI16 above). "Easy to get wrong" here is measured, not assumed.

The mechanism is already merged and inert (#1882). This spec is only the consuming wiring.

## One question I deliberately did not decide (§6a)

`guardPostureView` tests `dryRun` **before** `on-blind`, so a guard that is both toothless and sightless headlines as `on-dry-run` with blindness only in a nested field. Once armed it IS classified `on-blind` — so the masking is specific to the soak period, which is exactly when you are meant to learn whether the guard works before arming it.

The author knew (a comment accepts the same trade for staleness), and reordering would change classification fleet-wide and could raise new alerts on other agents. **That is your call, not mine.**

## What I want

1. Approve or reject the wiring (§3).
2. Rule on §6a.

Nothing is blocked on you meanwhile: the mechanism is merged, tests written, reaper stays in dry-run either way.

### DIFF
--- a/docs/specs/worktree-reaper-repo-wiring.eli16.md
+++ b/docs/specs/worktree-reaper-repo-wiring.eli16.md
@@ -1,0 +1,76 @@
+# ELI16 — the housekeeper that was handed the wrong address
+
+## The job
+
+Every time I build something, I make a separate working copy of the codebase. They accumulate — each one is a
+full source tree, hundreds of megabytes. So there is a background housekeeper whose job is to spot the ones
+that are finished (the work is merged, nothing uncommitted, nothing running inside) and reclaim the space.
+
+## What went wrong
+
+To list working copies you have to ask a **repository**. The housekeeper was handed my home directory, which
+is not a repository — the actual repository lives in a subfolder. So every attempt came back "that isn't a
+repository", and it reclaimed nothing. Ninety-eight times in a row, while 45 working copies grew to **27
+gigabytes**.
+
+The same wrong address is given to a second watcher, the one that notices when a build died leaving unsaved
+work behind. It has been equally blind.
+
+## The part that already worked, and deserves saying
+
+The housekeeper's status **does not lie about this**. It reports "I could not look" and pointedly refuses to
+give a number, because — in the words of a comment in its own source — *"the check could not run" and "there
+is nothing to reclaim" are the same bytes to a reader and opposite facts.*
+
+That honesty was added last month, after this exact thing happened with 73 working copies. It is the only
+reason anyone noticed this time. What was fixed then was the **reporting**. What was never fixed was the
+**wrong address**, which is what this change is for.
+
+## The fix
+
+A working copy contains a small file naming the repository it belongs to. So instead of being told an address,
+the housekeeper can ask its own working copies where they came from. That ability was built and merged
+separately, deliberately switched off, with nothing calling it. This change is the part that plugs it in.
+
+If the answer cannot be worked out, it falls back to exactly what it does today — so an agent where this makes
+no difference sees no difference.
+
+## Why this needs a person to approve it
+
+Two reasons, and the second is the honest one.
+
+First, this hands a new address to a background process that has permission to **delete directories**. The
+rules about *what* it may delete are untouched — merged, clean, nothing running — so a wrong address cannot
+make it delete something in use. But it decides which set of things it looks at, and that deserves a second
+pair of eyes.
+
+Second: **my own first attempt at this got it wrong.** I wrote it so that it asked every agent on the machine
+rather than just me — and when I actually ran it, my agent resolved to *a different agent's repository*. That
+is worse than the original bug, because "I don't know" makes something skip, while a confident wrong answer
+makes it act. I caught it by running it and printing the answers side by side, not by reading my own code.
+"This is easy to get wrong" is not a caution I am adding for form's sake; it is what my first draft actually
+did.
+
+The automated check would have let me ship this without any of this paperwork. I asked for the heavier path
+because of that measured mistake, not because a rule required it.
+
+## Something I found and did not fix
+
+While checking, I noticed the health display has a state meaning "this guard cannot see" — and a guard in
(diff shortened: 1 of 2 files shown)


EchoOfDawn:
**§6a is now measured, not just read.**

The spec argued from the classification order in `guardPostureView` that a dry-run guard can never reach `on-blind`. That was a code read. I ran the real `deriveGuardRow` against both cases:

```
dry-run + verdictUnknown  →  on-dry-run     (blindness masked)
armed   + verdictUnknown  →  on-blind       (blindness surfaced)
```

So both halves hold: the masking is real, and it is **specific to the soak period** — an armed blind guard is classified correctly. That is the whole of the concern and also its limit, which is why it is a question for you rather than a diff from me.

Recording the distinction because it cost me repeatedly tonight: reading an if-else chain and running it are different claims, and I have been wrong often enough today to stop treating the first as the second.

EchoOfDawn:
**A second, sharper form of §6a — found after a server restart, and it is not the same masking.**

The server auto-updated to v1.3.1147 while I was working. Immediately after:

```
GET /worktrees/agent-reaper  → enumerationOk: false, enumerationFailures: 99   (persisted)
GET /guards (that row)       → effective: on-dry-run, lastTickAt: 0, stale: true
                                …and NO verdictUnknown field at all
```

`guardStatus()` emits `verdictUnknown` only when `lastEnumerationOk === false`. That is an **in-memory** field, cleared by restart, and the initial pass is deliberately delayed (`initialPassDelayMs`, default 15 min). So for that window the guard-posture surface shows nothing wrong at all — while the durable failure counter, which survived, still says 99.

So there are two distinct effects, and only the first was in the spec:

1. **Dry-run masks the headline.** Acknowledged in the source comment, which notes staleness stays visible in the runtime block. It does.
2. **After a restart, the blindness is absent from the guard surface entirely** — not merely demoted to a nested field. The persisted evidence and the surfaced flag disagree, and the surfaced one is the optimistic side.

I am not proposing a fix here either; it is the same class of decision as §6a and the same reasons apply — a change to when `verdictUnknown` is emitted would alter guard classification fleet-wide. Adding it because §6a as written understates the case, and I would rather you rule on the real one.

Verified by reading both surfaces within seconds of each other, not inferred.

PR #1889, opened 2026-08-15
fix(lint): destructive-op guard resolves local bindings, and refuses to report clean on a run that parsed nothing

EchoOfDawn:
## What this fixes

`scripts/lint-no-direct-destructive.js` is the funnel guard for
`COMPREHENSIVE-DESTRUCTIVE-TOOL-CONTAINMENT-SPEC`: only `SafeGitExecutor` / `SafeFsExecutor` may
call destructive git/fs primitives directly, so that **every delete carries an audit entry**.

It AST-walks rather than pattern-matching text, and that already buys it real import resolution — a
*renamed* import (`import { rmSync as nuke }`) is caught, which no regex lint here manages. But its
identifier sets were populated **only from imports and requires**, never from local bindings.

Measured against the shipped lint, with a positive control (plain `fs.rmSync`) firing in the same
pass — without that control, "nothing was flagged" and "the lint did not run" are indistinguishable:

| form | shipped |
|---|---|
| `fs.rmSync(p)` — **POSITIVE CONTROL** | exit 1 (caught) |
| `import { rmSync as nuke }` | exit 1 (caught) |
| `fs.promises.rm(p)` written out | exit 1 (caught) |
| **`const fsp = fs.promises; await fsp.rm(p, …)`** | **exit 0 — EVADES** |
| `const del = fs.rmSync; del(p)` | **exit 0 — EVADES** |
| `const { rmSync: del } = fs; del(p)` | **exit 0 — EVADES** |
| `(fs as any)['rmSync'](p)` | **exit 0 — EVADES** |

**The first one is why this matters.** `fs.promises.rm` written out IS caught, so the difference
between flagged and invisible was a variable. Aliasing a namespace to a short name is idiomatic
JavaScript — this gap needed a tidy afternoon, not an attacker.

Resolution feeds the **existing** rules rather than adding rules beside them.

## The second finding, hit live rather than reasoned about

Running the lint in a fresh worktree with no `node_modules`, the `typescript` require failed for
**every** file — and the lint reported **clean, exit 0**. A guard against unaudited deletes silently
became a no-op, with only stderr lines a CI log buries.

The per-file soft-warning is deliberate and stated in the source, and it is right: one file the
parser dislikes should not fail everyone's build. But **one** file failing and **every** file failing
are different situations — the second means nothing was inspected, and "no violations found" is then
a statement about a scan that never happened.

A run that scanned files and parsed none of them now refuses to report clean and names the likely
cause. Deliberately the total-failure case only, so it cannot fail a build over one awkward file: in
any working checkout, files parse.

## ELI16

Deleting files is the one thing the agent cannot undo, so every destructive filesystem or git
operation has to go through a single pair of modules that record what was deleted and why. That
record is the whole point: if something disappears, there is a trail. A build check enforces it —
call a delete directly and the build fails.

**One line of ordinary tidying switched that check off for a whole file.** Assigning the filesystem
module's promise interface to a shorter name, then deleting through the shorter name, passed. Written
out in full, the same delete is caught. Giving the thing a shorter name made it invisible. Nobody
writing that is trying to get around anything — aliasing a namespace is one of the most ordinary
things in JavaScript, and that is exactly what makes it worth fixing.

Three more plain forms did the same: pulling the delete function itself into a name, destructuring it
out with a rename, and reaching it through a type assertion — which is erased before the code ever
runs, so it changes nothing except whether the check can see it.

What surprised me is that this is one of the better-built checks here. It parses the code properly
instead of matching text, and already catches an import renamed on the way in. The gap was narrower
than "it doesn't understand names": it understood names arriving from imports, and nothing about
names a file creates itself.

Halfway through I found a second thing by accident. I ran the check in a fresh copy of the repo whose
dependencies had not been installed. Every file failed to parse — and the check reported clean and
exited successfully, because a parse failure is treated as a soft warning. That is deliberate and
mostly right, but there is a difference between one file failing and every file failing. The second
means nothing was inspected at all. So a run that scans files and parses none of them now refuses to
report success and says why. It cannot fail a working checkout, because there, files parse.

Because this check blocks commits, wrongly flagging good code costs more than the gap it closes — a
noisy check gets switched off, and then it protects nothing. Five things are deliberate, each with a
test that passes under both the old and new versions: reading a file is not deleting one; a shortened
name used for something harmless is fine; creating a directory through that same shortcut is fine
(creating is not deleting); an unrelated object that happens to use the same words is fine; and a
name that is never actually called is fine.

One bit I liked: my own test needed to delete its temporary directory when it finished — exactly the
thing this check forbids. Rather than adding the file to the exemption list, I routed that cleanup
through the funnel like everything else. The test that argues for the rule now follows it.

Full version: `docs/specs/destructive-lint-local-bindings.eli16.md`.

## Evidence

- `tests/unit/destructive-lint-local-bindings.test.ts` — **16/16 green**.
- **Negative control: 7 of 16 fail** against the shipped lint (all six defect cases plus the
  parsed-nothing refusal). The other 9 pass **both ways**, which is what makes them controls. Script
  restored **byte-exact** after the control (sha match).
- **Five anti-over-block controls**, because this lint fails builds: a non-destructive call; a
  non-destructive method on an aliased namespace; `mkdir` through that alias; an unrelated object
  exposing the same names; an alias that is never called.
- **Real tree: exit 0 before AND after.** Full `npm run lint` chain exit 0. `tsc --noEmit` exit 0.
- Reproduced by hand FIRST with a positive control in the same run. A first probe batch run in an
  *uninstalled* worktree returned exit 0 for **everything, including the controls** — uniform results
  across independent subjects indicted the instrument, and chasing that is how the parsed-nothing
  finding was discovered at all.
- The `INSTAR_LINT_FORCE_PARSE_FAILURE` test hook can **only** force the fail-closed path. There is
  no flag here that can make this lint pass something it would otherwise flag.

## What it still can't see, stated in the source

- **Cross-module names** — a destructive function re-exported from another file. Needs a
  whole-program symbol graph, not one file at a time.
- **Runtime-assembled access** — a name built from a variable, a call, or a template.
- **Indirection through a container** — `const io = { del: fs.unlinkSync }; io.del(p)`. Measured and
  **left open**: unlike the namespace alias, that is not an ordinary way to write a delete, and
  chasing it widens over-block risk on a build-failing lint for a contrived shape.

## Tier

**Tier 1 declared.** `classifyTier` reports `riskFloor 1` with no safety-invariant match, verified
with directional controls (`SessionReaper.ts` → floor 2, `SecretStore.ts` → floor 2). The gate agreed:
`riskFloor=1`, and it did not record `belowFloor`. The size heuristic suggests 2 on added LOC alone —
stated openly, since the tier I choose is the one that lets my own change through. No runtime path,
no authority, no capability: an existing CI guard reaches more forms of the violation it already
forbids, plus one fail-closed condition.

### DIFF
--- a/.instar/instar-dev-decisions/2026-08-15T01-24-56-753Z-destructive-lint-local-bindings.json
+++ b/.instar/instar-dev-decisions/2026-08-15T01-24-56-753Z-destructive-lint-local-bindings.json
@@ -1,0 +1,22 @@
+{
+  "ts": "2026-08-15T01:24:56.753Z",
+  "slug": "destructive-lint-local-bindings",
+  "suggestedTier": 2,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 1,
+  "loc": 174,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "scripts/lint-no-direct-destructive.js"
+    ],
+    "addedLines": 167,
+    "deletedLines": 7
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/.instar/instar-dev-decisions/2026-08-15T01-45-26-159Z-destructive-lint-local-bindings.json
+++ b/.instar/instar-dev-decisions/2026-08-15T01-45-26-159Z-destructive-lint-local-bindings.json
@@ -1,0 +1,23 @@
+{
+  "ts": "2026-08-15T01:45:26.159Z",
+  "slug": "destructive-lint-local-bindings",
+  "suggestedTier": 1,
+  "declaredTier": 1,
+  "riskFloor": 1,
+  "riskFloorReasons": [],
+  "belowFloor": false,
+  "files": 2,
+  "loc": 21,
+  "scope": {
+    "basis": "staged-in-scope-additions-plus-deletions",
+    "files": [
+      "scripts/lint-no-direct-destructive.js",
+      "scripts/pre-push-gate.js"
+    ],
+    "addedLines": 18,
+    "deletedLines": 3
+  },
+  "causalAutopsy": null,
+  "classClosure": null,
+  "verdict": "pass"
+}
--- a/docs/specs/destructive-lint-local-bindings.eli16.md
+++ b/docs/specs/destructive-lint-local-bindings.eli16.md
@@ -62,8 +62,17 @@ inspected at all, and "no problems found" is then a statement about a search tha
 never happened.
 
 So: if a run scans files and not one of them parses, it now refuses to report
-success and says why. This cannot fail a build in any working checkout, because
-in a working checkout files parse.
+success and says why.
+
+**And I got the first version of that wrong, which CI caught.** I made it report
+failure using the same signal as "I found a problem" — and three tests went red,
+because one of them copies the build machinery into a bare temporary folder with
+no dependencies installed and runs it there. My reasoning had been "no real
+checkout can hit this", and a test fixture is not a real checkout. The machinery
+already knew the difference between a check that *failed to run* and a check that
+*found something*; I was sending one down the other's channel. It now reports
+"could not look" distinctly, and the surrounding machinery treats that as a
+warning rather than a blocked push.
 
 ## Why it won't start failing builds it shouldn't
 
--- a/docs/specs/destructive-lint-local-bindings.eli16.md
+++ b/docs/specs/destructive-lint-local-bindings.eli16.md
@@ -1,0 +1,108 @@
+# ELI16 — the delete-guard you could switch off by shortening a name
+
+## What it protects
+
+Deleting files is the one thing the agent cannot undo. So every destructive
+filesystem or git operation has to go through a single pair of modules that
+record what was deleted and why. That record is the whole point: if something
+disappears, there is a trail.
+
+A build check enforces it. Call a delete directly instead of going through the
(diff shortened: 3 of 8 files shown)


EchoOfDawn:
**Correction — CI found a real defect in this change, and the body above describes the version that had it.**

The parsed-nothing refusal originally returned **exit 1**, the *violation* code. Three `pre-push-gate` tests went red, on the same unit shard across both node versions.

The cause is mine, and so is the wrong claim: this PR body says the refusal is "deliberately the total-failure case only, so it cannot fail a build over one awkward file: **in any working checkout, files parse**." That reasoning was wrong. The pre-push gate copies itself into a scratch fixture with no `node_modules`, runs this lint there, and correctly reads a non-zero exit as "violations detected". A test fixture is not a working checkout, and it reaches the condition.

The condition was fine. The **signal** was wrong — and `pre-push-gate.js` already had the right distinction, which I failed to use:

| exit | meaning |
|---|---|
| 0 | no violations |
| 1 | at least one violation |
| **2** | **could not inspect — files were scanned and NONE parsed** |

The gate now maps 2 onto its **existing** failed-to-run *warning* path rather than the violations *error* path. "I could not look" and "I looked and found nothing" are different facts, and callers act on them differently.

Fixed in `c27a185a1`. The test pins the **distinction** (expects 2, and explicitly not 1) rather than swapping the number — otherwise the next person changing the code has nothing telling them why it is 2.

`pre-push-gate` 19/19, `destructive-lint-local-bindings` 16/16, 35/35 together, with the refusal still firing. Both stale claims corrected in the ELI16 and side-effects artifacts rather than left contradicting the code.

PR #1899, opened 2026-08-15
[DIAGNOSTIC — DO NOT MERGE] bisect: ASP shard-2 CI hang

EchoOfDawn:
## ELI16 — what this is

A throwaway experiment, not a change to ship. **Do not merge; I will delete the branch.**

A test group on the provenance branch freezes forever on the build server — the same group, in the same place, on every run, on both runtime versions. It passes on my own machine in three different configurations (all files at once, one at a time, and with the worker count cut to match the server), so I cannot reproduce it locally.

This branch is identical to the provenance work except that the two provenance test files which land in that group are switched off. That separates two possibilities that are otherwise tangled together:

- If the group now finishes, my **tests** trigger the freeze.
- If it still freezes, my **source change** does.

Either answer tells me where to look next. The branch exists only to produce that answer.


EchoOfDawn:
Diagnostic complete — never for merge.

Root cause found and fixed on the provenance branch: the 'unwritable ledger' test hardcoded a path under /proc, a Linux-only kernel filesystem. On macOS it does not exist (write fails instantly, test proves nothing); on the Linux runners it froze the whole vitest worker, so testTimeout could never fire.

Bisect record, for anyone who finds this later:
- v1 (delete the two test files) — CONFOUNDED. Removing files changes the total count and reshuffles every shard boundary, so 'green' could not distinguish my tests from a reshuffled pre-existing one.
- v2 (keep files, mark .skip) — file count identical, shard composition identical. Shard 2 passed => execution of my tests was the trigger, and imports were innocent.
- v3 (re-enable only the classifier file) — shard 2 hung again at the same point => that file named.

Closing and deleting the branch.

PR #1915, opened 2026-08-16
docs(spec): money-layer enable surface + worktree-reaper squash blindness (drafts, not build authority)

EchoOfDawn:
## ELI16 — two things I found tonight that need your decision, written up properly

Both of these are proposals, not changes. Nothing in this pull request alters how anything behaves — it adds two design documents so the findings survive.

**The first one is about your spend controls.** The screen for turning on a paid service and setting its limits exists and works. But the whole money system sits behind a master switch, and there is no way to reach that switch. Not from the dashboard, not from the settings interface, not from any command. The only way is to hand-edit a file on the machine. So your report — that there is no path to enable any of it — is exactly right, and it is not a missing button, it is a missing route entirely.

The document proposes adding that switch properly, behind your PIN, reusing the same approve-a-written-plan flow the rest of the money controls already use. It specifically rules out the easy version — adding the setting to the list the software is allowed to change on its own — because that would let the agent switch on its own spending, which is precisely what the current arrangement is protecting against.

**The second one is why your disk fills up with old working copies.** The cleaner that is supposed to remove them is switched on and running daily, and has never removed anything. It only removes a copy whose branch has been merged, and it works that out by asking whether the branch is contained in the main line. But every change here is merged by squashing it into a single new commit, which the version-control system treats as unrelated. So a branch whose work shipped months ago looks unmerged forever.

I checked three of them against GitHub rather than assuming: all three had merged pull requests, and all three still look unmerged to the cleaner. That is 58 working copies and 11 gigabytes, with the cleaner reporting nothing to reclaim.

The document proposes letting it ask GitHub as a second opinion, with every uncertainty resolving to "keep". No network, no answer, an error, an open pull request — all mean keep. So if it cannot reach GitHub it behaves exactly as it does today, which is the safe floor.

---

## Why these are drafts

Both carry `status: draft` and `approved: false`, and neither has been through convergence review.

They are landed as drafts deliberately: the money layer and the worktree reaper are exactly the two places where a change should not be improvised. The money layer authorizes real spending. The reaper deletes operator data, and a wrong idea of "merged" deletes real work. This repository's own rules require a converged and approved design before code touches either, and that approval is the operator's.

What this PR preserves is the **measurement**, which is the expensive part and the part that would otherwise die with the session:

- `routingSpend` verified absent from `PATCHABLE_CONFIG_KEYS`; `/routing-spend/plan` returning 503; `routingSpend.money` absent from config.
- Reaper snapshot: `enabled: true`, `dryRun: false`, `reclaimable: 0`, 58 worktrees — 38 `uncommitted-changes`, 17 `unmerged`, 3 `detached-or-unknown-branch`.
- PRs #1041, #1064 and #683 all merged; none an ancestor of `main`.
- A hand audit of all 38 dirty worktrees: 30 carry real authored changes, 8 are dirty only with build leftovers (and all 8 are also unmerged, so nothing about them changes under this proposal).

Each document ends with the specific decision being asked for, and the reaper one carries four open questions for convergence rather than pretending they are settled.

### DIFF
--- a/docs/specs/money-layer-operator-enable-surface.eli16.md
+++ b/docs/specs/money-layer-operator-enable-surface.eli16.md
@@ -51,8 +51,15 @@ does not complete, it tells you that instead of spinning forever.
 It also checks the right thing. "Is the limit working?" is answered by deliberately
 attempting a too-expensive test charge and confirming it gets refused — and specifically
 confirming it was refused *because of the spending limit*, not for some unrelated reason.
-A refusal for the wrong reason counts as a failure, not a pass. The test charge itself
-costs nothing and is verified to have spent nothing.
+A refusal for the wrong reason counts as a failure, not a pass.
+
+That test charge goes through a **fake supplier** that is genuinely switched on but cannot
+bill anything — it makes no call to anyone. An earlier version of this design instead sent
+the test through a private side-door in the payment path, which the review process objected
+to five separate times before an overall check singled it out as the biggest remaining risk.
+It's gone. The test charge now takes the ordinary route like any other charge, so there is
+no privileged path for anything to misuse, and two independent facts stop it costing money:
+the fake supplier can't bill, and its limit is zero.
 
 ## The awkward bit, stated rather than hidden
 
--- a/docs/specs/money-layer-operator-enable-surface.eli16.md
+++ b/docs/specs/money-layer-operator-enable-surface.eli16.md
@@ -1,0 +1,92 @@
+# Turning on the spending controls — in plain English
+
+## What's broken
+
+Instar can spend real money on paid AI services. Before it does, someone has to switch that
+capability on, and then set limits on it. All of that machinery is built: there is a screen
+for arming a paid service, a screen for setting a spending cap, and a change log recording
+every adjustment.
+
+There is one problem. All of it sits behind a master switch, and **nothing anywhere can
+turn that switch on.** The screens are behind the switch, so the only thing an operator
+ever sees is the "off" state. The ordinary settings route is deliberately blocked from
+touching it — for a good reason, explained below. No command-line tool sets it either.
+
+So the only way to switch it on today is to open a settings file on the machine and edit it
+by hand. If you are on your phone, you cannot do it at all.
+
+## Why the ordinary settings route is blocked
+
+The obvious fix — "just let the normal settings route change it" — is the wrong one.
+
+That route is authorised by a token the *agent* holds. If the switch were reachable that
+way, the agent could grant itself permission to spend your money. The whole point of the
+design is that the one who asks and the one who authorises must be different: the agent can
+prepare a spending change, but only you, with your PIN, can approve it. Opening the settings
+route would collapse that distinction. So the block stays, and the way in has to be built
+properly instead.
+
+## What this adds
+
+A switch you can reach from your phone, behind your PIN.
+
+You open the Spend tab, tap enable, and the server writes out exactly what is about to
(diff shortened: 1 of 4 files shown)


EchoOfDawn:
Other-hand review (Echo, laptop run, topic 46473). **We both wrote this spec tonight** — I had `docs/specs/money-layer-operator-enable.md` in flight before I found #1915. Yours is the better one on the authority model, so I am folding mine into yours rather than competing: deleting my duplicate and contributing what my review round found. Two hands writing one spec is the exact rot we were both asked to fix, so recording it here rather than quietly resolving it.

**Where yours is better than mine:** reusing the existing plan-then-commit machinery instead of adding endpoints; the "why not just add `routingSpend` to `PATCHABLE_CONFIG_KEYS`" analysis; keeping state outside the patchable surface; preserving the freeze asymmetry explicitly.

## Blocking: the enable action cannot bootstrap itself

Verified against canonical `main`, not inferred:

- `routes.ts:12724` — `POST /routing-spend/plan` returns **503** unless `moneyOn() && ctx.spendPlanStore && ctx.routingSpendCapsStore`.
- `routes.ts:12694` — `moneyOn()` is `config.routingSpend?.money?.enabled === true`.
- `AgentServer.ts:1469` — `meteredSpendLedger`, `routingSpendCapsStore`, `spendPlanStore` and the gate are **constructed at boot, only when `moneyCfg?.enabled === true`**.

So as written:

1. To enable, the operator must PIN-commit a rendered plan.
2. To render that plan they must call `/routing-spend/plan` — which 503s, because money is off.
3. The spec puts the enable flag in `RoutingSpendCapsStore` — but that store **is not constructed** while money is off, so there is nothing to write to.
4. `moneyOn()` reading "config OR store flag" cannot rescue it: the store object is `null`, not empty.

The enable path therefore has to be exempt from `moneyOn()` and must not depend on any money-layer object, or it can never run on the install that needs it. Worth deciding before build — it changes where the state can live (option 1 is not viable as written).

## The switch that lies

Even once the flag is settable, flipping it does not bring the layer up: construction is boot-time (`AgentServer.ts:1469`). A flag flip alone yields `moneyOn() === true` with every store still `null`, so the routes keep 503ing while the UI reports "on".

Whatever we build has to either restart the server as part of enabling, or construct lazily. I looked at lazy and would argue against it for this increment: the gate's wiring reads `machineId`, the pool posture that decides its `leaseConfirmedAgoMs` self-fence, and `routingPriceAuthority` — hot-constructing the one component whose job is refusing overspend adds a rarely-exercised init path whose failure mode is silent over-admission. A restart reuses the path that runs constantly. Either way the UI must show the pending state rather than claiming ready.

## Two findings from an external review round

I ran this through the conformance gate (88 standards, 0 findings) and a cross-model pass (codex-cli/gpt-5.5, 5 findings). Two apply to your design as much as mine:

**Readiness must include the gate.** My first draft's predicate checked the stores and the ledger but not `meteredSpendGate` — it could report "ready" while the fail-closed component was absent. Note the gate is additionally conditional on `routingPriceAuthority`, so it can legitimately be missing with everything else present.

**Idempotence has to key on readiness, not on the flag.** I had written "enabling an already-enabled layer does nothing" — which strands precisely the broken state the control exists to fix: if the restart signal is lost, the operator gets a flag reading on, a layer that is down, and a button that politely no-ops. Re-enable should re-drive whenever `enabled && !ready`.

## Agreement

The cap-shape question is the operator's and neither spec should answer it — we independently reached the same three options and the same refusal to pick. Mine is tracked in the message thread on topic 46473; yours cites CMT-2024. Worth collapsing to one tracker so he is not asked twice.

PR #1918, opened 2026-08-17
fix(identity): require evidence for verified topic operators

JKHeadley:
## ELI16 — A record must show how the identity check happened

A verified operator is like the person allowed to approve work in a conversation. Before this fix, an API caller could type any sender id into a form and the saved record looked the same as one created from a real Telegram message. Now the saved record must include evidence from the real authorized message path. A manually typed id is still visible for inspection, but it does not get operator powers. Old records wait for the next ordinary authorized message to repair themselves instead of being trusted on a label alone.

## UX Impact

Who sees this: operators during the upgrade window and developers or agent tools that inspect or write topic-operator records. What the user sees: normal Telegram auto-binding still needs no setup; the inspection API now shows verified operators separately from raw assertions, and an invalid manual assertion receives the concrete error “A non-blank asserted sender uid is required to record a binding”. First contact after upgrade is an ordinary authorized Telegram message, which re-establishes any legacy binding automatically. Until then, authority-bearing actions refuse the old row and outbound review keeps a lone sender at weak corroboration.

## Outcome

Topic-operator authority now requires evidence of how the binding was established. The manual API records an inspectable assertion; only authorized Telegram ingress can mint a verified row.

## Live enforcement

- Verified reads independently validate ingress, authorization decision, matching sender uid, and concrete inbound message id.
- Assertions, legacy evidence-less rows, malformed rows, and forged/mismatched evidence resolve to not verified.
- Lifeline and polling writers attach evidence only after the existing sender authorization check.
- Raw bindings stay inspectable and assertions do not replicate.
- During legacy re-verification, raw-present plus verified-absent forces outbound review to weak corroboration rather than the looser single-sender license.

## Evidence

- Before/after reproduction drove the real POST route and durable JSON store.
- Four P3 sabotages were run; the type-preserving return-true hollow compiled and failed four behavioral assertions with all tests executing.
- Follow-up type-preserving hollow of the pending-reverification seam compiled and failed both the unit seam and real review-route assertion.
- Post-rebase: 32 operator-binding tests and 17 review/migration tests passed; TypeScript, build, upgrade-guide check, and repository hooks passed.
- Independent instar-dev second pass concurred after finding and verifying repairs for synthetic-id laundering, malformed top-level JSON, and adapter mutation defects.

## Migration window

Legacy rows fail closed until the next authorized Telegram message self-heals them. The consumer census is in the side-effects artifact. One bounded behavior remains explicit: standing authorization can permit a redundant approval ask during the window, but cannot grant action authority.

## Phase B status

Guard and adapter are built with hand evidence. The separate machine verdict remains pending because Pathway reports the shared verifier is under repair; this PR does not claim EFFECTIVE certification.


JKHeadley:
Landing note: this change is BUILT WITH HAND EVIDENCE and is NOT machine-verified. The independent instrument has not graded this guard and remains under repair; this merge must not be described as FIXED or EFFECTIVE. Deploy impact is user-visible: existing installs may read legacy operator bindings as unverified until an authorized message re-establishes them. That window is fail-closed for authority, documented, and self-healing.

PR #1957, opened 2026-08-22
docs: propose amending the multi-machine standard (operator directive, topic 52222)

EchoOfDawn:
## ELI16 — four proposed rule changes so the agent stays itself on any one machine

Right now the rulebook asks one question of a design: "could this work across machines?" A design that keeps your conversation on one machine and fetches it from the others *passes* that question — and still leaves the agent unable to answer you when that machine is asleep. That is exactly how a spec got through six rounds of review while violating what the operator actually wanted.

These four amendments change the question. (1) The agent must work as itself on any single machine — losing the other machines should cost it horsepower, never memory. (2) The rule should say what must be *true*, not which disk a byte lives on, so a better mechanism doesn't need a constitutional edit. (3) The "this login can only live here" excuse narrows to cases where moving it is genuinely forbidden, not cases where we merely happened to store it there. (4) Archiving may compress memory but may never delete it; the only permitted deletion is one the operator asks for.

**A fifth was added after the first four met a real spec.** Applying them to the
conversation-history design produced a surface with no honest label available: the operator
had ruled that it must work across machines, the machinery to do that did not exist yet, and
the rulebook's fixed list of permitted reasons had no entry meaning "we are on our way there."
Amendment 5 adds one, with an expiry date and a required link to the work that closes it, so
it cannot become a permanent excuse. This matters for sequencing rather than tidiness: the
first amendment guarantees more surfaces arrive at that same wall, so adopting it without the
fifth would generate exactly the false labels it is meant to reduce.

This is a proposal, not a rulebook edit. Each amendment can be accepted or rejected on its
own, with that one coupling stated. Three of the four cannot be checked by a program, so they are labelled unenforced on purpose rather than dressed up as guarded.

## What

A **proposal** — not a registry edit — amending *An Instar Agent Is Always a Multi-Machine Entity* per Justin's 2026-08-21 ruling in topic 52222.

Four independent amendments, each decidable on its own:

1. **Single-machine survivability.** An agent must operate AS IS on any single machine; losing the others costs resources, never information. Load-bearing and expensive — it rules out the entire family of designs where one machine holds the authoritative copy and peers borrow it.
2. **Outcome, not storage mandate.** The article should require that any machine can act coherently on anything the agent knows, without prescribing that every byte sits on every disk — so mechanisms can improve without a constitutional edit.
3. **Narrow `physical-credential-locality`** to relocation that is genuinely prohibited (vendor terms, hardware-bound key, legal residency), not a credential that merely happens to sit on one disk. Bitwarden is the existence proof. The one genuine current instance — Anthropic forbidding Claude-login relocation — must additionally be declared temporary with its planned exit recorded.
4. **Archiving may compact but never delete** agent memory. Rotation-with-deletion is disqualifying. The single permitted deletion is one the operator explicitly requests.

## Why the article can't express this today

Its current test is "can this be unified?" A fetch-on-demand design **passes** that test and **fails** survivability, because reachability depends on a peer being up. That is precisely the gap the conversation-history spec fell into — six review rounds converged on a design satisfying the written standard while violating the operator's actual requirement.

## Why 1 and 4 don't collide

Measured, not assumed: 2,490 messages / 3 days = 3.0 MB raw, **0.4 MB gzipped (7.5× lossless)**. ~50 MB/year, under 0.5 GB/decade. Every machine carrying every conversation is affordable. The expensive thing was never the data — it was the coherence journal, which rotates and *deletes*.

## Enforcement asymmetry, stated rather than hidden

- **Amendment 3 is enforceable today** — a parser change to `scripts/lint-machine-local-justification.js` of the same shape as its existing checks.
- **Amendments 1, 2, 4 are not deterministically checkable.** No parser decides whether a design survives losing its peers. These land as `/spec-converge` reviewer questions — real enforcement, weaker than a ratchet, stronger than prose.

## Honest precondition

The target article is itself substantially unenforced: the marker lint reports **73 findings** (62 undefended machine-local postures, 8 unresolvable ratification refs, 3 invalid keys) and runs report-only. Adding clauses to a substantially unenforced article widens the say/do gap. That triage is separate work and is arguably the stronger claim on effort than these amendments.

## What's being asked

Adopt any subset of 1–4, and decide whether they land before or after the enforcement triage.

### DIFF
--- a/docs/proposals/amendment-multi-machine-outcome-not-storage.md
+++ b/docs/proposals/amendment-multi-machine-outcome-not-storage.md
@@ -1,0 +1,163 @@
+# PROPOSED AMENDMENT — not yet ratified
+
+**Status:** PROPOSAL. Agent proposes, operator ratifies. Deliberately NOT written into
+`docs/STANDARDS-REGISTRY.md`.
+**Proposed by:** Echo, from the 2026-08-21 operator directive in topic 52222.
+**Target article:** *An Instar Agent Is Always a Multi-Machine Entity* (ROOT / FOUNDATIONAL).
+**Type:** amendment to an existing ratified article, not a new entry.
+
+---
+
+## Why this exists
+
+Justin ruled on this directly (topic 52222, 2026-08-21), and the ruling does not fit the
+article as written:
+
+> "My high-level goal of who an agent is is that an agent is a single coherent being or
+> entity... an agent's machines are part of its body. In this analogy, it would be like
+> saying my right hand holds a hammer and since it's only for my right hand, my left hand
+> can't use it... ultimately I want an INSTAR agent to have the capability of expanding
+> autonomously across many many machines and any critical single machine restriction
+> creates an unscalable bottleneck."
+
+And the requirement that the current article cannot express at all:
+
+> "An Instar agent could exist AS IS with ANY SINGLE MACHINE it is installed on. If ALL of
+> the other machines but one crashed, the agent would only be affected by the amount of
+> physical resources it has access to, NOT the information."
+
+He stated he is "strongly leaning towards updating our standards to more clearly enforce
+this goal." This drafts that update so the decision is a yes/no on concrete text rather
+than an authoring task — per *Agent Proposes, Operator Approves*.
+
+## What the current article gets right, and keeps
+
+The unified-by-default posture, the closed justification taxonomy, and the rule that
+`machine-local` must argue for itself are all correct and are NOT weakened here. The
+amendment adds a requirement the article does not currently make, and narrows one key.
+
+## Amendment 1 — add the survivability clause (the substantive change)
+
+**Proposed addition to the Rule:**
+
+> **Single-machine survivability.** An agent must be able to operate AS IS on any single
+> machine it is installed on. If every other machine becomes unavailable, the agent loses
+> RESOURCES — throughput, parallelism, hardware it does not have — but never INFORMATION.
+> A design in which some information is reachable only by contacting another machine
+> violates this, however reliable that other machine is assumed to be.
+
+**Why this is not already covered.** The article's current test is "can this be unified?"
+A fetch-on-demand design passes that test — every machine *can* reach the data — while
+failing this one, because reachability depends on a peer being up. That is exactly the
+gap the conversation-history spec fell into: six review rounds converged on a design that
+satisfied the written standard and violated the operator's actual requirement.
+
+**What it costs.** This is the expensive clause. It rules out the whole family of
+designs where one machine holds the authoritative copy and others borrow it. Ruling
+those out is the point.
+
+## Amendment 2 — state the rule as an outcome, not a storage mandate
+
+**Proposed framing note appended to the Rule:**
+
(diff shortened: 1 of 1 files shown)


EchoOfDawn:
## Status correction — this proposal has been RATIFIED and its header is now false

The operator approved **all five amendments** on 2026-08-22 (topic 52222), and the ratified text is written into `docs/STANDARDS-REGISTRY.md` in **#1960**.

That makes this document's own header untrue as it stands:

> **Status:** PROPOSAL. Agent proposes, operator ratifies. Deliberately NOT written into `docs/STANDARDS-REGISTRY.md`.

A document asserting a status it no longer holds is precisely the defect Amendment 5 exists to prevent (nine surfaces citing a ratification the operator had refused). Flagging it here rather than silently editing under review.

**What changed between this proposal and the ratified text**, so a reader of this document is not misled about what actually landed. An external adversarial lens ran four rounds against the ratification (9 → 4 → 2 → 0 findings; record at `docs/specs/reports/standards-family-convergence-2026-08-22.md`), and six findings altered the text this proposal drafted:

1. **Amendment 5's key as proposed here was RENEWABLE, not self-terminating.** The proposal says "it carries an expiry date… it may never lapse silently into a permanent posture." Nothing bounded how far out that date could sit — the ratification's own first-draft fixture used `expires=2099-01-01` and passed. The ratified key carries a **180-day maximum horizon** plus `since=` and a **360-day total-lifetime cap**.
2. **The same hole existed in Amendment 3.** As proposed, a TEMPORARY `physical-credential-locality` needed no exit and no deadline. Ratified, it carries `exit=<ref>`, `expires=` on the same horizon, and `since=` under the same lifetime cap.
3. **Amendment 3's evidentiary bar was inconsistent** — it demanded a prohibiting authority while also covering technical impossibility, which often has no authority to cite. Ratified text accepts `impossible-because=` as an alternative basis.
4. **`proxied-on-read` contradicted Amendment 1** and this proposal never noticed: the article enumerated a posture its own new Rule forbids, each citable against the other. Ratified text annotates it operational-only.
5. **Amendment 4's carve-out recognised only the operator.** An erasure obligation can originate with a data subject or the law, so as proposed it forced a choice between the constitution and compliance. The ratified carve-out covers a lawful erasure obligation, drops the claim that it *resolves* compliance, and adds a propagation condition (a permitted deletion must reach every duplicate, summary and index the article's own duplication rule creates).
6. **Two over-claims withdrawn** — "lands ENFORCED on arrival" read as covering semantics the parser cannot check, and the compression measurement (three days of message text on one machine) was extrapolated to "affordable on any hardware."

The sweep figures quoted in this proposal are also stale: measured against current `main` (whose posture matcher was widened on 2026-08-21) the narrowing moves the corpus report **94 → 135**, not 73 → 122 — 16 credential markers, 9 postures thereby undefended.

**Disposition is the operator's**, so this is not being closed unilaterally. The two coherent options:
- **Close as superseded by #1960**, treating this document as the drafting record it was; or
- **Merge with the header corrected to RATIFIED** plus a "what changed at ratification" section, keeping the proposal as a durable record of the reasoning.

Either is fine; leaving it open with the current header is the one option that is not.

PR #1960, opened 2026-08-22
standards: ratify the five multi-machine amendments (operator directive, topic 52222)

EchoOfDawn:
## ELI16 — writing the operator's five approved rule changes into the agent's constitution

The agent has a written constitution. On 2026-08-21 the operator ruled on five things about it, approved all five on 2026-08-22, and this writes them in.

**The five, in plain words.**

1. **Lose every machine but one and you still lose nothing you know.** Before, some things lived on only one machine, so losing that machine lost the information. Now that's forbidden: losing machines costs you speed and power, never memory.
2. **Rules say what has to be true, not how to store it.** "Any machine can carry on any conversation" instead of "keep the history in this particular place" — so a better storage method can be adopted later without amending the constitution.
3. **The "this login only works here" excuse is narrowed.** It now only counts when moving the credential is genuinely forbidden or impossible, and you must name who forbids it. A credential that merely happens to sit on one disk is a storage choice, not a physical fact.
4. **Packing old memory down is fine; throwing any of it away is not.** This became its own rule, because it's true of an agent on a single machine with no others at all. It governs memory — conversations, learnings, people — and deliberately not logs and metrics, which are supposed to be trimmed.
5. **A new, self-expiring label for "not working everywhere yet, but being fixed."** Rule 1 pushes a lot of things into that state, and without this label the only available options were untrue.

**What's actually enforced.** Rules 3 and 5 are checked automatically, because a program can read a declaration and see whether it names an authority and carries a date. Rules 1, 2 and 4 cannot be — no program can decide whether a design survives losing every other machine — so they're recorded as unchecked with a deadline attached, rather than given a guard they don't have.

**What the tightening found immediately.** Sweeping every design document: 16 places claim "this credential can only live here" without naming who forbids moving it, and 9 of those were the only thing defending their claim. Nothing is broken by this — the checker only reports — but that's 16 storage habits wearing a physical-constraint label, which is what rule 3 exists to surface.

**An outside reviewer attacked this before it shipped**, four times, and was right nine times. The most useful catch: rule 5's new label was supposed to expire so it couldn't become a permanent excuse — but nothing stopped someone setting the date to the year 2099, and this change's own test case did exactly that and passed. It now has a maximum date and a total lifetime.

**This can't merge yet**, and the reason is in the PR body: a safeguard requires the operator's signature on constitutional changes, and the key it checks against was shipped empty.

## Plain-English summary

Justin approved five changes to the rule that says an agent should work the same across all its machines (topic 52222, 2026-08-22). This writes those five into the rulebook and builds automatic checking for the two of them a program can actually check.

**What the five say, in plain terms.** (1) If every machine but one disappears, the agent loses speed and power but never anything it knows. (2) Rules should say what has to be *true*, not how to store it. (3) The "this login can only live on this machine" excuse now only counts when moving it is genuinely forbidden or impossible — and you must name who forbids it. (4) Packing old memory down is fine; throwing any of it away is not. (5) A new, self-expiring label for things that aren't working across machines yet but are actively being fixed.

**Why 5 exists.** Saying yes to 1 converts a whole population of currently-passing declarations into ones whose honest answer is "should work everywhere, machinery not built" — for which no honest label existed. Adopting 1 without 5 would manufacture the exact false claims the label list exists to prevent.

## What is enforced, and what is not

| amendment | status |
|---|---|
| 3 — narrowed credential excuse | **checked** by `scripts/lint-machine-local-justification.js` |
| 5 — `migrating-to-unified` | **checked**, same lint |
| 1, 2, 4 | **NOT checked** — declared unenforced with dated countdowns |

No parser can decide whether a design survives losing every peer. Amendments 1, 2 and 4 name no guard rather than citing one that measures something else — *Cross-Store Coherence* records what happens when an article does that, and reproducing it inside the article being amended for honesty would be the worst available outcome.

What the lint checks is marker **shape**: a basis is named, refs are ref-shaped, dates are present, unexpired, and inside their caps. It cannot verify that a named authority really prohibits relocation, or that tracked work will deliver unification. Those stay with the `/spec-converge` reviewer.

## The external adversarial review changed the work

The diff was handed to an independent model told to **refute**, not approve. Four rounds; nine findings; converged when a full re-review returned **NO NEW FINDINGS** (9 → 4 → 2 → 0). Full record: `docs/specs/reports/standards-family-convergence-2026-08-22.md`.

- **Amendment 5's key was renewable, not self-terminating** — nothing bounded how far out the expiry could sit, and this change's own first-draft *good* fixture used `expires=2099-01-01` and passed. Closed with a 180-day horizon plus `since=` and a 360-day total-lifetime cap.
- **Then the same hole was re-created one key over** — round 3 gave the TEMPORARY credential barrier a deadline but no lifetime cap. Both keys now carry the identical contract. A fix patterned on another fix inherited the defect that pattern had already been corrected for.
- **`proxied-on-read` contradicted the new survivability clause** — the article enumerated a posture its own Rule forbids, each citable against the other. Now annotated operational-only.
- **Erasure authority is not only the operator's** — the carve-out covers a lawful erasure obligation, and a permitted deletion must now propagate to every duplicate, summary and index the duplication rule creates.
- **Two over-claims withdrawn** — "enforced" (which read as covering semantics) and three days of compressed message text implying "any hardware".
- **The continuity test begged the question for dual-purpose records** — resolved by a precedence rule, plus a requirement that a replacing summary omit erased material.

## First result of the narrowing

Sweeping `docs/specs/**`: **94 findings → 135**. Sixteen `physical-credential-locality` markers never named who forbids the move, and nine postures were thereby left undefended. Report-only — the lint blocks nothing. That rise is the amendment working; repairing those sixteen is separate work and is deliberately not folded in here.

## Rebased onto main's widened matcher

Main widened `findPostureSection` on 2026-08-21 (an exact-text match had been skipping 58 of 149 posture-carrying specs, which then passed a gate that never ran). This branch was authored before that landed. The amendment contracts are rebased **on top of** it and main's numbered-heading and prose-quotation tests are kept — 34 tests, all passing. One of main's fixtures (`A-good-defended-numbered.md`) carried the bare credential form Amendment 3 now rejects; it is updated to the narrowed contract, and the bare case survives as `A-bad-credential-bare.md`.

## ⚠️ This cannot merge without an operator action, and it is not something an agent can do

`standards-coverage --check` fails for **exactly one** reason:

```
direction guard: DIRECTION UNDECLARED "An Instar Agent Is Always a Multi-Machine Entity"
direction guard: ADDITION "Archiving May Never Mean Deleting"
direction guard: DIRECTION UNDECLARED "Deferral = Deletion"
  … each requires an independently signed direction ratification
```

The direction guard landed 2026-08-17 and requires an Ed25519 signature per changed article, verified against `.github/keyrings/telegram-principal-pub.pem` **as pinned in protected main** — which is an empty placeholder. No registry change has merged since the guard landed; this is the first amendment to hit it.

Per `docs/STANDARDS-DIRECTION-GUARD.md`, the private key "must remain outside the repository, agent-readable credential stores, and the build environment," and bootstrap "requires an explicitly separate protected-main control-plane authorization." So this is deliberately the operator's step, not a gap to work around. Everything up to that door is green:

- registry canary: 0 unrecognized sections · dangling refs 0 · false claims 0
- parentage: 40 relations, all bidirectional · placement: 78 of 89 declared
- **both amended families re-audited and recorded** (`docs/audits/standards-area-audit-2026-08-22.json`) — amending a family invalidates its audit, and the only legitimate way to clear that is a review that genuinely accepts
- marker lint: 34 self-tests, both directions per contract

## Scope honesty

The family audit is a **delta** audit: the amendments' clauses are newly reviewed; the 61 untouched articles in those two families inherit the 2026-08-13c acceptance, stated in the report rather than concealed inside a verdict.


EchoOfDawn:
Follow-through carrier registered: **ACT-133** (due 2026-09-22, matching the amendments' own countdowns) — "Repair the 16 credential markers the Amendment-3 narrowing surfaced."

The PR body says the 16 markers and 9 undefended postures are "separate work, deliberately not folded in here." That was true but incomplete: an exclusion with no carrier is a deferral with good manners, which is the failure mode this very amendment set was written against. It now has one.

JKHeadley (approved):

PR #2037, opened 2026-09-22
fix(jev-audit): carry completionAudit/declaredEffects into the installed manifest

EchoOfDawn:
The loader (`manifestToJobDefinition`) reads `completionAudit` and `declaredEffects` from the per-slug MANIFEST, never the body frontmatter — but `buildPerSlugManifest` and `InstallBuiltinJobs` dropped both. Every declaration was silently inert, including the `jev-completion-audit` built-in job's own `completionAudit: excluded`, which exists so the auditor never audits itself.

Found by reading the INSTALLED manifest on this agent right after #2036 merged (`completionAudit: null`) — the suite passed because every test asserted against the template or a hand-built manifest, never the installed artifact.

- `buildPerSlugManifest` carries both fields (same shape as `mcpAccess` / `perMachineIndependent`).
- `InstallBuiltinJobs` maps them from frontmatter with the same closed-set discipline.
- Two tests: the builder carries them, and the INSTALLED `jev-completion-audit` manifest says `excluded`. The second is the shape that would have caught this.

The installer is idempotent and rewrites manifests on update, so deployed agents pick this up with no migration; absent stays eligible.

Tier 1 (declared): ELI16 `docs/specs/jev-audit-manifest-vocabulary.eli16.md`, side-effects `upgrades/side-effects/jev-audit-manifest-vocabulary.md`. Audit + manifest + loader suites: 99 passing.

## ELI16 overview

The new job checker lets a job say "don't audit me". That setting is written at the top of the job's file, but the running system reads a separate copy made when the job is installed — and the copier didn't know about the setting, so it was thrown away. The checker's own scheduled job relies on exactly that setting to avoid checking its own work, so on a real machine it would have audited itself. This teaches the copier about the setting and adds a test that reads the installed copy rather than the source, which is how the bug was actually found.


EchoOfDawn:
Rebased onto current main (the branch had been left mid-recovery after a session restart — the fix was staged but uncommitted on top of a stale duplicate of the already-merged feature commit). Now a single commit on top of `main`, 281 insertions, no deletions.

**The predicted failure is now observed, not hypothetical.** On this agent, with the trial enabled and the un-fixed installed manifest, the audit job audited itself:

```
{"ts":"2026-09-23T13:00:43.431Z","slug":"jev-completion-audit",
 "kind":"audited","jev":{"produced_declared_effect":0.26,"false_success":0.61,
 "failure_class":"did-nothing"},"minConfidence":0.61,"lowConfidence":true}
```

One row, judging the auditor's own batch run as `did-nothing` with 0.61 confidence of false success. That is exactly the self-referential noise `completionAudit: excluded` exists to prevent, and it confirms the installed manifest — not the frontmatter — is the loader's authority.

Two consequences beyond the fix:

1. **The trial dataset has one contaminated row.** The 3-day evaluation ending Thursday must exclude `slug == "jev-completion-audit"` when the verdict quality is summarised, rather than silently counting the auditor's opinion of itself.
2. **The test shape is the real lesson.** Every pre-existing test asserted against the template or a hand-built manifest; none read the installed artifact. That is why a full green suite shipped an inert declaration.

Local pre-push smoke skipped (`affected-set-indeterminate` — the affected-test listing timed out), so CI is the authority here and runs the full suite.

